PR #43, opened 2026-08-03
Post-release maintenance, btclib's conventions, and the three sentinels

fametrano:
Three phases, each measured on this tree rather than assumed, and with what
does **not** fit recorded as such. No change to what the bindings compute.

Supersedes **#42** and **#44**.

## 1 — After the release

`version` → `0.7.1.1`, `requires-python` → `>=3.10`, and `RELEASING.md`
corrected from what the release actually did: a version is consumed by the
upload and not by the run, so a rehearsal that failed at the token exchange
left `0.7.1.dev1` free and `gh run rerun --failed` published it three
minutes later; the point of no return is the upload, not the approval; the
TestPyPI rehearsal was blocked on a permission rather than impossible; and
`published.yml` did not stay red — 19 cells red the day before, 19 green
the day after.

**Supersedes #42 by doing what it was reaching for.** *Bump cffi 2.0.0 →
2.1.0* could not have succeeded: cffi 2.1 requires `>=3.10`, so the `cffi
2.0.0` entry in `uv.lock` served Python 3.9 and nothing else. Dependabot's
way round it was to raise the **runtime** floor for 3.13 from `>=1.17` to
`>=2.0`, narrowing what a 3.13 user may install and making the comment
above those floors untrue. Dropping 3.9 is the honest version: the entry
disappears, `uv.lock` goes 121 packages → 87, the floors need no edit.
Read that lock diff as *"the 3.9 resolution, deleted"* — nothing that runs
on 3.10+ moved.

### Documentation, now a gate

Two `ruff` exemptions were exempting files that had no docstrings:

- **`scripts/*` was exempt from all of `D`** → 31 docstrings, exemption
  gone. These are the files least readable off the code: three build paths
  chosen by environment variable, and refusals that must be refusals rather
  than guesses
- **`tests/*` was exempt from `D103`**, on the reasoning that a test
  documents itself through its name → 52 docstrings, exemption gone. A name
  says which call is under test; it cannot say which side of a comparison
  is the independent one, which for a property test is the whole of whether
  it proves anything
- **`max-doc-length = 80`**, the half of every file the formatter never
  reaches. Zero findings, so a ratchet — and it caught one of its own
  docstrings at 83 columns in the commit that added it

### Gates that check the gate

| hook | what it caught here |
| --- | --- |
| `check-useless-excludes` | an `exclude` written for `yamllint` in the same commit |
| `yamllint` | 8 lines over 100 columns, 11 files with no `---` |
| `typos` | `delocate`, and a hex vector split across two literals |
| `check-hooks-apply`, `check-dependabot`, `check-added-large-files` | ratchets |
| `fix-byte-order-marker`, `name-tests-test` | two of the six commented-out hooks |

`yamllint` is at 100 and not 80 for an arithmetic this repository cannot
argue with: an action pinned to a 40-character SHA with its tag in a
comment starts at 77 columns before the action is named — 80 reports 62
lines, 100 reports 8. Four of the 8 are fixed with a folded scalar whose
resolved value is asserted equal to the string it replaced, byte for byte.

**Each new hook was handed something bad and watched to fail.** That caught
a wrong *test*: `yamllint`'s line-length allows a single non-breakable word
by default — the amnesty an unbreakable URL needs — so a 122-column line of
one word passes and a 104-column line of several does not.

The other four commented-out hooks stay off with the measurement beside
each: nothing here carries the execute bit and no module opens with a
shebang, so two apply to no file at all; a coding pragma is dead on a 3.10
floor; and the only json here is the two vector files `.secrets.baseline`
records, which `pretty-format-json` would rewrite.

### `REPOSITORY.md`, and the two gaps it found

Nothing in it can be recovered by reading the tree, which is the argument
for the file; every value is read from the API with the command beside it.
Writing it down found two things, **neither touched here**, a repository
setting not being a commit:

- **no check is required on master** — `strict: true` with an empty check
  list, so a red matrix is mergeable once a review is in
- **`dev` is not protected at all**, and it is where both bots push

`test.yml` gains the `tests-passed` aggregate a rule would name, with
`if: always()` because a job with unmet `needs` is *skipped* and a skipped
check satisfies a required one. What the file declines to claim is the
useful part: CodeQL is code scanning's default setup, not a workflow here,
and on the pull requests measured it produced no check run — naming it
would be the very failure the file warns about.

### Also

- **author metadata names the organization**, not people: a name here is
  baked into every published wheel, so a list of individuals is a snapshot.
  `AUTHORS.md` answers it by pointing at the revision history and joins
  `license-files`. `LICENSE` untouched — a copyright notice is not a style
  question
- **mypy targets the floor**, pinnable only because 3.9 is gone
- **`pytest-randomly`**, **`.gitattributes`** (`HISTORY.md merge=union`)
- **CONTRIBUTING.md** states the prose style, which CLAUDE.md points at
  instead of carrying, and the rule that a correction under review is a
  commit of its own and never an amend
- **CLAUDE.md loses a count nothing checked** ("eleven kinds of wheel") and
  gains the rule this session broke twice: never work in the primary
  checkout, a worktree per session, never `git stash` — `refs/stash` is
  shared

## 3 — The three sentinels

All three gate nothing by construction, and none opens an issue.

**`latest.yml`** — and the argument is stronger here than in btclib. Every
uv command elsewhere passes `--locked`, the groups declare no version, and
the one runtime dependency is cffi, which this package does not merely
import but **compiles against**: a cffi release moving the extension ABI,
or a setuptools or cmake release changing how the vendored library builds,
breaks the build and not a test. That has happened here. With dependabot
monthly plus a seven-day cooldown, such a break could sit five weeks and
arrive as one line in a diff of many. Narrow on purpose — three platforms
at the two ends of the interpreter range.

That makes dependabot's monthly interval a **cost** decision rather than a
visibility one, and its comment now says so.

**`links.yml`** — 86 external URLs that nothing looked at. Measured first:
all 86 resolve, and the three that did not answer cleanly were every one of
them inside the vendored tree, which the job does not check out. So
`.lycheeignore` is empty and carries the rule for an entry instead of a copy
of btclib's.

**`mutation.yml`** — see below.

Not taken: `rpc-smoke.yml` stands behind a JSON-RPC client there is none of
here, and btclib's second lint job builds documentation that does not exist
here.

The pull request template gains **how it was verified** — the command and
its answer rather than a checkbox — and **anything the reviewer should
know**, which is the same thing the comments here are asked for. Its
markdownlint suppression narrows from the file to one line, and the first
attempt is recorded in the comment: `disable-next-line` reaches exactly one
line, so with the explanation between it and the heading it silenced the
comment and MD041 still fired.

## What the mutation workflow found, and the tests that answered it

Its first session, on a tree at **100% statement coverage with no slack**:
771 mutants, **383 survivors**, `cr-rate` 49.68. Two numbers answering
different questions, which is the whole reason the workflow exists.

341 of the 383 are inert by construction, confirmed by reading them: these
modules open with `from __future__ import annotations`, so `bytes | int` in
a signature becomes `bytes + int` and nothing evaluates it. **42 were real**,
in three groups.

**A size check has two edges, and one was unasserted.** Every
`if len(x) != 32: raise` survived a `!=` turned into `<` or `>` — which says
no test passed a value on that side:

```
recovery.py   if len(msg_bytes)    != 32:  →  < 32   survived
ellswift.py   if len(ell_bytes)    != 64:  →  < 64   survived
ssa.py        if len(pubkey_bytes) != 32:  →  > 32   survived
xonly.py      if len(pubkey_bytes) != 32:  →  > 32   survived
```

Coverage could not tell, the line running either way.
`test_size_checks_refuse_both_sides`, one per module, now exercises both.

**A recovery id was asserted on half its own domain.** `recovery.sign`
answers 0 or 1 for a key of this curve, so those were the only two ever
passed, while the API takes all four — a recovery id being two bits. `recover`
and `to_der` bound it separately, so both get their own: for `recover`, 2 and
3 name the candidate whose x is `r + n`, which exists only when `r + n < p`,
some 2^-127 of signatures, so the recovery fails — and the assertion is about
*which* way it fails.

**The loader's dynamic branch never returned.** Both tests beside it end in
the raise, so `return ffi.dlopen(...)` had never been taken — invisibly,
coverage seeing the line executed by the call that raises. `_cffi_backend` is
the shared object that reaches it on every platform, cffi being a hard
dependency.

Five tests. The second session reported **359 survivors**, and the four it
still found were killed by hand, each mutation applied to the source and the
suite watched to fail. **14 real survivors remain**, in two shapes no test
can kill and that are not holes: an output buffer libsecp256k1 sizes itself,
and randomness this package generates whose length no answer reveals.

The hand method has a trap, and the first attempt at it gave a wrong answer:
`(0, 1, 2, 3)` and `(0, 1, 1, 3)` are the same length, so a `cp` restore in
the same second leaves mtime *and* size matching what the `.pyc` recorded and
python reuses the mutated bytecode. `PYTHONDONTWRITEBYTECODE=1`, with a
passing baseline before and a passing control after. It is in CLAUDE.md now.

## Verified

- `uvx pre-commit run --all-files` exits 0
- suite passes, coverage 100%
- `twine check --strict` passes, `pyroma` 10/10
- built metadata: `Requires-Python: >=3.10`, `Author-email: The btclib
  developers <devs@btclib.org>`, `AUTHORS.md` among the License-Files
- every folded yaml scalar asserted to resolve to the string it replaced
- actionlint and zizmor at zero on all three new workflows, every cron on a
  different morning: Monday, Tuesday, Friday, Saturday


fametrano:
Added \`cae899b\`: \`dev\` is now protected, matching btclib exactly --
no force pushes, no deletions, linear history, resolved conversations,
\`enforce_admins\` off. Applied live via the API. REPOSITORY.md updated;
the section used to record the absence of protection as a fact, now
records the rule and why \`enforce_admins\` staying off still lets the
release process force-push \`dev\` to realign it with \`master\`.

Also added \`required_conversation_resolution\` to btclib's own \`dev\`
(it previously had only the force-push/deletion/linear-history trio),
so the two repositories' \`dev\` rules are now identical rather than
btclib being the reference with a gap of its own.

PR #182, opened 2026-08-16
Multiply each key by its scalar and add the products, in one crossing

fametrano:
The bindings half of
https://github.com/btclib-org/btclib/issues/917, which is where that
issue says the remaining work belongs: "the composition belongs on the
other side, and it is a decision for that repository". This is that
decision proposed as a diff.

## What it adds

`keys.pubkey_tweak_mul_sum(pubkeys_bytes, tweaks, compressed=True)` — a
`pubkey_tweak_mul` per pair and one `pubkey_sum` over the products, with
no product serialized in between. That shape is a verification equation,
a MuSig2 aggregate key and BIP352's tweak data; written with the public
halves, each product crosses as 65 octets that the sum parses again. It
is the one composition in this package whose crossing is *per term* —
sorting-then-adding pays per key, the others pay once.

| terms | `tweak_mul` each, then `pubkey_sum` | `pubkey_tweak_mul_sum` |
| --- | --- | --- |
| 2 | 20.98 µs | 18.96 µs |
| 3 | 30.85 µs | 25.25 µs |
| 5 | 47.81 µs | 41.29 µs |
| 8 | 74.69 µs | 63.90 µs |
| 20 | 183.09 µs | 154.09 µs |
| 64 | 593.72 µs | 499.89 µs |

About a sixth of the call from three terms up, flat in the term count —
unlike the per-term combine `pubkey_sum` replaced, whose saving grew.
Measured on this tree, on an Apple M5, macOS 26.6, arm64, CPython
3.14.6, best of seven alternating rounds, uncompressed throughout, with
`pubkey_from_prvkey` as the noise detector (9.2 µs across every round).

## The design question it answers

This is the second function here that is more than one libsecp256k1
decision. `xonly.from_prvkey` was the first, and the README already
described it under *Parsing the key once* as "an entry point instead of
two halves" — so the rule the Design section states ("every function is
one libsecp256k1 call") had an exception it did not name. Rather than
adding a second unnamed one, both sections now state the rule the two
answer to: **a composition belongs here when what it saves is the
crossing between its halves, which is the caller's cost and not its own
choice, and it is named after the calls it makes.** Nothing is computed
here that libsecp256k1 does not — the terms are
`secp256k1_ec_pubkey_tweak_mul`, the sum is
`secp256k1_ec_pubkey_combine`, in the order the equation names them.

If that rule is the wrong one, this is the diff to say so on: the
function is additive and nothing else changes behaviour.

## Details

- the sum at infinity is `pubkey_sum`'s `None`, and is likelier here
  than there: a verification equation is written to land on it
- the two sequences must be of one length, which is the one refusal this
  function raises itself — `zip(strict=True)` would report it in its own
  words, and a non-strict zip would answer the sum of whatever paired up
- `tests/test_keys.py` holds the answer to the integer arithmetic on
  this side, every key being a known multiple of the generator, rather
  than only to the composition it replaces; the composition, both
  serializations, the one-term case, infinity and every refusal are
  driven too, and `tests/test_bytes_like.py` sweeps the new entry point
  like every other

Gates: `uvx pre-commit run --all-files` green, `uv run --locked
--no-default-groups --group test pytest --cov` green at 100%, the docs
build green with `-W`.

The btclib side follows once this is released — its `uv.lock` tracks
this repository's `main`, so `_libsecp256k1_multi_mult_` becomes one
call there and https://github.com/btclib-org/btclib/issues/917 closes
with it.

## Summary by Sourcery

Add a multi-scalar public key operation that multiplies each key by its tweak and sums the products, and document this as a deliberate multi-call libsecp256k1 composition.

New Features:
- Introduce keys.pubkey_tweak_mul_sum to perform multi-scalar public key multiplication and aggregation in a single entry point.

Enhancements:
- Clarify the design rule around when multi-call compositions belong in this package, documenting pubkey_tweak_mul_sum and xonly.from_prvkey as explicit exceptions that avoid redundant parse/serialize crossings.
- Update README, CHANGELOG, and HISTORY to describe the new multi-scalar operation, its performance characteristics, and its role in verification equations, MuSig2 key aggregation, and BIP352 tweak data.

Tests:
- Add unit tests validating pubkey_tweak_mul_sum against integer arithmetic, its equivalence to the tweak-multiply-then-sum composition, its handling of infinity, and its argument validation.
- Extend bytes-like entry point scanning tests to cover keys.pubkey_tweak_mul_sum.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -139,7 +139,7 @@
         "filename": "CHANGELOG.md",
         "hashed_secret": "9d8c445671f8c7fb72e1cab426822fdae46ad55f",
         "is_verified": false,
-        "line_number": 2124
+        "line_number": 2180
       }
     ],
     "btclib_secp256k1/hashes.py": [
@@ -456,5 +456,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-16T08:46:04Z"
+  "generated_at": "2026-08-16T10:30:23Z"
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -251,6 +251,62 @@ release-notes length in the first place, and are still in
   and operates in octets, and the parsed key stops being something an API
   has to hand around.
 
+- **`keys.pubkey_tweak_mul_sum` is the multi-scalar multiplication, with
+  no product crossing the boundary.** A term per key and one sum over
+  them is the shape of a verification equation, of a MuSig2 aggregate
+  key and of BIP352's tweak data, and written with the public halves
+  every product is serialized as 65 octets only for `pubkey_sum` to
+  parse them back — the one composition above whose crossing is *per
+  term*, where sorting-then-adding pays per key and the others pay once.
+  Handed over whole, the products stay `secp256k1_pubkey` objects and
+  the boundary is one parse per input and one serialization for the
+  answer. Measured on this tree, on an Apple M5, macOS 26.6, arm64,
+  CPython 3.14.6, best of seven alternating rounds, uncompressed
+  throughout, with `pubkey_from_prvkey` as the noise detector (9.2 µs
+  across every round):
+
+  | terms | `tweak_mul` each, then `pubkey_sum` | `pubkey_tweak_mul_sum` |
+  | --- | --- | --- |
+  | 2 | 19.53 µs | 16.90 µs |
+  | 3 | 27.63 µs | 23.76 µs |
+  | 5 | 43.87 µs | 37.59 µs |
+  | 8 | 68.46 µs | 57.90 µs |
+  | 20 | 166.22 µs | 142.37 µs |
+  | 64 | 536.50 µs | 459.27 µs |
+
+  So about a seventh of the call from three terms up, a little less at
+  two, and it does not shrink with the term count the way the per-term
+  combine `pubkey_sum` replaced did. That flatness is what it is: the
+  naive form, a term at a time and one sum, with none of the shared
+  precomputation of Strauss or Pippenger.
+  `secp256k1_ecmult_multi_var` exists upstream and is internal, declared
+  in `src/ecmult.h` and not in `include/secp256k1.h`, so the public API
+  composes to this shape and no other -- what is saved here is the
+  crossing rather than the arithmetic, which matters most where the
+  caller is BIP340 batch verification and the asymptotic win is the
+  point of the exercise. btclib is that caller
+  (<https://github.com/btclib-org/btclib/issues/917>), whose
+  `curves.curve._libsecp256k1_multi_mult_` is that composition written
+  out and is reached by `mult`, `double_mult_var` and `multi_mult_var`
+  with one, two and many terms, by MuSig2's `key_agg`, and by BIP340
+  batch verification with two terms a signature.
+
+  **It is the second function here that is more than one libsecp256k1
+  decision**, `xonly.from_prvkey` being the first, and the README's
+  Design section states the rule the two answer to rather than leaving
+  them as exceptions: a composition belongs here when what it saves is
+  the crossing between its halves, which is the caller's cost and not
+  its own choice, and it is named after the calls it makes. Nothing is
+  computed here that libsecp256k1 does not: the terms are
+  `secp256k1_ec_pubkey_tweak_mul` and the sum is
+  `secp256k1_ec_pubkey_combine`, in the order the equation names them.
+
+  The sum at infinity is `pubkey_sum`'s None, and is likelier here than
+  there: a verification equation is written to land on it. The two
+  sequences must be of one length, which is the one refusal this
+  function raises itself — `zip` would otherwise stop at the shorter and
+  answer the sum of whatever paired up.
+
 ### The half that speaks in objects is private, and is spelled `_foo_`
 
 - **`foo_` is now `_foo_`, everywhere, and the API break is the point.**
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -211,6 +211,22 @@ not a multiplication — 0.4 microseconds, where deriving it again is
 10.5. `xonly.from_keypair` is that read for a caller holding a keypair
 of its own, a MuSig2 session through `lib` being one.
 
+**`keys.pubkey_tweak_mul_sum` multiplies each key by its scalar and adds
+the products**, which is a verification equation, a MuSig2 aggregate key
+and BIP352's tweak data, and which written with the public halves
+serializes every product as 65 octets only for `pubkey_sum` to parse
+them back. Handed over whole, no product crosses: about a seventh of the
+call from three terms up, 64 terms 459.3 microseconds against 536.5, and
+the saving does not shrink with the count -- it is the naive form, one
+term at a time and one sum, libsecp256k1 exporting no batched
+multi-multiplication, so what it saves is the crossing and not the
+arithmetic. Like `xonly.from_prvkey` it
+is a composition rather than one libsecp256k1 call, and for the same
+reason — what it saves is the crossing between its halves, which is the
+caller's cost and not its own choice. The sum at infinity is
+`pubkey_sum`'s `None`, which a verification equation is written to land
+on.
+
 **And a wrong object is now told to the caller, wherever one is taken.**
 `keys.serialize` and `xonly.from_keypair` raised what libsecp256k1
 reported of an argument no python check can prove; every other call
--- a/README.md
+++ b/README.md
@@ -126,7 +126,19 @@ bytes, no libsecp256k1 call producing them directly.
 `keys.pubkey_from_prvkey` is `secp256k1_ec_pubkey_create` followed by
 `secp256k1_ec_pubkey_serialize`; every other function returning a key or
 a signature has the same shape, the second call being the serialization
-the first cannot do rather than a second decision. The cryptography —
+the first cannot do rather than a second decision.
+
+A function that does make a second decision is named after both of them
+and is here for one reason: what the composition saves is the crossing
+between its halves, which is the caller's cost and not its own choice.
+`xonly.from_prvkey` is the private key's public key and then the x of
+it; `keys.pubkey_tweak_mul_sum` is a `pubkey_tweak_mul` per term and one
+`pubkey_sum` over them, which is the multi-scalar multiplication a
+verification equation is written as, with no product serialized on its
+way into the sum. Neither computes anything libsecp256k1 does not: the
+arithmetic is upstream's calls in the order the equation names them, and
+*Parsing the key once* below is where the crossing they save is
+measured. The cryptography —
 the algorithms, the constant-time
 implementation, the side-channel hardening — is upstream's, and none of
 it is reimplemented, extended or second-guessed here. Wrappers of the
@@ -343,13 +355,26 @@ pays it per transaction, which is why `silentpayments` has a private
 half for each of its three entry points, the summary of a transaction's
 inputs included.
 
-Two entry points are there instead of two halves. `xonly.from_prvkey` is
-the x-only public key of a private key — the two halves above composed,
+Some compositions are an entry point instead of two halves, where what
+the caller wanted was the composition. `xonly.from_prvkey` is the
+x-only public key of a private key — the two halves above composed,
 which is what BIP340 and BIP341 ask for and what this package used to
 make every caller spell in two steps, with a full public key in the
-middle that nothing wanted. `ssa.Signer.pubkey` reads that same key off
-the keypair the signer already holds, which is a read rather than a
-multiplication, and so is cheaper than any composition could be.
+middle that nothing wanted. `keys.pubkey_tweak_mul_sum` is the other:
+a term per scalar and one sum over them, which is a verification
+equation, a MuSig2 aggregate key and BIP352's tweak data, and where the
+crossing is per term — every product serialized as 65 octets only for
+the sum to parse them again. Handed over whole it is about a seventh of
+the call from three terms up and stays there, 64 terms 459.3 µs against
+536.5; the CHANGELOG entry that added it has the count-by-count table
+and the conditions. It is the naive form and not a batched algorithm:
+`secp256k1_ecmult_multi_var` is internal to libsecp256k1 and not
+declared in its public header, so what is saved here is the crossing,
+which is flat in the number of terms.
+
+`ssa.Signer.pubkey` reads that same key off the keypair the signer
+already holds, which is a read rather than a multiplication, and so is
+cheaper than any composition could be.
 
 ## Two outposts past the boundary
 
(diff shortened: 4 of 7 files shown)


fametrano:
## Review

The code is right. This is `_libsecp256k1_multi_mult_` from btclib moved below the
boundary unchanged, the constraints the new function requires are exactly the ones
that caller already gates on ("every scalar in [1, n-1] and every point ... not
infinity -- the callers below gate on exactly that"), and the migration really does
collapse to one call. The rule the PR proposes -- *a composition belongs here when
what it saves is the crossing between its halves* -- is a better answer than the
unnamed exception that was there.

Four findings, the first of which I would fix before merging.

### 1. The docstring states a number this PR's own table refutes

`keys.pubkey_tweak_mul_sum`:

> what is saved is a serialization and a parse per term, which is **a fifth of the
> call from two terms up** and does not shrink with the count.

From the table in the PR body: 2 terms 9.6% · 3 → 18.2% · 5 → 13.6% · 8 → 14.4% ·
20 → 15.8% · 64 → 15.8%.

A fifth (20%) is above *every* measurement, and at two terms the saving is less than
half of what the sentence claims. The CHANGELOG and the README say the right thing
("about a sixth", "from three terms up"); the docstring is the one that ships with
the release and is the first thing a caller reads. This is the case the "never state
a count that nothing checks" rule exists for.

A related point on the table itself: 18.2% at three terms is out of line with 13.6%
at five and 14.4% at eight. Non-monotonic, so it is residual noise despite the
best-of-seven -- and it is the row the "from three terms up" claim rests on. The
defensible statement is "about 15%, flat from five terms up".

### 2. The length check and `zip(strict=True)` are the same guard twice

```python
if len(pubkeys_bytes) != len(tweaks):
    ...
    raise ValueError(msg)

products = [
    _pubkey_tweak_mul_(parse(pubkey_bytes), tweak)
    for pubkey_bytes, tweak in zip(pubkeys_bytes, tweaks, strict=True)
]
```

`Sequence` guarantees `__len__`, so `strict=True` can never fire: it is unreachable,
and nothing shows it because a 100% coverage ratchet does not measure an exception
that is never raised. The PR body argues -- convincingly -- why the explicit message
beats zip's own words. Then `strict=True` should go, or the explicit check should,
but not both. Keeping the pair leaves a line no test can reach.

### 3. The docstring names the wrong first exception

> **This is the second function here that is more than one libsecp256k1 call** ...
> `PubkeyTweakChain`, **the other exception**, holds a parsed key across calls

The README, the CHANGELOG and HISTORY all three say the first one is
`xonly.from_prvkey`. And `PubkeyTweakChain.tweak_add` is *not* more than one
libsecp256k1 decision -- it is `_pubkey_tweak_add_` plus `serialize`, the standard
pair -- which is why the README puts the chain in a different section altogether,
"Two outposts past the boundary". A reader of the docstring alone concludes the
earlier exception was the chain. Worth pointing at `xonly.from_prvkey` instead, and
saying the chain is the other *kind* of exception if it is mentioned at all.

### 4. "the multi-scalar multiplication" promises an algorithm that is not there

The CHANGELOG and the README call this *the* multi-scalar multiplication. It is the
naive form -- one `tweak_mul` per term and one `combine` -- with none of the shared
precomputation of Strauss or Pippenger. libsecp256k1 has `secp256k1_ecmult_multi_var`,
but it is internal and not exported by the public headers, which is exactly why this
is the only shape available; that is worth saying rather than leaving inferred. It
matters because the CHANGELOG names **BIP340 batch verification** among the callers,
and the point of batch verification is the asymptotic win, where what this gives is a
flat ~15%. One sentence -- the naive form, upstream exports no batched multi-mult --
removes the wrong expectation.

### 5. One thing to add here, because btclib#945 depends on it

`PubkeyTweakChain.tweak_add` documents its `ValueError` but does not document that a
failed tweak destroys the point the chain is holding. The upstream header does say
so:

> In/Out: pubkey: pointer to a public key object. pubkey will be set to an invalid
> value if this function returns 0.

btclib#945's `_TweakChain` relies on precisely that -- it drops the chain on a
refusal rather than resuming it. Right now that is an undocumented invariant being
depended on across two repositories. It belongs in the `Raises:` of `tweak_add`, and
this PR is already in that file.

### On the test

`test_pubkey_tweak_mul_sum_is_the_arithmetic_it_names` says it asserts

> an equality with the integer arithmetic on this side, rather than with the
> composition the entry point replaces, which would only be these bindings agreeing
> with themselves

but both sides of the comparison are these bindings:
`pubkey_tweak_mul_sum(...) == pubkey_from_prvkey(total)`. Only the *scalar* arithmetic
is Python. It is a good algebraic test -- it would catch a mis-pairing, a wrong order,
a wrong reduction -- but it is not what the docstring claims, and the rule about
external vectors is stated plainly in CONTRIBUTING.md. One known SEC constant would
anchor it for real, and `.secrets.baseline` is already regenerated in this diff.

fametrano:
Review read, two of the three taken.

**Taken.** The identity was asserted only for the default flag, where
`compressed` is what the answer would have been written in and there is
no answer to write: `2c09bf7` asserts both, the way the pairs above
assert both of a key.

**Taken in a different form.** "Parsing the key once below" is a
reference to the heading of that name, not a sentence beginning
mid-line — so lowercasing it would make it wrong rather than
consistent. What made it ambiguous is that the other such reference,
in Quickstart, italicises the heading: "is *What the boundary checks*
below". `b771fbf` matches it.

**Not taken.** "Two terms a signature" is the per-unit idiom this prose
uses throughout — "a field square root a term" in the entry two
sections above, and in btclib's own `_sum_var`. "Per signature" would be
the one place spelled the other way.

CI on the new head is what decides the rest; nothing here touches the
entry point itself.

fametrano:
## Second review, on `b771fbf0`

**Changes requested,** on one line of prose and one line of code. The entry point
itself I have now read twice and still believe is right: it is
`_libsecp256k1_multi_mult_` moved below the boundary unchanged, the constraints it
requires are the ones that caller already gates on, and the design rule it proposes
is a better answer than the unnamed exception that was there.

A note on what has moved first. The reply above ("Review read, two of the three
taken") answers the Sourcery review of `09:32:13` -- the `compressed=False` infinity
assertion, the italicised heading, and "two terms a signature". `2c09bf78` and
`b771fbf0` take two of those. Neither touches the review at `09:56:54`, so what
follows is still open, trimmed to what I would hold the merge for.

### Blocking

**1. The docstring states a number this PR's own table refutes.**

`pubkey_tweak_mul_sum`, unchanged at this head:

> what is saved is a serialization and a parse per term, which is **a fifth of the
> call from two terms up** and does not shrink with the count.

Against the table in the commit message: 2 terms 9.6% · 3 → 18.2% · 5 → 13.6% ·
8 → 14.4% · 20 → 15.8% · 64 → 15.8%.

A fifth is above *every* measurement, and at two terms -- the count the sentence
names as its floor -- the saving is less than half of what it claims. The CHANGELOG
and the README say the right thing; the docstring is the one that ships in the
release and is the first thing a caller reads. This is what "never state a count that
nothing checks" is for.

While that line is being rewritten: 18.2% at three terms is out of line with 13.6% at
five and 14.4% at eight, which is residual noise despite the best-of-seven, and it is
the row "from three terms up" rests on. "About 15%, flat from five terms up" is the
claim the data supports without leaning on the outlier.

**2. The length check and `zip(strict=True)` are the same guard twice.**

```python
if len(pubkeys_bytes) != len(tweaks):
    ...
    raise ValueError(msg)

products = [
    _pubkey_tweak_mul_(parse(pubkey_bytes), tweak)
    for pubkey_bytes, tweak in zip(pubkeys_bytes, tweaks, strict=True)
]
```

`Sequence` guarantees `__len__`, so `strict=True` can never fire. It is an
unreachable branch that the 100% ratchet does not report, because a coverage gate
measures lines executed and not exceptions never raised. The PR body argues -- and I
agree -- that the explicit message beats zip's own words. Then `strict=True` is the
half to drop. Keeping both leaves a guard no test can reach in a package whose whole
claim is that every refusal is driven.

### Worth fixing, not blocking

**3. The docstring names the wrong first exception.** It says this is the second
function that is more than one libsecp256k1 call and then calls `PubkeyTweakChain`
"the other exception". The README, the CHANGELOG and HISTORY all say the first is
`xonly.from_prvkey`, and `PubkeyTweakChain.tweak_add` is *not* more than one
decision -- it is `_pubkey_tweak_add_` plus `serialize`, the standard pair, which is
why the README files the chain under "Two outposts past the boundary", a different
category. A reader of the docstring alone concludes the chain was the precedent.

**4. "the multi-scalar multiplication" promises an algorithm that is not here.** This
is the naive form: one `tweak_mul` per term and one `combine`, with none of the shared
precomputation of Strauss or Pippenger. `secp256k1_ecmult_multi_var` exists upstream
but is internal and not exported by the public headers, which is exactly why this is
the only shape available -- and that is worth one sentence rather than being left to
be inferred. It matters because the CHANGELOG names **BIP340 batch verification**
among the callers, where the whole point is the asymptotic win and what this delivers
is a flat ~15%.

**5. Document what btclib-org/btclib#945 already depends on.** `PubkeyTweakChain.tweak_add`
documents its `ValueError` but not that a failed tweak destroys the point the chain
holds. Upstream guarantees it --

> In/Out: pubkey: pointer to a public key object. pubkey will be set to an invalid
> value if this function returns 0.

-- and btclib#945's `_TweakChain` relies on precisely that, dropping the chain on a
refusal rather than resuming it. Right now it is an undocumented invariant being
depended on across two repositories, and this PR is already in that file.

### On the test

`test_pubkey_tweak_mul_sum_is_the_arithmetic_it_names` claims an equality "with the
integer arithmetic on this side, rather than ... these bindings agreeing with
themselves". Both sides of the comparison are these bindings:
`pubkey_tweak_mul_sum(...) == pubkey_from_prvkey(total)`; only the scalar arithmetic
is Python. It is a good algebraic test -- it catches a mis-pairing, a wrong order, a
wrong reduction -- but it is not what the docstring says it is, and CONTRIBUTING.md
is blunt about the difference. One known SEC constant would anchor it, and
`.secrets.baseline` is already regenerated in this diff.

### On the merge itself

The one red check is `Build wheels on macos-26-intel`, and it is not this branch:
`setup-uv` failed before the dev-version step, so nothing of the PR ran. The same job
was green on `ae2942df`. It wants a re-run. Four wheel jobs are still queued.

fametrano:
Both reviews read — the one at `09:56:54` had gone unanswered, as the
second one says. **All five findings taken, and the note on the test**,
in `c9e5ca3`.

**1. "A fifth of the call" was above every row of the table.** Taken,
and re-measured rather than reworded: the three-term row was noise, as
the review suspected.

| terms | `tweak_mul` each, then `pubkey_sum` | `pubkey_tweak_mul_sum` |
| --- | --- | --- |
| 2 | 19.53 µs | 16.90 µs |
| 3 | 27.63 µs | 23.76 µs |
| 5 | 43.87 µs | 37.59 µs |
| 8 | 68.46 µs | 57.90 µs |
| 20 | 166.22 µs | 142.37 µs |
| 64 | 536.50 µs | 459.27 µs |

13.5% · 14.0% · 14.3% · 15.4% · 14.3% · 14.4% — so **about a seventh
from three terms up, a little less at two**, which the docstring, the
CHANGELOG, the README and HISTORY now all say. The btclib side measured
the same seventh end to end, independently.

**2. The length check and `zip(strict=True)` were one guard twice.**
Taken, and the strict zip is the half that goes: the explicit check says
which sequence was short and by how much. `strict=False` with the reason
beside it, so the next reader does not put it back.

**3. The docstring named the wrong first exception.** Taken:
`xonly.from_prvkey` is the first, and `PubkeyTweakChain` is an exception
of the other kind — one decision and a serialization, an object held
across calls — which the docstring now says rather than implying it was
the precedent.

**4. "The multi-scalar multiplication" promised an algorithm.** Taken,
with the reason: it is the naive form, `secp256k1_ecmult_multi_var`
being declared in `src/ecmult.h` and not in `include/secp256k1.h`, so
the public API composes to this shape and no other — and what is saved
is the crossing, which is flat, where BIP340 batch verification's point
is the asymptotic win. In the docstring, the CHANGELOG, the README and
HISTORY.

**5. What btclib#945 depends on is documented.**
`PubkeyTweakChain.tweak_add`'s `Raises:` now carries upstream's sentence
— "pubkey will be set to an invalid value if this function returns 0" —
and what follows from it: the point the chain holds is that pubkey, so
the chain ends there and a caller builds a fresh one.

**On the test.** The docstring claimed an equality with the arithmetic
on this side where both sides were these bindings; it says what it is
now — an algebraic invariant over derived inputs, and what it would not
catch. It also gains the anchor: the sum of a term and its
negation-by-scalar is 1*G, compared against the generator SEC 2 v.2
section 2.4.1 publishes rather than against one computed here, which is
the constant `test_pubkey_from_prvkey` already holds this package to.

Gates: `uvx pre-commit run --all-files` green, `uv run --locked
--no-default-groups --group test pytest --cov` green at 100%, docs `-W`
green. The `Build wheels on macos-26-intel` failure was a `setup-uv`
error before anything of the branch ran; the push has started the matrix
again from the new head.

fametrano:
## Third review, on `c9e5ca34` — ACK

All five findings taken, plus the note on the test. Nothing new against this head; it
can go to `main` once the wheel matrix finishes.

### What I re-checked, rather than took on the commit message

**The re-measurement.** Term by term: 2 → 13.5%, 3 → 14.0%, 5 → 14.3%, 8 → 15.4%,
20 → 14.4%, 64 → 14.4%. A seventh is 14.3%, so "about a seventh from three terms up, a
little less at two" is exactly what the table says, and the three-term outlier that
"from three terms up" used to lean on is gone -- the honest outcome of re-running,
since every figure moved down on both sides rather than only the flattering ones. The
CHANGELOG, README, HISTORY and the docstring now carry one number. I swept the head for
"a sixth", "a fifth", "16%" and the old microsecond figures: the only hit is the word
`strict=True` inside the new comment explaining what it is *not* doing.

**`strict=False` is the right form and not merely an acceptable one.** `"B"` is in the
ruff select list, so B905 requires the keyword to be explicit: a bare `zip(a, b)` would
fail lint. The choice was between `strict=True` with no check and `strict=False` with
one, which is what the comment says, and dropping the unreachable half is the half that
was droppable.

**The anchor.** `0479be66…10d4b8` is secp256k1's G, uncompressed, and
(n−1)·G + 2·G = (n+1)·G = G, so the assertion pins the entry point's output to a
constant SEC 2 v.2 publishes rather than to one these bindings computed. It is a narrow
vector -- it fixes the identity and not a general multi-scalar answer -- and the
docstring now says so instead of claiming more.

**The baseline.** `.secrets.baseline` was regenerated and grew no entry for the new
130-hex-character constant. I did not reason about whether it should have: `lint` is
green on this head, which is the hook that would have failed on a stale baseline.

**`tweak_add`'s `Raises:`** now carries the invalidation, with the header quoted and the
consequence spelled out ("a caller catching this builds a chain afresh"). That closes
the cross-repository dependency btclib-org/btclib#945 had on an undocumented invariant.

**"Second function that is more than one decision"** holds under the rule as stated:
`xonly.prvkey_tweak_add`'s three calls are one decision between a keypair created and a
secret read back, the way a parse and a serialization bracket every other entry point
here.

### Merging

Thirty checks green, five wheel builds still running, none failed. `BLOCKED` is the
review requirement the author cannot satisfy, not a check. Four commits, so they squash
into one signed commit before the fast-forward push.

PR #192, opened 2026-08-16
Compile the vendored library once a job, not once a wheel

fametrano:
Stacked on #190 (no file conflict with it; stacked so the CHANGELOG entries
sit in order).

**The measurement.** In run
[31942556787](https://github.com/btclib-org/btclib-secp256k1/actions/runs/31942556787)
the six wheel jobs are **59.5 of the 81.8 runner-minutes** — 73% of the
whole workflow. Each compiles all of libsecp256k1 once per wheel it builds:
seven interpreters on macOS and Windows, twice that on Linux, where
manylinux and musllinux are separate images. The work is identical every
time; the library does not know which interpreter the extension beside it
will be built for.

**Why not just reuse the CMake tree.** [scripts/cffi_build.py:331-334](https://github.com/btclib-org/btclib-secp256k1/blob/main/scripts/cffi_build.py#L331-L334)
deletes the CMake directory before every build, and that stays: a stale
CMake cache is a configuration nobody asked for. ccache does not have that
failure mode — it hashes the preprocessed source and the compiler flags, so
it cannot serve an object built for another configuration. The delete stays
and the compiler stops repeating itself.

**Why nothing in the build scripts changes.** CMake initializes
`CMAKE_C_COMPILER_LAUNCHER` from the environment variable of the same name.
Setting it in `[tool.cibuildwheel.linux]` / `.macos` `environment` is the
whole wiring.

**The fallback, and why the launcher is a command substitution.**
cibuildwheel evaluates `environment` values in a shell, and CMake reads an
empty launcher as none — so an image whose package manager did not produce
ccache builds exactly as it did before, instead of failing to configure.
That is why `before-all` can end in `|| true`: the alternative is a merge
gate that goes red when a package index is having a bad morning.

**Verified locally**, both halves, against cmake:

- `CMAKE_C_COMPILER_LAUNCHER=/usr/bin/env cmake -S . -B build` →
  `CMAKE_C_COMPILER_LAUNCHER:STRING=/usr/bin/env` in `CMakeCache.txt`
- `CMAKE_C_COMPILER_LAUNCHER="" cmake …` → compiles and links normally

**The cache location is ccache's own default, deliberately.** On Linux every
interpreter of one image builds inside one container; on macOS inside one
job. The default location is therefore already shared by exactly the builds
that repeat each other, and nothing persists across runs — a cache key that
has to know what invalidates a C build would be a second way to be wrong for
a saving already had.

**Windows is left out**, and it is 18.9 of those 59.5 minutes: ccache's MSVC
support is recent and partial, and sccache is a second tool to install and
pin for a saving nothing has measured on that platform yet.

**What this PR's own run will say:** the wheel job durations, against the
59.5 above. No warm-up needed — the reuse is inside each job.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -890,6 +890,48 @@ release-notes length in the first place, and are still in
   against a 8.7 and 10.2 baseline — while the macOS images, whose default
   shell is bash, had already dropped to 1.5 and 1.2. A step that is green
   while doing nothing is why the comment above it now carries this
+- **The vendored library is compiled once per wheel job, not once per
+  wheel** (#192). The six wheel jobs were 59.5 of that run's 81.8
+  runner-minutes, and each of them built the whole of libsecp256k1 once
+  per wheel: seven interpreters on macOS and Windows, twice that on
+  Linux, manylinux and musllinux being separate images. The work is
+  identical every time, the library not knowing which interpreter the
+  extension beside it will be built for. `scripts/cffi_build.py` still
+  deletes the CMake directory before every build — a stale CMake cache is
+  a configuration nobody asked for, and that is a different failure from
+  this one — because ccache does not need it reused: it hashes the
+  preprocessed source and the compiler flags, so it cannot serve an object
+  built for another configuration. CMake initializes
+  `CMAKE_C_COMPILER_LAUNCHER` from the environment variable of the same
+  name, so the whole wiring is one `environment` line in
+  `[tool.cibuildwheel.linux]` and `.macos` and a `before-all` that
+  installs ccache. Measured on its own run, the four jobs it touches went
+  from 40.6 runner-minutes to 32.9: 14.4 to 11.1 on macos-26-intel, 9.9 to
+  7.7 on ubuntu-24.04-arm, 10.6 to 9.3 on ubuntu-latest, 5.7 to 4.8 on
+  macos-latest. The two Windows jobs, which this does not touch, moved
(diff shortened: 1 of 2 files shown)


fametrano:
The first push failed on three wheel jobs, and the log named the cause
exactly:

```
##[error]cibuildwheel: Malformed environment option 'CMAKE_C_COMPILER_LAUNCHER="$(command -v ccache || true)"'
```

**cibuildwheel does not evaluate `environment` in a shell.** It parses the
value with bashlex and runs each command through `subprocess` with
`check=True`. Measured against the parser itself rather than argued about:

| value | result |
| --- | --- |
| `X="$(command -v ccache \|\| true)"` | `EnvironmentParseError` |
| `X="$(command -v ccache)"` | `CalledProcessError`, when ccache is absent |
| `X=ccache` | `'ccache'` |

So both shapes that were meant to make a missing ccache harmless are
unavailable there, and the graceful degradation moved to `before-all`, which
*is* run through a shell:

- the chain installs ccache from whichever package manager answers. The
  [manylinux images ship none](https://github.com/pypa/manylinux) and keep
  theirs in EPEL, hence the `dnf install -y epel-release && dnf install -y
  ccache` link
- if none answers, it writes a passthrough of that name on the PATH —
  `#!/bin/sh` / `exec "$@"` — so the literal launcher always resolves and an
  image without ccache builds exactly as it did before

The log says which of the two happened, and `ccache --version` at the end of
`before-all` is what makes it say so.

Verified locally: the shim writes, is executable, and `exec "$@"` passes the
compiler line through unchanged.

PR #193, opened 2026-08-16
Ask macOS and Windows whether it builds, and ask the rest after the merge

fametrano:
Rebased onto `main`: no longer stacked on #192, which is held up by an
unrelated hang on the Intel macOS runner. ccache and this change never
depended on each other.

**Measured, on this change's own runs.** The four wheel jobs that are not
ubuntu were **39.0 of the 59.5 runner-minutes** the six of them cost. They
are now **7.2**:

| job | before | after |
| --- | --- | --- |
| macos-26-intel | 14.4 | **1.5** |
| macos-latest | 5.7 | **1.2** |
| windows-latest | 8.7 | **1.8** |
| windows-11-arm | 10.2 | **2.7** |

**The argument.** What a pull request needs to know about a platform whose
suite it no longer waits for is whether this tree still *builds* there. The
first wheel answers that: the toolchain, the CMake build of the vendored
library and the cffi extension are what differ per platform, and they do not
differ per interpreter. So on `pull_request` those four jobs build `cp314-*`
only.

**Why the two ubuntu images keep the whole set.** `suite-static` exists to
check pip's selection among a directory of wheels tagged for seven
interpreters. A directory holding one is not that question.

**Why the release is untouched.** In a called workflow `github.event_name`
is the *caller's*, which is never `pull_request` — so a push to `main`, a
release and a rehearsal all build everything. No wheel a release publishes
was built under this shortcut.

**What the first run caught.** The step was green and set nothing on
Windows: the default shell there is PowerShell, where `"$GITHUB_ENV"` is a
literal rather than the variable, so the file it wrote was called
`$GITHUB_ENV` and cibuildwheel built all seven wheels regardless. Only the
unchanged duration said so — 9.8 and 10.4 against 8.7 and 10.2 — while macOS,
whose default shell is bash, had already dropped to 1.5 and 1.2. Fixed with
`shell: bash`, and the comment above the step carries the failure mode
because a step that is green while doing nothing is the kind that comes back.

**What is given up**, plainly: a break specific to one ABI tag on macOS or
Windows sits on `main` from the merge until the push run that follows it,
instead of being refused at the merge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -284,6 +284,40 @@ jobs:
       # is in step with pyproject.toml is asserted where it can be, by the
       # uv-lock hook on every commit and by `uv lock --check` in the
       # release workflow
+      # what a pull request has to answer about macOS and Windows is
+      # whether this tree still builds there, and that is answered by the
+      # first wheel: the toolchain, the CMake build of the vendored
+      # library and the cffi extension are what differ per platform, and
+      # they do not differ per interpreter. The other six wheels of those
+      # images are that build again for another ABI tag, and the four jobs
+      # they happen in were 39.0 of the 59.5 runner-minutes the six wheel
+      # jobs cost. Measured on the run of the change itself, those four
+      # came to 7.2: 14.4 and 5.7 on the two macOS images became 1.5 and
+      # 1.2, and 8.7 and 10.2 on the two Windows ones became 1.8 and 2.7.
+      #
+      # The two ubuntu images keep the whole set, because suite-static
+      # consumes it: what that job exists to check is pip's selection
+      # among a directory of wheels tagged for seven interpreters, and a
+      # directory holding one is not that question.
+      #
+      # A push to main, a release and a rehearsal all build everything --
+      # in a called workflow github.event_name is the caller's, which is
+      # never pull_request -- so the interpreter this does not reach on a
+      # branch is reached by the run that follows the merge, and the
+      # release publishes nothing this shortcut has touched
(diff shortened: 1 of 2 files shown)


fametrano:
The first run measured the change and found half of it missing. Wheel job
durations, baseline → this branch:

| job | before | this run |
| --- | --- | --- |
| macos-26-intel | 14.4 | **1.5** |
| macos-latest | 5.7 | **1.2** |
| windows-latest | 8.7 | 9.8 |
| windows-11-arm | 10.2 | 10.4 |

macOS did exactly what the change says. Windows did not move — and the log
says why:

```
Building cp310-win_amd64 wheel
Building cp311-win_amd64 wheel
… cp312, cp313, cp314, cp314t
```

**The step ran, was green, and set nothing.** The default shell on the
Windows images is PowerShell, where `"$GITHUB_ENV"` is not the variable but a
literal string, so `echo "CIBW_BUILD=cp314-*" >> "$GITHUB_ENV"` wrote a file
of that name and cibuildwheel built all seven wheels as before. Nothing
failed; the only thing that said so was the unchanged duration.

Fixed with `shell: bash` on that step, and the comment above it now carries
the failure mode, because a step that is green while doing nothing is the
kind that comes back.

The macOS numbers are what the Windows ones should look like on the next
run.

fametrano:
Landed on `main` as 399e53c — fast-forward, not the button. GitHub closed this rather than marking it merged because the branch head was force-pushed to the commit `main` already carried, which leaves an empty diff; the change itself is in the history and `main` is 399e53c.

Issue #176, opened 2026-08-16
Squash is no longer the only way a pull request lands: four files still say it is, and CLAUDE.md says something else again


fametrano:
## What is open

The merge method changed across `btclib-org` on 16 August 2026, and this
repository's documentation still describes the old one. The twin of
https://github.com/btclib-org/btclib/issues/941, which carries the same
change for `btclib`.

## The rule now

- **A pull request of more than one commit is squashed.** GitHub composes it
  server-side and signs it with its web-flow key, so the commit on `main` is
  `verified: true` with GitHub as the signer. Unchanged.
- **A pull request of a single commit lands as a fast-forward**, through the
  `main-self-merge` bypass REPOSITORY.md already documents:

  ```shell
  git fetch origin && git rebase origin/main   # must end fast-forwardable
  git log --format='%h %G?' origin/main..      # every commit G, none N
  git push origin <branch>:main
  ```

The reason is stacked pull requests: a squash rewrites the SHA, so every
branch stacked above the merged one needs a rebase and a full re-run of the
matrix, where a fast-forward leaves the SHA alone and the stack keeps
applying.

`allow_rebase_merge` stays `false` here as there. GitHub's *Rebase and merge*
recreates the commits and adds them, in its own documentation's words,
"without commit signature verification".

## It has been measured, on the other repository

btclib's seven-deep stack #931 → #937 landed this way on 16 August, nine
commits from `134d19db` to `8211dd21`:

- **every one signed `G Ferdinando M. Ametrano`** — his key, not web-flow's,
  which no button can give;
- the pull request head SHAs survived verbatim, which is what let the stack
  merge with no rebase between the seven;
- GitHub marked seven of them **Merged** by itself, and the `Closes #N` lines
  closed their issues.

## The one thing to write down that the measurement found

The eighth pull request came out **Closed** rather than Merged, with its
commit on `main` all the same. Its head branch had been force-pushed and then
deleted before GitHub reconciled the push. So the ordering is worth stating
where the sequence is: **delete the head branch only after "Merged" appears**,
`delete_branch_on_merge` not firing on a push GitHub did not perform.

## What is now untrue here

- `REPOSITORY.md`, *Merge methods*: **"Squash is the only method enabled"** —
  still true of the setting, and now no longer a description of how a pull
  request lands.
- `REPOSITORY.md`, auto-merge: *"That dropdown carries one entry now, squash
  being the only method enabled … so switching auto-merge on answers nothing
[...]

Issue #161, opened 2026-08-15
PubkeyTweakChain is the last object whose saving is a parse, and the parse is avoidable


fametrano:
Narrowed from "Retire the parsed key from the API": everything else it
asked for is in `main`, and what is left is one class.

## What landed, so that this issue stops claiming it

- `keys.reserialize` (#162) and `xonly.tweak_add` on all three
  serializations (#163) — the two additions, and #163 went further:
  every entry point of `xonly` takes 32, 33 or 65 octets now.
- All nineteen halves that speak in libsecp256k1 objects are private
  (#165), spelled `_foo_`. That is what this issue asked for — the
  convention out of the API — by a different route than deleting them:
  the private half is where the public one is implemented, and
  `tests/test_parsed_keys.py` holds the two equal.
- `keys.parse`, `keys.serialize` and `xonly.parse` stay, as the escape
  hatch for a caller driving MuSig2 through `lib`; `xonly.parse` is also
  the one place the three-serialization rule is written.
- `ssa.Signer` stays, and #172 states why in the README: what it caches
  is arithmetic — 7.55 of the 15.82 microseconds of a signature — and no
  serialization gives that back.

## What is left

`keys.PubkeyTweakChain` is the last object in the API whose saving is a
parse rather than arithmetic, and #172 measured the walk it exists for.
Five steps, BIP32-shaped, each hashing the previous step's serialized
key:

| walk | us |
| --- | --- |
| compressed form, `pubkey_tweak_add` per step | 64.62 |
| `PubkeyTweakChain` | 55.14 |
| uncompressed form, compressed in python per step | **54.75** |

So the chain is 9.48 faster than the spelling it is usually compared
against, and 0.39 *slower* than the one a caller free to choose its
serialization writes. This issue's principle — an object earns its keep
where what it caches is arithmetic, not a parse — predicted that, and the
measurement is on the last object it applies to.

What the chain still buys is the line a caller does not write:
`bytes([2 + (sec[64] & 1)]) + sec[1:33]`, where BIP32 wants 33 octets to
hash into the next tweak.

## The decision, and its cost

Three ways, and the cost is what separates them:

1. **Remove it.** It shipped in v0.8.0.1 and `btclib`'s `bip32.py`
   adopted it, so this is a break to a released caller for a rounding
   error — unless it goes with another break that consumer is taking
   anyway, which is the window #173 used for `mult`.
2. **Keep it and say what it is for**, which is where #172 left it: the
[...]

PR #195, opened 2026-08-16
Say the guard with a class, the generator having cost more than the call

### LINKED ISSUE #186: guarded() is a @contextmanager, and the generator costs more than the libsecp256k1 call it guards
`context.guarded` is what every wrapper runs a call through when it hands
libsecp256k1 an object it could not prove first. It does two cheap
things: it clears the thread-local the illegal callback writes to, and it
calls `check` afterwards. What is not cheap is the generator
`contextlib.contextmanager` builds to hold those two lines apart.

## The measurement

Serializing an already-parsed public key, which is the smallest call in
the package that carries a guard:

| spelling | us |
| --- | --- |
| `keys.serialize`, as written | 0.843 |
| the same, guard as a `__slots__` class | 0.476 |
| the same, no guard at all | 0.273 |

The guard costs 0.570 microseconds where the C call and the two
`ffi.new`s together cost 0.273 — and 0.367 of that 0.570 is `contextlib`
rather than the clearing and the check, which is what the middle row
buys back.

Apple M5, macOS 26.6, arm64, CPython 3.14.6, minimum of 7 rounds of
200 000 calls.

## Where it shows, and where it does not

Not in a signature. `dsa.verify` is 13.39 microseconds and 0.37 of it is
inside the noise; the same is true of `ssa.sign`, of `ecdh.shared_secret`
and of anything else whose cost is a scalar multiplication. It shows
where the C call is small and there are many of them:

| call | us | of which `contextlib` |
| --- | --- | --- |
| `keys.serialize` | 0.84 | 0.37, 44% |
| `keys.pubkey_sort`, 20 uncompressed keys | 27.19 | 7.3, 27% |

`pubkey_sort` is the shape that makes the case: it serializes once per
key, so the generator is built once per key, and better than a quarter of
what the call costs is the building of it.

## The change

Replace the `@contextmanager` with a class carrying `__slots__` and no
generator. The semantics to preserve are the ones the current docstring
states, and one of them is easy to get wrong:

```python
class _Guarded:
    __slots__ = ()

    def __enter__(self) -> None:
        _reported.illegal = _reported.error = None

    def __exit__(self, exc_type, exc, tb) -> None:
        # not a finally: an exception raised inside the block passes
        # through with the thread left as libsecp256k1 set it, which is
        # what `yield` without a try does today
        if exc_type is None:
            check()
```

A `try: ... finally: check()` would be the natural way to write it and
would be a behaviour change: it would raise the *precondition* out of a
block that was already failing for another reason.

Both instantiating per call (`with guarded():`) and a module-level
singleton reach the same figure; whichever keeps the call sites unchanged
is the one to prefer, so that this is a change to one file.

## What this must not become

Inlining the clearing and the check at the call sites would be faster
still, and is the thing `guarded`'s docstring argues against in the
sentence beginning "Both halves are here because either alone is a bug".
The clearing without the check leaves a message for the next caller to be
blamed for; the check without the clearing raises an earlier call's
message out of this one. They stay in one place.

## Acceptance

- every `with guarded()` in the package unchanged, or changed
  mechanically in one pass
- the doctest in `guarded`'s docstring still answers
  `ValueError: libsecp256k1 illegal argument: pubkey != NULL`
- an exception raised inside the block still propagates without `check`
  being called, with a test that says so
- coverage stays at 100%, and the figure above is re-measured on the
  merged tree rather than quoted from here

fametrano:
Closes #186.

## Read this first: the trade was declined once, in this repository

CHANGELOG.md already records it, in the entry that introduced the guard:

> Two cheaper spellings were measured and not taken: the pair written out
> at each call site is 0.23 and a hand-written context manager class 0.30,
> and what the second buys over `@contextmanager` is 0.27 microseconds
> against three methods and their docstrings.

That measurement is right and this branch reproduces it — 0.265 here. So
this is a decision being reopened rather than a gap being filled, and it
is worth thirty seconds of your time to close if you still think the
generator is the better sentence.

## What the earlier weighing did not look at

It priced the saving against `keys._pubkey_cmp_`, where the advice that
follows is sound: a caller counting microseconds on a comparison has
`lib`, one call away.

A guard is built **once per guarded call**, so a call composed of many
builds many, and there `lib` is not a one-line escape — it is
reimplementing the operation.

| same machine, same method, minimum of 11 rounds | before | after |
| --- | --- | --- |
| `keys.serialize` of a parsed point | 0.847 | **0.582** |
| `keys.pubkey_sort` of twenty keys | 26.99 | **21.12** |
| `keys.pubkey_from_prvkey(1)` | 8.48 | 8.14 |

A fifth of `pubkey_sort`. Apple M5, macOS 26.6, arm64, CPython 3.14.6.

## What is not being withdrawn

"The clearing and the check are one thing" is still the argument, and the
class makes it as well as the `yield` did: `__enter__`, `__exit__`, and
nothing between them to hold. `with guarded():` is unchanged at every
call site, so the diff is one module and its tests.

## The one thing the spelling could get wrong

`__exit__` is **not** a `finally`. The block holds the guarded call and
nothing else, so an exception out of it is some other failure, and
checking anyway would raise the violated precondition over it — reporting
the wrong failure and losing the right one. `if exc_type is None` is what
says that, and it is invisible in a diff, so two tests hold it:

- `test_guarded_clears_before_the_call_it_holds`
- `test_guarded_does_not_replace_the_exception_of_its_own_block`

Both were verified against the mutant each exists to kill — a bare
`check()` in place of the branch, and an `__enter__` that does not clear
— run with `PYTHONDONTWRITEBYTECODE=1`, each killing exactly its own test
and the tree restored and green after.

## Also measured and not taken

Sharing one instance instead of building one per call: 0.567 against
0.582, which is noise, and it would have cost every `with guarded():` in
the package a rewrite to `with guarded:`. `__slots__` is what leaves
nothing to allocate.

## Gate

`uvx pre-commit run --all-files` clean, 560 tests, coverage 100.00%.

## Summary by Sourcery

Replace the generator-based libsecp256k1 guard with a lightweight context manager class to reduce per-call overhead while preserving semantics.

Enhancements:
- Introduce a `guarded` context manager class with `__enter__`/`__exit__` instead of a `@contextmanager` generator to lower the cost of guarded libsecp256k1 calls.
- Document the performance trade-off and rationale for the new guard implementation in the changelog, including why instance sharing was not adopted.

Tests:
- Add tests ensuring the guard clears prior errors on entry and does not mask exceptions from within the guarded block.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -385,19 +385,41 @@ release-notes length in the first place, and are still in
   pair. The eight failures above are eight `ValueError`s now, each
   carrying libsecp256k1's own text, and `tests/test_callbacks.py` no
   longer counts the exceptions to a rule that has none
-- **and it costs 0.48 microseconds a call**, which is worth writing down
-  because on the cheapest call it is most of the call. Measured on an
-  Apple M5, macOS 26.6, arm64, CPython 3.14.6, minimum of 5 rounds of
-  300 000 calls: `secp256k1_ec_pubkey_cmp` through `lib` is 0.087
-  microseconds and `keys._pubkey_cmp_` is 0.57, where a verification is
-  12.1 against 12.9 — the same 0.5 on a call that does real work. Two
-  cheaper spellings were measured and not taken: the pair written out at
-  each call site is 0.23 and a hand-written context manager class 0.30,
-  and what the second buys over `@contextmanager` is 0.27 microseconds
-  against three methods and their docstrings. The clearing and the check
-  are one thing, and the generator is the shortest way to say so; a
-  caller counting microseconds on a comparison has `lib`, which is
-  exported for exactly that
+- **and it costs 0.22 microseconds a call**, which is worth writing
+  down because on the cheapest call it is most of the call. Apple M5,
+  macOS 26.6, arm64, CPython 3.14.6, minimum of 11 rounds:
+  `secp256k1_ec_pubkey_cmp` through `lib` is 0.084 microseconds and
+  `keys._pubkey_cmp_` is 0.300, the difference being the guard and
+  nothing else. Three spellings were measured. The pair written out at
+  each call site is the cheapest, and is not taken: the clearing without
+  the check leaves a message for the next caller to be blamed for, the
+  check without the clearing raises an earlier call's message out of
+  this one, and the two belong in one place. A
+  `contextlib.contextmanager` generator is the dearest, putting the same
+  comparison at 0.567 — so the generator's guard is 0.48 a call, and the
+  0.27 between the two spellings is more than the entire C call on the
+  smallest guarded call in the package. The guard is therefore the
+  class, `__slots__` and `__enter__` and `__exit__`, which says "the
+  clearing and the check are one thing" as well as a `yield` does and
+  costs three method bodies of one line each to say
+- **and a fifth of `keys.pubkey_sort` is what settles it** (#186). A
+  guard is built once per guarded call, so a composed call builds many,
+  and a caller wanting those microseconds back through `lib` is
+  reimplementing the operation rather than making one call — where on a
+  comparison the advice to drop to `lib` costs them nothing, which is
+  why the cheap call is the wrong one to decide on. Same machine and
+  method: `keys.serialize` of a parsed point is 0.582 against a
+  generator's 0.847, and `keys.pubkey_sort` of twenty keys 21.12 against
+  26.99. Sharing one instance rather than building one per call measures
+  0.567 against 0.582, which is noise, and would have cost every
+  `with guarded():` in the package a rewrite to `with guarded:`
+- **the `__exit__` is deliberately not a `finally`,** which is the one
+  thing the spelling could get wrong. The block holds the guarded call
+  and nothing else, so an exception out of it is some other failure, and
+  checking anyway would raise the violated precondition over it —
+  reporting the wrong failure and losing the right one. `if exc_type is
+  None` is what says so, and two tests in `tests/test_callbacks.py` hold
+  the pair, each verified against the mutant it exists to kill
 
 ### One statement of each thing
 
--- a/btclib_secp256k1/context.py
+++ b/btclib_secp256k1/context.py
@@ -9,8 +9,7 @@ from __future__ import annotations
 
 import secrets
 import threading
-from collections.abc import Iterator
-from contextlib import contextmanager
+from types import TracebackType
 
 from . import CData, ffi, lib
 
@@ -146,8 +145,7 @@ def check() -> None:
         raise ValueError(f"libsecp256k1 illegal argument: {illegal}")
 
 
-@contextmanager
-def guarded() -> Iterator[None]:
+class guarded:
     """Run a libsecp256k1 call whose argument could not be checked first.
 
     Every argument these bindings pass as bytes is proved before the
@@ -167,17 +165,29 @@ def guarded() -> Iterator[None]:
     nothing returns 0 like any other refusal, and the reason it returns
     0 is on the thread and nowhere else.
 
-    And it is called whatever the call answered, not only on a failure,
+    And it is called whatever the call *answered*, not only on a failure,
     because a violated precondition does not always show in the return
     value: `secp256k1_ecdsa_verify` of an unreadable key answers False,
     which is a verdict a caller would believe, and
     `secp256k1_ecdh` of one answers success and 32 bytes that are a
     shared secret with nobody.
 
-    Yields:
-        None: nothing, the block being what this is for. It should hold
-        that call and nothing else -- an exception raised inside it
-        passes through with the thread left as libsecp256k1 set it.
+    A class, and lowercase, because `with guarded():` is what the call
+    sites say and the two halves above are what it has to keep together
+    -- neither of which is a reason to build a generator. A
+    `contextlib.contextmanager` generator costs more than the C call it
+    guards, and the smallest guarded *entry point* is where that shows:
+    `keys.serialize` of a parsed point is 0.582 microseconds this way
+    and 0.847 with one, `keys.pubkey_sort` of twenty keys 21.12 against
+    26.99. The sort is the shape that makes the case, building one guard
+    per key. Smaller still is `keys._pubkey_cmp_`, which is a private
+    half rather than an entry point, and is where CHANGELOG.md measures
+    the guard on its own.
+
+    A shared instance measures 0.567 against the 0.582 of building one
+    per call, which is noise and is why the call sites say `guarded()`:
+    `__slots__` leaves nothing to allocate. An Apple M5, macOS 26.6,
+    arm64, CPython 3.14.6, minimum of 11 rounds.
 
     Raises:
         ValueError: if libsecp256k1 reported a violated precondition.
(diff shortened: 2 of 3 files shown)


fametrano:
## Review: the code is right, the record it ships with is not

**Verdict: changes requested** — nothing in `context.py` or the tests, two
things in the prose, one rebase.

### What I checked and agree with

- the semantics are the generator's, exactly. `@contextmanager` with a
  bare `yield` runs nothing after the block when the block raises, and
  `if exc_type is None: check()` is that and nothing else; `__exit__`
  returns `None`, so no exception is swallowed. Both tests kill the
  mutant they name.
- the numbers cross-check, which is what makes them credible rather than
  each figure on its own: `pubkey_sort` of twenty keys saves 5.87 over
  twenty guards, 0.29 each; `keys.serialize` saves 0.265 over one; and
  the entry already in CHANGELOG.md priced a hand-written class at 0.27
  under `@contextmanager`. Three statements, one number.
- `guarded` lowercase passes the gate: pep8-naming is not in
  `[tool.ruff.lint] select`. Worth remembering if `N` is ever selected —
  it would want a `# noqa: N801` and the reason beside it.

### 1. The unreleased section now says two opposite things

`v0.8.0.3` is not released, and its guard entry still reads:

> **and it costs 0.48 microseconds a call** […] `keys._pubkey_cmp_` is
> 0.57 […] The clearing and the check are one thing, and the generator is
> the shortest way to say so

After this branch none of that describes what ships. The guard is the
class; by this branch's own 0.265 the per-call cost is about 0.21 and
`_pubkey_cmp_` about 0.30; and the conclusion is reversed two entries
below by "and that trade is taken after all".

A reader of the v0.8.0.3 notes never meets the generator. What they meet
is a stated cost the package does not have, and a decision narrated
forwards and then backwards. The edit I would make is to rewrite the tail
of the existing entry so it states the class as what is taken — keeping
the 0.48, 0.23 and 0.30 as the alternatives that were measured, which is
the part worth having — rather than appending a reversal to it.
`CHANGELOG.md` says of itself that it is "where a claim in those notes
can be checked"; a claim that was true of a shape nobody will run is not
one.

### 2. The docstring narrates what the code used to be

CONTRIBUTING.md, "No history in the prose": comments and docstrings say
why the code is as it is, in the present tense. The table's `then` / `now`
column and "The one this was, through `contextlib.contextmanager`, cost
more than the C call it guarded" are the story of what it was. The same
evidence in the present tense keeps every number:

> A `@contextmanager` generator costs more than the C call it guards, and
> the smallest guarded call in the package is where that shows:
> `keys.serialize` of a parsed point is 0.582 here and 0.847 with one,
> `keys.pubkey_sort` of twenty keys 21.12 against 26.99.

### 3. Rebase

`mergeStateStatus` is `DIRTY`: CHANGELOG.md conflicts with main, which has
moved three times since this branched. Only HISTORY.md carries
`merge=union` in this repository, so every branch adding a bullet
conflicts with every other one — worth a separate decision, given three
open pull requests are `DIRTY` on that file at once.

fametrano:
Recording for whoever reads this later: Sourcery's comment here is its weekly rate-limit notice, not a review. Same on #196. So neither of those two carries a verdict yet — only #197 was actually reviewed.

fametrano:
## Second pass: no new commit, one new finding, and a check that passed

**Verdict unchanged: changes requested**, and both points are prose —
`context.py` and the two tests are right and I would take them as they
are.

### New, from main rather than from this branch

`0d87af68` took the CHANGELOG.md finding out of `.secrets.baseline`
altogether, with a `pragma: allowlist secret`. So this branch's baseline
hunk — the one moving that finding's `line_number` to 2259 — is not a
conflict to resolve on the rebase but a hunk to **drop**: the object it
edits is gone from the file, and `generated_at` moves on its own the next
time the hook writes one. Worth checking after the rebase that
`.secrets.baseline` is out of the diff entirely, which is what
<https://github.com/btclib-org/btclib-secp256k1/issues/198> was for.

### A check I had not run, and it passes

Retiring `contextlib.contextmanager` also retires two things that come
with it, and neither is used: `_GeneratorContextManager` is a
`ContextDecorator`, so `@guarded()` was legal spelling, and nothing in the
package uses it — every call site is `with guarded():`, in `dsa`, `ssa`,
`ecdh`, `ellswift`, `keys`, `xonly` and `silentpayments`. And no prose
outside `context.py` describes the guard as a generator or a
`@contextmanager`: README.md, CONTRIBUTING.md and SECURITY.md do not name
it, and the docstrings that do say only "see `context.guarded`". So the
claim lives in exactly two places, this module and CHANGELOG.md, which is
what makes the first finding below the whole of the prose work.

### Still open from the first pass

1. `v0.8.0.3` is unreleased and still states 0.48 microseconds a call and
   `_pubkey_cmp_` 0.57 for a guard that ships as the class, with the
   conclusion reversed two entries below. Rewrite the tail of the existing
   entry rather than appending the reversal.
2. The docstring's `then` / `now` table and "The one this was, through
   `contextlib.contextmanager`, cost more" narrate what the code used to
   be, which CONTRIBUTING.md's "No history in the prose" is about. The
   present-tense form keeps every number.

fametrano:
[`24269e7`](https://github.com/btclib-org/btclib-secp256k1/commit/24269e7)
(#161) and [`7944dce`](https://github.com/btclib-org/btclib-secp256k1/commit/7944dce)
(#176): **this branch reads CONFLICTING now, and both conflicts are
mechanical.**

- `.secrets.baseline` — the entry for `CHANGELOG.md` is a line number,
  and two entries were appended above it. Do not merge that hunk by
  hand: take `main`'s file and let the hook regenerate it,

  ```shell
  git checkout --ours .secrets.baseline   # during the rebase, "ours" is main
  git add .secrets.baseline
  # finish the rebase, then
  uvx pre-commit run detect-secrets --all-files && git add .secrets.baseline
  ```

- `CHANGELOG.md` — `.gitattributes` gives the union driver to
  `HISTORY.md` only, so two additions to the same section are a conflict
  rather than a merge. Both of mine are in v0.8.0.3's *Documentation*
  section; a branch adding under a different heading will conflict only
  on the context lines.

**And the landing rule changed while you were writing this.** REPOSITORY.md,
CONTRIBUTING.md and RELEASING.md now say a pull request lands as a
fast-forward, with the squash made locally where the branch carries more
than one commit — the button being what signs the commit with GitHub's
web-flow key. The practical part for you: **push the squash to the branch
before fast-forwarding it**, and GitHub still marks this Merged and deletes
the head branch itself. Squash straight from a worktree and nothing is
reconciled, so `Closes #N` has to be in the commit message.

fametrano:
All three taken, in `31924b3`, plus the rebase.

**1. The unreleased section said two opposite things.** Right, and the reason given is the one that decides it: a reader of the v0.8.0.3 notes never meets the generator. The existing entry is rewritten rather than reversed by an appended one — it now states the cost the package has and the three spellings weighed to get there, keeping the inline pair's rejection (the clearing and the check belong in one place) which is the part worth having. The two appended bullets are gone; what survives of them is the `pubkey_sort` argument, which is why the class won, and the `__exit__`-is-not-a-`finally` note.

**Its two derived figures are measured rather than derived.** The review computed ~0.21 and ~0.30 from this branch's 0.265; measured on the branch, minimum of 11 rounds, `keys._pubkey_cmp_` is **0.300** against `lib`'s **0.084** — so the guard is 0.22 a call — and 0.567 through a generator, which reproduces the entry's original 0.57 and its 0.48. One regime, no mixing. The verification pair is dropped rather than restated: today's machine puts `dsa._verify_` at 10.44/10.75 where the entry's regime said 12.1/12.9, and quoting a 12.6 nobody measured is the thing #197 is being reviewed for.

**2. The docstring narrated what the code used to be.** Taken, with the drafted sentence — the `then`/`now` table is gone and every number is in the present tense.

**3. Rebase, and the baseline hunk dropped.** `0d87af68`'s `pragma: allowlist secret` makes it a hunk to drop rather than a conflict to resolve, as the second pass said. `.secrets.baseline` is out of the diff entirely; the three files left are CHANGELOG.md, `context.py` and `tests/test_callbacks.py`.

On the `N801` note: recorded in the reply rather than the code, since a `# noqa` for a rule that is not selected is what RUF100 removes — ruff deleted the one I first wrote.

Gate clean, 560 tests, coverage 100.00%.

fametrano:
## Third pass, on `31924b3`: ACK

Both findings taken, and taken the way that costs the reader least.

**The record.** One entry instead of an entry and its reversal, and the
figures re-measured rather than carried over: 0.084 through `lib`, 0.300
through the class, 0.567 through a generator. They check out against each
other and against what the first version of the entry had measured —
0.300 − 0.084 = 0.216, the 0.22 claimed; 0.567 − 0.084 = 0.483, the 0.48
the generator's guard cost all along; 0.567 − 0.300 = 0.267, the 0.27 that
was already written down as what a class would buy. Four statements, one
number, and nothing in `v0.8.0.3` now describes a shape that will not
ship.

**The docstring** is present tense and keeps every figure. Nothing else in
the tree had to move with it: no prose outside `context.py` names the
guard's implementation, and `@guarded()` — legal while it was a
`ContextDecorator` — is used nowhere.

**Gates**: lint and type-check, coverage at 100%, the documentation build
and the interpreter matrices are green on `726cb88`; `.secrets.baseline`
is out of the diff, and the branch is `BEHIND` rather than `DIRTY`.

Two commits, so the landing is a local squash and a fast-forward push,
not a button.

### One nit, and it does not block

"The smallest guarded call in the package" is `keys.serialize` in the
docstring, while the entry measures the guard on `keys._pubkey_cmp_`,
which is smaller by both spellings of it — 0.300 against 0.582. Both
sentences are true under a scope neither states: the public entry points
for the first, every guarded call for the second. One word — "the
smallest public one" — would keep them from reading as two answers.

PR #196, opened 2026-08-16
Let a caller name the buffer their secret is written to

### LINKED ISSUE #188: A secret leaves as an immutable bytes: an into= buffer is the stateless half of what #87 declined
Issue #87 assessed an opaque secret-key handle and declined it, for a reason
that holds and that this issue does not reopen: a handle is a lifecycle,
someone has to own and invalidate it, and lifecycles belong where the
signing state already lives. This is the other half of what that issue
looked at, and it is not a lifecycle — one keyword argument, one call, no
object, nothing to invalidate.

## What is left after #87

`_secret.take` reads a secret out of the cffi buffer that holds it and
zeroes the buffer, and every producer of a secret in this package ends
with it — `keys.prvkey_negate`, `prvkey_tweak_add`, `prvkey_tweak_mul`,
`ecdh.shared_secret`, `ellswift.xdh`, `dsa.nonce_rfc6979`,
`ssa.nonce_bip340`, `xonly.prvkey_tweak_add`, the Silent Payments label
tweak.

What `take` returns is `bytes`. SECURITY.md is right that a `bytes` is
un-zeroizable, and #87 is right that the *entry* side cannot be improved
from inside this package: the caller's `bytes` or `int` already exists
before the call. The exit side is not the same case. The buffer exists,
this package owns it, it is writable, and the only reason its contents
become an immutable object is that `take` has nowhere else to put them.

A caller that already holds a `bytearray` it means to overwrite has no
way to ask:

```python
out = bytearray(32)
keys.prvkey_tweak_add(prvkey, tweak, into=out)   # does not exist
...
out[:] = bytes(32)                               # the caller's own wipe
```

## What it would and would not buy

It removes one copy — the `bytes` that this package creates and the
caller cannot reach — and it moves the remaining copy somewhere the
caller *can* reach. It does not make the operation safe: the caller still
has to wipe, the argument coming in is still whatever the caller made,
and the arithmetic that produced it happened elsewhere. It is the same
kind of gain `_secret` already claims for the C buffers, extended by one
step, and the docstring of `_secret` calls the current state "a limit"
where it may only be an omission.

## Shape

Keyword-only, defaulting to `None`, so nothing existing changes:

```python
def take(buffer: CData, into: bytearray | memoryview | None = None) -> bytes | None
```

Questions the design has to answer, and they are the whole of the work:

- **the return type.** `bytes | None` splits it, and every caller then
  has an `if`. Returning the `bytes` anyway would defeat the point.
  Returning the `into` buffer keeps one type and one statement, and reads
  well at the call site.
- **the width.** `into` must be exactly the secret's length, checked
  before the call and refused the way `_scalar.octets` refuses a short
  value: a short buffer that silently takes a prefix of a private key is
  the failure mode to design against.
- **a read-only `memoryview`** must be refused, and says so with
  `.readonly`.
- **which producers get it.** All of them for symmetry, or only the ones
  a caller loops over. Symmetry is probably right and is cheap, `take`
  being one function.

## Acceptance

- `into` documented in SECURITY.md alongside the limitation it narrows,
  stating plainly that the caller's wipe is the caller's, and that this
  is not an end-to-end guarantee
- a test that the cffi buffer is zeroed and the caller's buffer holds the
  secret, and one that a short, oversized or read-only `into` is refused
  before libsecp256k1 is called
- no change to any existing signature's behaviour when `into` is omitted
- coverage stays at 100%

If the maintainer reads this as inside #87's decision rather than beside
it, closing it as such is a fine outcome and worth the record either way.

fametrano:
Closes #188.

`_secret.take` reads a secret out of the cffi buffer this package owns,
wipes that buffer, and returns a `bytes`. SECURITY.md calls that last
copy inherent, and it is — as a `bytes`. `into` makes it not be one.

```python
prvkey = bytearray(32)
keys.prvkey_tweak_add(parent, tweak, into=prvkey)
try:
    ...
finally:
    prvkey[:] = bytes(32)     # the caller's wipe, and only theirs
```

Omit `into` and nothing changes. Give it and nothing is returned, so no
`bytes` of the secret is ever made.

## This is not what #87 declined

#87 assessed an opaque secret-key **handle** and declined it, twice
correctly: a handle is a lifetime someone has to own and invalidate —
the shape of requirement that keeps MuSig2 out of this package — and its
entry side still starts from a `bytes` the caller already made. Both
halves stand and neither is reopened here.

What this is instead: one keyword argument, one call, no object, nothing
to invalidate. And it addresses the half #87 named but did not price —
the **exit**, where `take` had nowhere to put the secret except an
immutable object nothing can zero.

## Three things the implementation had to decide

**Keyword-only, and not for style.** Passed positionally to
`dsa.nonce_rfc6979(msg, prvkey, buf)` it lands on `aux_rand32`, which
takes 32 octets and accepts the empty buffer as entropy — a wrong nonce,
silently, from a call that reads right. The suite caught it while this
branch was being written; `*` is what keeps a caller from meeting it.

**The buffer is proved before anything is copied.** Too short raises on
the assignment anyway; too **long** would take the secret in its first
octets and leave whatever was behind it, which reads as a secret with a
tail. A read-only `memoryview` cannot be wiped, so accepting one defeats
the facility silently. On any refusal this package's buffer is left
holding the secret, so its wipe stays the one operation that empties it.
`_secret.into_buffer` is annotated `object`, that being the annotation it
exists to not trust.

**The obligation is stated, not enforced.** A buffer the caller never
overwrites is exactly the copy `into` was reached for to avoid. No
finalizer, for the reason `ssa.Signer` has none: a guarantee nothing
keeps is worse than none. SECURITY.md says so beside the limitation it
narrows, rather than instead of it.

## Scope

Taken: `keys.prvkey_negate`, `keys.prvkey_tweak_add`,
`keys.prvkey_tweak_mul`, `xonly.prvkey_tweak_add`, `ecdh.shared_secret`
and its private half, `ellswift.xdh`, `dsa.nonce_rfc6979`,
`ssa.nonce_bip340`.

**Left out: `silentpayments.label`**, whose secret is one half of a
returned pair — `into` there would name which half in the argument list
and nowhere else. Stated rather than guessed at; say the word and it
goes in.

## The cost, which is in the type system

Two `@overload`s per entry point, so omitting `into` still types as
`bytes` and a downstream `mypy --strict` sees no change. The one visible
consequence: four negative-typing tests moved from
`type: ignore[arg-type]` to `[call-overload]`, which is what an
overloaded call reports. No coverage configuration change was needed —
overload stubs are already not counted.

## Gate

`uvx pre-commit run --all-files` clean, 564 tests, coverage 100.00%.

## Summary by Sourcery

Add an optional keyword-only `into` buffer parameter to secret-producing APIs so callers can receive secrets in mutable memory instead of new bytes objects, and update internals, typing, tests, and documentation accordingly.

New Features:
- Allow `_secret.take` and all public secret-producing functions to write secrets into a caller-provided mutable buffer via a keyword-only `into` parameter that suppresses the bytes return value.

Enhancements:
- Introduce a `MutableBytesLike` type alias and a shared `_secret.into_buffer` helper to validate and normalize caller-provided secret buffers before writing.
- Extend type hints with overloads so existing call sites still type as returning `bytes` when `into` is omitted, while supporting `None` returns when `into` is used.

Documentation:
- Document the new `into` mechanism and its security properties in SECURITY.md and CHANGELOG, clarifying caller responsibilities for wiping buffers.

Tests:
- Add and adjust tests to cover `_secret.take` buffer semantics, the new `into` parameter across all secret-producing functions, and updated typing expectations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -421,6 +421,85 @@ release-notes length in the first place, and are still in
   None` is what says so, and two tests in `tests/test_callbacks.py` hold
   the pair, each verified against the mutant it exists to kill
 
+### The copy a caller can take back
+
+- **every producer of a secret takes `into`** (#188), a keyword-only
+  writable buffer the secret is written to instead of a `bytes`:
+  `keys.prvkey_negate`, `prvkey_tweak_add` and `prvkey_tweak_mul`,
+  `xonly.prvkey_tweak_add`, `ecdh.shared_secret` and its private half,
+  `ellswift.xdh`, `dsa.nonce_rfc6979` and `ssa.nonce_bip340`. Omitted,
+  nothing changes and the `bytes` comes back as before. Given, nothing
+  is returned, and the copy the caller ends up holding is one they can
+  overwrite — which is the entire claim, and SECURITY.md now states it
+  beside the limitation it narrows rather than instead of it. It does
+  not wipe the buffer for them; a buffer never overwritten is exactly
+  the copy `into` was reached for to avoid, and the obligation is
+  stated rather than enforced for the reason `ssa.Signer` has no
+  finalizer
+- **which is not what #87 declined.** That issue asked whether an opaque
+  secret *handle* buys assurance and answered no: a handle is a lifetime
+  someone has to own and invalidate, the same shape of requirement that
+  keeps MuSig2 out, and its entry side still starts from a `bytes` the
+  caller already made. Both halves of that stand. What is new is neither
+  a lifetime nor an entry: one keyword argument, one call, no object,
+  and it addresses the *exit* — where `_secret.take` had nowhere to put
+  the secret except an immutable object nothing can zero
+- **`into` is keyword-only, and that is not a style choice.** Passed
+  positionally to `dsa.nonce_rfc6979` it would land on `aux_rand32`,
+  which takes 32 octets and would have accepted the empty buffer as
+  entropy — a wrong nonce, silently, from a call that reads right. The
+  suite caught it on the way in; keyword-only is what keeps a caller
+  from meeting it
+- **and the buffer is proved before anything is copied, the wipe
+  happening either way.** A wrong length is refused rather than filled:
+  too short raises on the assignment anyway, but too *long* would take
+  the secret in its first octets and leave whatever was behind it, which
+  reads as a secret with a tail. A read-only view cannot be wiped, so
+  accepting one would defeat the facility silently. And a view of more
+  than one dimension passes every one of those checks and then fails the
+  copy with `NotImplementedError`, while a strided one succeeds and
+  leaves the secret scattered through an owner twice its size, so both
+  are refused as not contiguous octets. Beyond that the buffer protocol
+  decides rather than a type: `mmap` and `array.array("B")` are taken,
+  an `mlock`ed `mmap` being a plausible destination for exactly this
+  caller, and `memoryview` itself is what refuses a non-buffer. That
+  breadth is the run time's rather than the annotation's, which stays
+  the two named types until `collections.abc.Buffer` is reachable at a
+  3.12 floor: a caller under `mypy --strict` passes the rest as
+  `memoryview(x)`, and SECURITY.md says so where the mismatch is met.
+  `_secret.into_buffer` is annotated `object` and not
+  `MutableBytesLike`, that being the annotation it exists to not trust
+- **the wipe is in a `finally`, which is the defect this facility could
+  have shipped with.** `take` had no failure mode before `into`, and
+  with one on the straight line a refused buffer returned through the
+  entry point leaving a live private key in cffi memory that is freed
+  without being overwritten -- none of the eight has a `finally` of its
+  own, and `xonly.prvkey_tweak_add`'s wipes the keypair rather than the
+  key. Nothing is lost by wiping on the way out: the value never reached
+  the caller, and the call they retry recomputes it
+- **the two secrets `silentpayments` answers are left out**, each being
+  one member of a returned tuple where an argument could not say which:
+  the tweak of `label`, and the per-output tweak `scan_outputs` hands
+  back. SECURITY.md names both rather than claiming every entry point
+- **and what holds the set is a walk rather than a list.** A hardcoded
+  tuple of eight cannot notice a ninth, which is the whole point of the
+  test; `tests/test_secret.py` walks the package for functions whose
+  source calls `_secret.take` and requires `into` of each, one
+  exemption named. Its limit is written down beside it: a secret that
+  never goes through `take` is invisible to it, which is exactly how
+  `_found_output` escaped the first draft of that sentence
+- **the cost, which is in the type system rather than at run time.**
+  Three `@overload`s per entry point, so that omitting `into` still
+  types as `bytes` and a downstream `mypy --strict` sees no change; the
+  implementation signature is the union. The third is the one a caller
+  needs to *forward* an optional buffer through a wrapper of their own,
+  which is how btclib would surface this argument, and with only two it
+  is that caller who writes the two-branch version or a `cast`. Four
+  negative-typing tests moved from `type: ignore[arg-type]` to
+  `[call-overload]`, which is
+  what an overloaded call reports, and is the one way this is visible
+  to a caller who was already type-checking
+
 ### One statement of each thing
 
 - **the duplications are gone**, each of them two spellings of one
--- a/README.md
+++ b/README.md
@@ -534,6 +534,15 @@ private keys, and the x-only public key itself, from a full public key
 `hashes` provides the BIP340 tagged hash, the domain separation the
 taproot tags are built on.
 
+Wherever one of these answers a *secret* — a tweaked or negated private
+key, a shared secret, a nonce — it also takes a keyword-only `into`: a
+writable buffer of exactly 32 contiguous octets, which receives the
+secret in place of the `bytes` the call would otherwise return, so that
+the copy the caller is left holding is one they can overwrite. It is an
+addition and not a change; omit it and nothing differs. SECURITY.md is
+where what it does and does not buy is stated, and names the two
+`silentpayments` secrets it does not reach.
+
 Two of those have a second spelling for a caller doing arithmetic rather
 than holding a key. `keys.pubkey_sum` is `pubkey_combine` with the point
 at infinity answered as `None` instead of refused: `P + (-P)` is the
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -77,6 +77,43 @@ These are known and inherent, not vulnerabilities:
     and the buffer zeroed before it is dropped. That is one copy taken back,
     not safety: the `bytes` handed to the caller holds the same secret
     and cannot be overwritten
+- **`into` moves that last copy somewhere the caller can overwrite,**
+    and is the whole of what it does. Eight entry points take a
+    keyword-only `into` — a writable buffer of exactly the secret's
+    length, contiguous and one octet an item, `bytearray` and
+    `memoryview` and `mmap` and `array.array("B")` alike — write the
+    secret there and return None, so no `bytes` of it is ever made.
+    That breadth is the run time's: the annotation is a `bytearray` or a
+    `memoryview`, `collections.abc.Buffer` being python 3.12 where the
+    floor here is 3.10, so a caller running `mypy --strict` passes
+    anything else as `memoryview(x)`, which copies nothing. The eight
+    are `keys.prvkey_negate`, `keys.prvkey_tweak_add`,
+    `keys.prvkey_tweak_mul`, `xonly.prvkey_tweak_add`,
+    `ecdh.shared_secret`, `ellswift.xdh`, `dsa.nonce_rfc6979` and
+    `ssa.nonce_bip340`. **The two secrets `silentpayments` answers do
+    not**, each being one member of a returned tuple, where an argument
+    could not say which: the tweak of `label`, and the per-output tweak
+    `scan_outputs` hands back. Both are `bytes` and neither can be
+    zeroed, which is the limitation above and not this narrowing of it.
+    What the caller then does with the buffer is theirs: this does not
+    wipe it for them, and a buffer they never overwrite is exactly the
+    un-zeroizable copy `into` was reached for to avoid. Nor does it
+    touch the *entry* side, which is the half that cannot be improved
+    from inside this package: the `bytes` or `int` handed in already
+    existed before the call, and the arithmetic that produced it
+    happened where this package cannot see. The obligation is stated
+    rather than enforced, for the reason `ssa.Signer` has no finalizer:
+    a guarantee nothing keeps is worse than none
+
+    ```python
+    prvkey = bytearray(32)
+    keys.prvkey_tweak_add(parent, tweak, into=prvkey)
+    try:
+        ...                       # use it
+    finally:
+        prvkey[:] = bytes(32)     # the caller's wipe, and only theirs
+    ```
+
 - the one buffer whose zeroing is the caller's to ask for is
     `ssa.Signer`'s. Everything above is wiped inside the call that made
     it — read out and zeroed in the one operation, or wiped in a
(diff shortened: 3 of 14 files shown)


fametrano:
## Review: the argument is right, and one refusal path drops a secret

**Verdict: changes requested.** The design holds — keyword-only for the
`aux_rand32` reason, proof before the copy, no finalizer, overloads so a
downstream `mypy --strict` sees nothing. One finding is a defect in the
facility itself, the rest are smaller.

### 1. A refused `into` leaves the secret in a buffer nothing wipes

`_secret.take` proves and copies before `wipe(buffer)`, so any refusal
returns through the entry point with this package's buffer still holding
the secret — and no entry point wipes it:
`keys.prvkey_negate`, `prvkey_tweak_add`, `prvkey_tweak_mul`,
`dsa.nonce_rfc6979`, `ssa.nonce_bip340`, `ecdh._shared_secret_` and
`ellswift.xdh` have no `finally`, and `xonly.prvkey_tweak_add`'s wipes the
keypair, not `prvkey_buffer`. So

```python
keys.prvkey_tweak_add(prvkey, tweak, into=bytearray(31))   # ValueError
```

raises with a live private key in cffi memory that is freed without being
overwritten. `test_take_refuses_a_buffer_of_the_wrong_length` asserts
exactly that (`== SECRET`), so the branch locks the behaviour in rather
than leaving it unnoticed.

The stated reason — "a refusal has to leave this package's buffer holding
the secret, so that the wipe below is the only thing that ever empties
it" — keeps the letter of the pair and inverts its purpose. The value is
unreachable to the caller on both paths, so wiping on refusal loses
nothing a caller can want: the operation failed, and calling again with a
buffer of the right length recomputes it. The `finally` is what makes the
read and the wipe one operation, not the straight line:

```python
    try:
        view = into_buffer(into, len(memory))
        view[:] = memory
    finally:
        wipe(buffer)
    return None
```

Before this branch `take` had no failure mode, so this is a new way for a
secret to survive a call — in the one facility whose entire claim is about
where copies of secrets live. It is the finding I would not merge without.

### 2. A memoryview of `ndim > 1` passes `into_buffer` and fails in the copy

Not writable, not the wrong length, not wider items, and still wrong:

```python
>>> into = memoryview(bytearray(32)).cast("B", (4, 8))
>>> into.readonly, into.itemsize, into.nbytes
(False, 1, 32)
```

All four checks pass, and `view[:] = memory` then raises
`NotImplementedError: memoryview slice assignments are currently
restricted to ndim = 1` — neither of the two types the docstring's
`Raises` names, and not a message that says what the caller did. I
reproduced it against a stand-in for `take`. One line closes it:

```python
    if view.ndim != 1:
```

Worth deciding beside it: a strided view is accepted today.
`memoryview(bytearray(64))[::2]` is writable, one octet an item and 32
octets long, so the secret lands scattered over 64 octets of the owner. It
writes and wipes correctly through that same view, so it may well be
intended — but `ndim != 1` does not catch it and `view.c_contiguous` does,
and the docstring's "in all three ways it can be wrong" would then be
four.

### 3. "Every entry point above" is an absolute claim with two exceptions

SECURITY.md: "Every entry point above that answers a secret takes a
keyword-only `into`". `silentpayments.label` is one exception, named in
CHANGELOG.md but not there. The second is named nowhere:
`silentpayments._found_output` reads the per-output tweak — a secret, as
its own docstring says — out of the struct as a `bytes` without going
through `take`, and `scan_outputs` hands it to the caller. Both belong in
that sentence, or the sentence should not say "every".

### 4. `isinstance(into, (bytearray, memoryview))` refuses writable buffers

`mmap` and `array.array("B")` are both writable and both refused, and an
`mlock`ed `mmap` is a plausible destination for exactly the caller this
argument is written for. The `readonly` check already refuses `bytes` and
read-only views, so the isinstance gate narrows the facility without
catching anything the checks below it miss; `memoryview(into)` raises
`TypeError` on a non-buffer by itself. The one thing it buys is the better
message for `bytes` — cheap to keep by catching `TypeError` around the
`memoryview` call and re-raising with the same text.

### 5. The set test cannot catch what its docstring claims

> a producer added without `into` is what it is meant to catch

A hardcoded tuple of eight cannot catch a ninth. What would: walk the
package's modules for functions whose source calls `take(` and assert each
has `into` in its signature — nine call sites today, which is also the
check that would have named `silentpayments`. Otherwise the sentence
should say what the test does, which is that the two spellings agree.

### 6. Rebase

`DIRTY` on CHANGELOG.md; main has moved three times since this branched.

fametrano:
## Second pass: no new commit, and one thing the overloads keep for themselves

**Verdict unchanged: changes requested.** The secret left unwiped on a
refused `into` is still the one I would not merge without; below is what
this pass adds.

### The third overload is written for the internal caller and not the external one

`_secret.take` and `ecdh._shared_secret_` get three overloads; the eight
public entry points get two. The third is exactly the one a caller needs
to forward an optional buffer through a wrapper of their own, which is how
btclib would surface this argument:

```python
def negate(prvkey: bytes, into: MutableBytesLike | None = None) -> bytes | None:
    return keys.prvkey_negate(prvkey, into=into)
```

`mypy --strict`, against a stand-in with this branch's two overloads:

```
error: Argument "into" to "prvkey_negate" has incompatible type
"bytearray | memoryview[int] | None"; expected "bytearray | memoryview[int]"
[arg-type]
```

So the caller writes the two-branch version, or an `# type: ignore`, or a
`cast`. `shared_secret` did not have to, because `_shared_secret_` was
given the overload that lets it forward — the asymmetry is visible inside
the diff. Four lines per entry point closes it, and the pull request
already argues that the cost of this feature is meant to be in the type
system and not at run time. If the asymmetry is deliberate, it is worth a
sentence saying a caller forwards `None` by not passing the argument.

### From main rather than from this branch

`0d87af68` took the CHANGELOG.md finding out of `.secrets.baseline` with a
`pragma: allowlist secret`. The baseline hunk here edits a `line_number`
that no longer exists, so on the rebase it is dropped rather than
resolved, and `.secrets.baseline` should leave the diff entirely.

### Minor, and only if you agree

README.md's "Wrapped modules" describes each module in prose and names
what it answers; `into` is now a keyword on nine entry points across six
of those modules and is described only in SECURITY.md and the docstrings.
One clause there would be where a reader meets it.

### Still open from the first pass

1. a refused `into` leaves the secret in this package's buffer, unwiped,
   on every one of the eight entry points — `try: … finally: wipe(buffer)`.
2. a `memoryview` of `ndim > 1` passes `into_buffer` and fails the copy
   with `NotImplementedError`.
3. SECURITY.md's "Every entry point above" has two exceptions,
   `silentpayments.label` and the tweak `scan_outputs` hands out.
4. `isinstance(into, (bytearray, memoryview))` refuses `mmap` and
   `array.array("B")`, which the `readonly` check does not need it to.
5. the set test cannot catch the ninth producer its docstring promises.

fametrano:
[`24269e7`](https://github.com/btclib-org/btclib-secp256k1/commit/24269e7)
(#161) and [`7944dce`](https://github.com/btclib-org/btclib-secp256k1/commit/7944dce)
(#176): **this branch reads CONFLICTING now, and both conflicts are
mechanical.**

- `.secrets.baseline` — the entry for `CHANGELOG.md` is a line number,
  and two entries were appended above it. Do not merge that hunk by
  hand: take `main`'s file and let the hook regenerate it,

  ```shell
  git checkout --ours .secrets.baseline   # during the rebase, "ours" is main
  git add .secrets.baseline
  # finish the rebase, then
  uvx pre-commit run detect-secrets --all-files && git add .secrets.baseline
  ```

- `CHANGELOG.md` — `.gitattributes` gives the union driver to
  `HISTORY.md` only, so two additions to the same section are a conflict
  rather than a merge. Both of mine are in v0.8.0.3's *Documentation*
  section; a branch adding under a different heading will conflict only
  on the context lines.

**And the landing rule changed while you were writing this.** REPOSITORY.md,
CONTRIBUTING.md and RELEASING.md now say a pull request lands as a
fast-forward, with the squash made locally where the branch carries more
than one commit — the button being what signs the commit with GitHub's
web-flow key. The practical part for you: **push the squash to the branch
before fast-forwarding it**, and GitHub still marks this Merged and deletes
the head branch itself. Squash straight from a worktree and nothing is
reconciled, so `Closes #N` has to be in the commit message.

fametrano:
All six taken, plus the second pass's two, in `6051e3e`.

**1. The refused buffer left a secret unwiped — correct, and it was mine.** `take` had no failure mode before this branch, so putting one on the straight line was the whole of the mistake. The wipe is in a `finally`, and the reasoning in the comment is replaced rather than patched: the value never reached the caller, so keeping it costs them nothing. `test_take_refuses_a_buffer_of_the_wrong_length` now asserts the zeroed buffer — the assertion that locked in the defect is the one that holds the fix. Verified against the mutant: with the `finally` removed, four tests in that file fail.

**2. `ndim > 1`.** Reproduced exactly as described. Refused, and so is the strided view, for the reason given — a wipe of the owner would look right and not be. One check, `ndim != 1 or not c_contiguous`, because `cast("B", (4, 8))` *is* `c_contiguous` and `[::2]` is not, so neither test catches the other. The docstring's "three ways" is gone; it now says what is required.

**3. "Every entry point above" had two exceptions.** Both named in SECURITY.md, with why an argument could not say which member of a tuple it fills.

**4. The isinstance gate.** Dropped; `memoryview(into)` refuses a non-buffer by itself, and the `bytes` message is kept by catching that `TypeError`. `mmap` and `array.array("B")` are taken, with a test. One wrinkle worth your eye: `MutableBytesLike` cannot express that — `collections.abc.Buffer` is 3.12 and the floor is 3.10 — so a typed caller passes those as `memoryview(x)`, which copies nothing. Said in the alias's comment and in the test.

**5. The set test could not catch a ninth.** Replaced by the walk you describe: functions whose source calls `_secret.take`, one exemption named. It found what the list could not — and it also found the *limit* of that population, which is now written beside it: `_found_output` never calls `take`, so no walk defined this way contains it. That is exactly how it escaped my first draft of SECURITY.md's sentence. Verified by removing `into` from `ellswift.xdh`: the test names it.

**Second pass — the third overload.** Taken on all eight. The asymmetry was accidental: `_shared_secret_` got it because `shared_secret` needed to forward, and nothing made me ask who else would.

**Second pass — README.** Taken; one paragraph in "Wrapped modules", where a reader meets the modules.

**Baseline and rebase.** `.secrets.baseline` is out of the diff entirely.

Gate clean, 567 tests, coverage 100.00%.

fametrano:
## Third pass, on `6051e3e`: ACK, once the coverage job reports

All eight findings taken, and the two that mattered are fixed rather than
documented.

**The wipe is in a `finally`**, and the test that asserted the leak now
asserts the wipe — including on the refusals that reach it through
`into_buffer`, which the other three refusal tests each check too. That
was the one I would not have merged without.

**Contiguity is refused as one condition** (`ndim != 1 or not
c_contiguous`), which takes both shapes: the two-dimensional view that
used to reach `NotImplementedError`, and the strided one that used to
succeed and scatter the secret through an owner twice its size. The
docstring says why the second is refused even though it works, which is
the part a later reader would otherwise undo.

**The buffer protocol decides**, so `mmap` and `array.array("B")` are
destinations now, and `memoryview` itself is what refuses a non-buffer —
with the `# type: ignore[arg-type]` carrying the reason the annotation is
`object`.

**The third overload is on all eight**, so a caller can forward an
optional buffer through a wrapper of their own; that was the asymmetry
between what the internal callers had and what the external one got.

**The walk replaces the list**, and what it cannot see is written down
beside it rather than left to be discovered — which is how
`_found_output` got into SECURITY.md, where "every entry point" is now
eight named ones and two named exceptions.

**Gates**: lint and type-check and the documentation build are green on
`91ef185`; `.secrets.baseline` is out of the diff; the branch is `BEHIND`
rather than `DIRTY`. The coverage job is still queued, and it is the one
that has anything new to say — the walk's `pragma: no cover` and the four
refusal branches are what it will report on. ACK conditional on it, and
on nothing else.

Two commits, so a local squash and a fast-forward push.

### Two notes, neither blocking

1. SECURITY.md now says the argument takes "`bytearray` and `memoryview`
   and `mmap` and `array.array("B")` alike", and that is true of the run
   time and not of the annotation: `MutableBytesLike` is the two, so a
   caller running `mypy --strict` passes the other two as
   `memoryview(x)`. `__init__.py`'s comment says exactly this, in the
   file that caller does not read. One clause in SECURITY.md would put it
   where the mismatch is met.
2. The walk's population is the call sites of `_secret.take`, so a
   *public* entry point that delegates to a private producer is not in
   it: `ecdh.shared_secret` is that shape today, and it is the second
   thing the walk cannot see. The docstring records the first blind spot;
   this one costs it a sentence.

PR #185, opened 2026-08-16
Ask where the walk starts, which is what the three numbers leave out

fametrano:
The outpost section measures three spellings of a five-step BIP32 walk.
This asks the question none of them does — **where the walk starts** —
and drops a claim that does not survive being measured.

## What the branch first said, and why it was wrong

That `chain.tweak_add(compressed=False)` is the fast spelling of four.
It is the slowest of three. A chain re-parses nothing in either
serialization, the point being what it holds, so asking it for 65 octets
saves no parse and pays a serialization and the caller's own compression
line at every step. Measured here — five steps, medians of nine
alternating rounds, sha256 control, absolutes about 5% above the
section's — 59.27 for the composition against 58.62 for
`chain.tweak_add()` and 58.26 for the uncompressed walk. The section's
own numbers predicted a loss of 0.98, one compressed parse in against
five uncompressed parses out; the measured one is 1.01.

## What replaces it

The uncompressed walk's 54.75 starts from 65 octets a BIP32 caller does
not have. An xpub carries the 33, so reaching that form costs
`keys.reserialize` once — a compressed parse and a serialization, more
than the 0.39 that separated the two walks. Counted, the same run puts
that walk at 61.55 against the chain's 58.62.

| five steps, from the 33 octets an xpub carries | µs |
| --- | --- |
| compressed walk | 68.15 |
| uncompressed walk, `reserialize` counted | 61.55 |
| `chain.tweak_add()` | **58.62** |
| `chain.tweak_add(compressed=False)` | 59.27 |

So the chain is *ahead* of the spelling the section calls it a rounding
error behind, on the key its caller actually holds — the parse it pays
at construction being the one that caller could not have avoided.

That is a better answer to #161 than the claim was: the class rests on a
measurement rather than on the cost of breaking a released caller.
Documentation only; the API is untouched.

Closes #161

Issue #153, opened 2026-08-15
BIP340 signing rebuilds the keypair on every call, and no sign_ takes one already built


fametrano:
## The finding

In btclib-benchmarks' wrappers table, `ssa.sign` runs about 2x behind
secp256k1-py's `schnorr_sign` for the same one signature. None of that gap
is the vendored libsecp256k1 revision: it is `secp256k1_keypair_create`,
paid on every call.

Decomposed on an Apple M5, CPython 3.13.14, btclib-secp256k1 0.8.0.2 —
minimum of 10 rounds of 20k calls each, microseconds per call:

| step | µs/call |
|---|---|
| `ssa.sign`, as shipped: keypair per call, real aux | 15.5 |
| `secp256k1_keypair_create` alone | 7.3 |
| `secp256k1_schnorrsig_sign32` over a prepared keypair | 8.0 |
| argument validation (`scalar` + `octets`) | 0.2 |
| `wipe` of the keypair | 0.1 |

The pieces sum to the whole, and the keypair is half of it. secp256k1-py
posts 7.7 µs for the same operation because its `PrivateKey` constructor
builds the keypair once and its `schnorr_sign` reuses it. The signing
itself is at parity — 8.0 through this package's v0.8.0 against 7.7
through the 2021 master secp256k1-py vendors — so the C library is not the
lever; the keypair rebuilt per call is.

## The asymmetry

The verify side already has the split this needs, since #147: `keys.parse`
and `xonly.parse` produce the parsed object once, and every `verify_`
takes it in place of bytes, so a caller verifying many signatures under
one key pays the parse once. The sign side has no analogue: `ssa.sign` and
`ssa.sign_custom` build the keypair from bytes inside the call and wipe it
in a `finally`.

A `sign_` over a keypair the caller prepared once would roughly halve the
per-signature cost for anyone signing repeatedly under one key. BIP340
only: `secp256k1_ecdsa_sign` takes the secret key directly, so `dsa.sign`
has no keypair to hoist.

## The trade-off it has to name

The wipe-in-`finally` is a policy, not an accident: today no secret
outlives the call that needed it, and a keypair is the private key in a
cffi struct. A keypair-holding API hands that lifetime to the caller, so
it belongs with the discussion #87 opened about opaque secret objects: an
opaque keypair whose docstring says what the caller now owns, wiped
deliberately — a `wipe()` method, or a context manager — rather than
implicitly. If the trade is unwanted, this issue is the place to say no
[...]

PR #197, opened 2026-08-16
State the test a third outpost would have to pass, where the two are

### LINKED ISSUE #191: The rule #161 decided by lives in an issue: state it in the README as the test a proposed stateful object has to pass
Issue #161 states the rule this package judges a stateful object by: one
earns its keep where what it caches is arithmetic, not a parse. It is
what decided `PubkeyTweakChain`'s fate, and it is the answer to the next
person who proposes a held public key for `verify`, `ecdh` or `combine`.

It is written in an issue. The README explains the two objects that
exist — "Two outposts past the boundary" separates them by exactly this,
"A keypair is *arithmetic* ... A parsed public key is a *parse*" — but it
explains the two that are here rather than stating the test the third one
has to pass. That test is the part a reader needs and the part that is
not in the tree.

## The rule, in the form the measurement supports

"Arithmetic, not a parse" is the shorthand. The criterion under it is
sharper, and it can be checked rather than asserted:

> **The saving an object offers is the cost of rebuilding, from octets,
> the thing it holds.** An object is worth its state when that cost is
> worth it; a parse is not, because the caller can already pay it once
> through `parse` and hand the object on.

Rebuilding, measured on an Apple M5, macOS 26.6, arm64, CPython 3.14.6,
minimum of 11 rounds:

| what an object would hold | cost of rebuilding it from octets | us |
| --- | --- | --- |
| a point, from 65 octets | `keys.parse` | 0.27 |
| a point, from 33 | `keys.parse`, a field square root | 2.40 |
| an x-only key, from 32 | `xonly.parse`, a field square root | 2.42 |
| a keypair, from its 32-octet secret | `secp256k1_keypair_create`, a point multiplication | 7.65 |

And the identity holds where it can be checked directly:

| | us |
| --- | --- |
| `ssa.sign(msg, prvkey)` | 16.67 |
| `ssa.Signer(prvkey).sign(msg)` | 9.01 |
| difference | **7.66** |

which is the 7.65 of the keypair rebuild in the row above. What `Signer`
saves is exactly what rebuilding what it holds would cost. Nothing else
about it is doing any work, and the same arithmetic run on
`PubkeyTweakChain` gives 0.27 — which is why #161 reached the conclusion
it did.

## The three things the README statement has to get right

These are where the shorthand is thin, and each will otherwise be
rediscovered as an objection.

**A parse is recoverable, and what recovers it is `parse`, not a
serialization.** The README argues the point through the uncompressed
form — 0.269 against 2.326, so carry 65 octets. That argument does not
reach x-only, which has no second serialization: 2.42 microseconds to
rebuild, and no cheaper octets to carry. The rule survives anyway, and
for the reason that matters: the escape hatch is the public
`parse`/`serialize` bridge that #161 deliberately kept, so a caller who
wants the parse held holds it, without this package growing a class. Say
that, or the x-only number reads as a counterexample.

**Statelessness is the other side of the ledger, and it is the same
argument that keeps MuSig2 out.** The README already states it there —
this package is stateless by construction, so a lifecycle belongs where
the signing state lives. An object is an exception to that, with an
owner, an invalidation and a threading story, and the rule is what the
exception has to be paid for with. The two paragraphs should point at
each other; at present only the MuSig2 one exists.

**Ergonomics is out of scope, and by a decision rather than an
oversight.** #161 concedes what the chain buys is "the line a caller does
not write", and rules it insufficient. That is right for a thin binding
whose stated job is the wrapping and not the interface, but it is a
judgement, and unstated it looks like the rule simply forgot to weigh
convenience. The place for convenience is btclib, same as the lifecycle.

One more, on how much weight the numbers carry: the walk that decided
`PubkeyTweakChain` was 55.14 against 54.75, which is 0.7% and is noise on
another machine. The measurement *confirmed* the rule; what establishes
it is the structural fact that a parse has a public bridge and arithmetic
has none. Written the other way round, the rule inverts the first time
someone re-measures on different hardware.

## Where it goes

In "Two outposts past the boundary", as the paragraph the two examples
are then instances of, rather than a conclusion drawn after them. The
section already carries every number it needs except the keypair rebuild.

## Acceptance

- the rule stated as a test a proposed object must pass, in the README,
  not only in this issue and #161
- x-only named, so the one parse with no cheaper serialization is
  answered where a reader will look for it
- the statelessness paragraph and the MuSig2 paragraph cross-referring
- the `links` workflow still green, and the README doctests still pass

fametrano:
Closes #191. Documentation only; the API is untouched.

"Two outposts past the boundary" explains the two objects that exist and
separates them by the right thing. The reader who needs that distinction
is the one proposing a **third** — a held public key for `verify`, for
`ecdh`, for `combine` — and what they need is the *test*, not the two
instances of it. The test lived in #161.

## What is added

The rule, as one sentence:

> what an object saves is the cost of **rebuilding, from octets,
> whatever it holds**. It earns its state when that cost is arithmetic,
> and does not when it is a parse.

And the reason it can be checked rather than argued: the saving is not
merely bounded by that cost, it **is** that cost.

| Apple M5, macOS 26.6, arm64, CPython 3.14.6, min of 11 rounds | µs |
| --- | --- |
| `ssa.sign(msg, prvkey)` | 16.67 |
| `ssa.Signer(prvkey).sign(msg)` | 9.01 |
| the difference | **7.66** |
| `secp256k1_keypair_create` alone | **7.65** |

The same subtraction on a parsed point gives 0.27. The section already
carried every number it needed except that keypair rebuild, which is
what turns the distinction into an identity.

## Two objections answered before they are raised

**What recovers a parse is `keys.parse`, not a cheaper serialization.**
The section argues through the uncompressed form, 0.269 against 2.326 —
an argument that does not reach an x-only key, 2.42 microseconds to
rebuild with no second serialization to escape to. Unsaid, that number
reads as a counterexample. The rule holds on the other leg: the public
parse/serialize bridge is the escape, and needs no class beside it.

**Statelessness is the other side of the ledger**, and it is the same
argument that keeps MuSig2 out. The two paragraphs now point at each
other; before, only the MuSig2 one existed.

Also said, because the difference matters: the numbers **confirm** the
rule, they do not establish it. A rule resting on which of two
measurements is larger inverts on the first machine that measures them
differently — and one of the pair #161 weighed was 55.14 against 54.75.

## Overlap with #185

#185 is open on this same section and refines the chain's case from the
other direction. The hunks do not touch: this inserts at the rule
paragraph, #185 after the chain example. Whichever lands second may want
one cross-reference — #185's finding, that the chain is *ahead* when the
walk starts from the 33 octets an xpub carries, is a sharper instance of
"what recovers a parse is not always cheap" than the x-only case used
here, and the two paragraphs would read well together.

## Gate

`uvx pre-commit run --all-files` clean, 558 tests pass (README doctests
included).

## Summary by Sourcery

Document the rule that determines when a stateful object is justified and integrate its explanation and supporting benchmarks into the README and changelog.

Documentation:
- Clarify the distinction between stateful keypair holders and parsed points by stating a general rule and test for proposed third stateful outposts in the README, supported by performance measurements and examples.

Chores:
- Record the new rule and its rationale in the changelog, including how it applies to stateful objects, parses, stateless design, and the role of benchmarks in confirming rather than establishing the rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -500,6 +500,48 @@ release-notes length in the first place, and are still in
   what an overloaded call reports, and is the one way this is visible
   to a caller who was already type-checking
 
+### The test a stateful object has to pass
+
+- **the rule the outpost section applies is now stated there** (#191),
+  rather than only in <https://github.com/btclib-org/btclib-secp256k1/issues/161>
+  where it was reached. That section explained the two objects that
+  exist and separated them correctly — a keypair is arithmetic, a parsed
+  point is a parse — but the reader who needs it is the one proposing a
+  *third*, and what they needed was the test, not the two instances of
+  it. It is written as one sentence: what an object saves is the cost of
+  rebuilding, from octets, whatever it holds
+- **and the saving is that cost by construction, not by measurement,**
+  which is the stronger form of the same claim and the one the section
+  now makes. `ssa.sign` builds the keypair, signs with it and wipes it;
+  `signer.sign` is that middle term with the first already paid, so the
+  difference between them is `secp256k1_keypair_create` on any machine,
+  and the numbers only add that nothing else of consequence sits in it.
+  They are the section's own — 15.82 against the signer's 8.27, of which
+  the keypair is 7.55, and 15.82 − 8.27 is that 7.55 — rather than a set
+  of its own. The chain gets the same treatment: what it saves is the
+  parse it holds, and which parse that is depends on the serialization
+  its caller carries — the pair the section states four paragraphs
+  above, pointed at rather than restated. That is the sharper form of
+  the distinction, because it says why this saving is the one a caller
+  can make cheap and the keypair's is not
+- **and which figures survive a different machine is said**, because a
+  section arguing from microseconds should say which of them it is
+  arguing from: the identity, and it carries because it is structural,
+  not because a ratio travels better than an absolute. The counts are
+  one laptop's
+- **two objections are answered before they are raised.** What recovers
+  a parse is `keys.parse`, not a cheaper serialization: the
+  uncompressed-form argument reaches neither an x-only key, which has no
+  second serialization, nor a walk starting from the 33 octets an xpub
+  carries, where reaching the uncompressed form costs a `reserialize`
+  the chain does not pay — which is the finding of
+  <https://github.com/btclib-org/btclib-secp256k1/pull/185>, now in the
+  same section and pointed at from here. And statelessness is the other
+  side of the ledger, the same argument that keeps MuSig2 out, so the
+  two paragraphs point at each other. Ergonomics is weighed by neither,
+  deliberately: the line a caller does not write belongs in btclib with
+  the lifetimes
+
 ### One statement of each thing
 
 - **the duplications are gone**, each of them two spellings of one
--- a/README.md
+++ b/README.md
@@ -400,6 +400,60 @@ carrying the uncompressed form instead: 0.269 microseconds against
 2.326. So the signer saves what nothing else can, and the chain saves
 what a caller free to choose its serialization could have saved itself.
 
+**The rule that separates them is the test a third outpost would have to
+pass**, and it is worth stating on its own, because the answer to every
+proposal of one — a held public key for verifying, for ECDH, for
+combining — is here rather than in the proposal:
+
+> what an object saves is the cost of *rebuilding, from octets, whatever
+> it holds*. It earns its state when that cost is arithmetic, and does
+> not when it is a parse.
+
+The saving is not merely bounded by that cost; it *is* that cost, and by
+construction rather than by measurement. `ssa.sign` builds the keypair,
+signs with it and wipes it; `signer.sign` is that middle term with the
+first already paid — so the difference between them is
+`secp256k1_keypair_create` on any machine, and what the numbers below
+add is that nothing else of consequence sits in it. They are the ones
+this section already states: 15.82 microseconds against the signer's
+8.27, of which the keypair is 7.55, and 15.82 − 8.27 is that 7.55.
+
+The chain reads the same way and answers differently, which is the
+distinction of the paragraph above arrived at from the other end. What
+it saves is the parse it holds, so *which* parse decides the figure —
+and that is the pair the paragraph above states, the caller's to choose
+by the serialization they carry. A keypair offers its caller no such
+choice.
+
+Two things the rule needs said, or it reads as narrower than it is. The
+first is what recovers a parse, and it is **`keys.parse` itself, not a
+cheaper serialization**. The argument through the uncompressed form does
+not reach an x-only key, which is a field square root to rebuild and has
+no second serialization to escape to — and it does not always reach a
+full one either, which is what the walk below measures: from the 33
+octets an xpub carries, reaching the uncompressed form costs a
+`reserialize` that the chain does not pay. A caller who wants the parse
+held holds it — `keys.parse` hands back the object and the private
+halves take it, which is what "the private halves above hand one object
+from a call to the next" already promised. What a class would add there
+is a name, not a saving. The second is the other side of the ledger:
+this package is [stateless by construction](#design), which is also why
+MuSig2 is not
+wrapped, so an object is an exception to that, with an owner and an
+invalidation and a threading story, and the rule is what the exception
+has to be paid for with. What it does not weigh is ergonomics — the line
+a caller does not have to write is real, and belongs in btclib along
+with the lifetimes.
+
+So the numbers **confirm** the rule rather than establish it, and it is
+worth being clear which of them would survive a different machine. Not
+the microsecond counts, which are one laptop's; the identity is what
+carries, and it carries because it is structural rather than because a
+ratio travels better than an absolute. A rule resting instead on which
+of two measurements is larger would invert on the first machine that
+measured them differently — and one such pair is a rounding error apart,
+which the walk below is about.
+
 `ssa.Signer` holds the keypair:
 
 ```python


fametrano:
## Review: the rule is worth stating, the table under it is not this one

**Verdict: changes requested.** Documentation only, and the prose is the
product here, so the numbers in it are the review.

### 1. The section already states these three quantities, differently

Below the insertion point, unchanged by this branch:

> it is about half of what a BIP340 signature costs here, 15.82
> microseconds against the signer's 8.27, of which the keypair is 7.55

The new table says 16.67, 9.01 and 7.65 for the same three things. A
reader meets the table first and the sentence a screen later, and the
section states `ssa.sign` twice with two values. Either the table takes
the numbers already there, or the section moves to the new set in one
pass — not both. (CHANGELOG.md's entry for
<https://github.com/btclib-org/btclib-secp256k1/issues/153> carries a
third set, 15.7 / 8.2 / 7.3, from a different rounds-and-calls regime;
that one at least sits under a released-work heading and reads as its own
measurement.)

### 2. The premise of the change is not what the section says

> The section already carried every number it needed except that keypair
> rebuild

It carries it: "of which the keypair is 7.55". And 15.82 − 8.27 = 7.55, so
the identity this branch is written to establish is already there to three
digits — arrived at, not stated. What the section genuinely lacks is the
*rule*, stated as a rule, which is the part of this pull request I would
keep whole.

### 3. The table's second row cannot be what was timed

`Signer.__init__` builds the keypair eagerly (`self._keypair =
keypair(prvkey)`), so `ssa.Signer(prvkey).sign(msg)` as written pays the
7.65 the fourth row names, and would measure near the 16.67 of the first
row rather than 9.01. The 9.01 is the per-signature cost with the signer
built outside the timed call — which is what the prose below already means
by "the signer's 8.27". Write the row as `signer.sign(msg)`, or the table
subtracts a cost from itself.

### 4. The overlap section is stale

<https://github.com/btclib-org/btclib-secp256k1/pull/185> landed in
`24269e72` and is in main. So "whichever lands second may want one
cross-reference" is now this branch's own edit to make: the finding that
the chain is *ahead* counted from the 33 octets an xpub carries is, as the
body says, a sharper instance of "what recovers a parse is not always
cheap" than the x-only case, and the two paragraphs now sit in the same
file. The rebase is needed regardless — `DIRTY` on CHANGELOG.md.

### What I checked and agree with

- the rule as one sentence is the right thing to add, and stating it where
  the proposals arrive rather than in
  <https://github.com/btclib-org/btclib-secp256k1/issues/161> is right.
- "the numbers confirm the rule and do not establish it" earns its place,
  and the 55.14 against 54.75 is the reason it does.
- the x-only leg answers the objection the uncompressed-form argument
  invites, which is the objection a reader would otherwise raise.
- `#design` resolves: `## Design` is a heading of this README.

fametrano:
Both points read and one taken.

**Taken: which figures travel.** Fair of a table of four microsecond counts from one laptop. Said in the paragraph that already separated confirming the rule from establishing it — what carries between machines is the *identity* between two of the figures, the difference and the rebuild being the same number, which is a ratio; a ratio survives a change of machine where a count does not. The absolutes are named as one laptop's.

**Declined: the density.** The long multi-clause sentence is this repository's prose, not an accident of this diff — CONTRIBUTING.md's "Documentation and comments" section states the register and the whole README is written in it. Tightening these paragraphs alone would make them the odd ones in their own section. The scannable form the comment asks for is there and is deliberate: the rule is a block quote of one sentence, and the identity is a four-row table. Everything around them is the reasoning, which is what this file is for.

fametrano:
## Second pass, on `67c202c6`

**Verdict unchanged: changes requested.** The new paragraph is an
improvement and does not reach the three findings of the first pass — and
it leans on the row those findings are about.

### What the new commit gets right, and where its reason is weaker than the true one

Naming the four figures as one laptop's is right. But "a ratio survives a
change of machine where a microsecond count does not" is a statistical
argument for something that is structural, which is odd in a paragraph
whose next sentence says the establishing reason is structural.

`ssa.sign` builds the keypair, signs with it and wipes it; `Signer.sign`
is that middle term with the first already paid. So the difference **is**
`secp256k1_keypair_create` by construction, on any machine, and what the
measurement adds is that nothing else of consequence is in that
difference. That is a stronger sentence than a ratio's portability, it is
this section's own kind of argument, and it costs no more words.

Smaller, in the same paragraph: of the four rows, three are measurements
and one is the subtraction of two of them, so "the four absolutes above"
counts a derived figure among the measured ones.

### Still open, and the third is now load-bearing

1. **The section states these three quantities twice.** Below the
   insertion point, untouched: "15.82 microseconds against the signer's
   8.27, of which the keypair is 7.55". The table above says 16.67, 9.01
   and 7.65 for the same three things. Take the numbers already in the
   section, or move the section to the new set in one pass.
2. **The premise is not what the section says**, and it ships:
   CHANGELOG.md's "The section had every number it needed except the
   keypair rebuild". It had it — "of which the keypair is 7.55" — and
   15.82 − 8.27 = 7.55 is the identity, already there to three digits.
   What the section lacks is the rule stated as a rule, which is the part
   worth landing.
3. **The table's second row cannot be what was timed.** `Signer.__init__`
   builds the keypair, so `ssa.Signer(prvkey).sign(msg)` pays the 7.65 of
   the fourth row and would measure near the 16.67 of the first, not 9.01.
   `67c202c6` makes this matter more, not less: the paragraph it adds now
   rests everything that travels between machines on the identity between
   rows three and four, and row two is the row that produces row three. As
   written, the row that carries the claim is the row that names the cost
   it is supposed to have avoided.

### Also

<https://github.com/btclib-org/btclib-secp256k1/pull/185> is in main as
`24269e72`, so the "Overlap with #185" section is stale and its
cross-reference is now this branch's edit to make. `DIRTY` on CHANGELOG.md
still. And `0d87af68` removed the CHANGELOG.md finding from
`.secrets.baseline`, so both baseline hunks on this branch — `308b5ee5`'s
and `67c202c6`'s — are dropped on the rebase rather than resolved.

fametrano:
[`24269e7`](https://github.com/btclib-org/btclib-secp256k1/commit/24269e7),
which rewrote part of the very section this branch adds a rule to. **This
one is not only a rebase**, unlike #195 and #196.

## What moved under it

Two paragraphs were added to *Two outposts past the boundary*, and #161
was closed on them rather than on the prediction it carried:

1. **The composition is not a third saving.** `chain.tweak_add(compressed=False)`
   was claimed to be the fast spelling of the walk and measured slower —
   a chain re-parses nothing in either serialization, so asking it for 65
   octets pays a serialization and the caller's compression line for no
   parse saved. 59.27 against 58.62, where the section's own numbers
   predicted a loss of 0.98.
2. **The three numbers never asked where the walk starts.** The
   uncompressed walk's 54.75 begins from 65 octets an xpub does not
   carry; reaching that form costs one `keys.reserialize`, a compressed
   parse and a serialization. Counted, the chain is ~2.9 *ahead* of the
   walk the section had called it a rounding error behind.

## Why that is a gain for the rule this branch states, not a problem

> a parse has a public bridge and arithmetic has none, and a rule resting
> on which of two measurements is larger inverts on the first machine
> that measures them differently

That is the right shape, and (2) is now the measured demonstration that
the bridge is **not free**: the caller crosses it with `keys.reserialize`,
and the crossing costs exactly the parse the object would have saved. So
the rule can be stated more sharply than "a parse has a bridge" — the
bridge exists, and whether taking it is cheaper than holding the point
depends on which serialization the caller already has. A parsed-key
outpost still fails the test, and now for a reason a reader can check
against the paragraph directly above it rather than against the two
instances.

What I would reread before rebasing: the sentence about the chain saving
"a rounding error" now carries the clause *where the caller already holds
the 65 octets*, so a rule paragraph that repeats the unqualified version
would put the section back into disagreement with itself.

## The mechanical part

`.secrets.baseline` is a line number — take `main`'s file and let the
hook regenerate it rather than merging the hunk. `CHANGELOG.md` has no
union driver (`HISTORY.md` only), so its section is a real conflict; my
entry is in v0.8.0.3's *Documentation*, and this branch adds a heading of
its own, so the collision is context lines.

And the landing rule changed too: fast-forward, with the squash made
locally where the branch carries more than one commit. Push the squash to
the branch before landing it and GitHub still marks this Merged.

fametrano:
All four taken, in `0393089`, and three of them by the same edit: **the table is gone.**

**1 and 2 together.** You are right and my premise was false — I stopped reading the section at the paragraph I was inserting after. `15.82 microseconds against the signer's 8.27, of which the keypair is 7.55` is a screen below, and 15.82 − 8.27 *is* that 7.55, so the identity was already there, arrived at rather than stated, and my table stated it a fourth time in this repository with a fourth set of figures. The prose now points at the section's own numbers and adds none. What is kept whole is the rule, which is the part you say is worth landing.

**3.** Correct, and it was a labelling defect rather than a measurement one: the benchmark built the signer outside the timed lambda, and I wrote the row as the spelling that would not have. It goes with the table.

**Second pass, on the reason.** Taken, and it is the better sentence: `ssa.sign` builds the keypair, signs and wipes; `signer.sign` is that middle term with the first paid; so the difference **is** `secp256k1_keypair_create` by construction, on any machine, and what a measurement adds is only that nothing else of consequence sits in it. The paragraph says that instead of arguing that a ratio ports better than an absolute — which, as you say, was a statistical reason offered two clauses before a structural one. "The four absolutes" is gone with the table it counted.

**4.** `24269e72` is in main, so I made the cross-reference rather than suggesting it. It is now the second leg of "what recovers a parse is not always cheap": the uncompressed-form argument reaches neither an x-only key, which has no second serialization, nor a walk starting from the 33 octets an xpub carries, where reaching that form costs a `reserialize` the chain does not pay. And the last paragraph's "one such pair is a rounding error apart" now points forward at your walk instead of leaving the 55.14/54.75 unattached. The stale "Overlap with #185" section of the description is what this replaces.

**Baseline.** Both hunks dropped rather than resolved; the diff is CHANGELOG.md and README.md.

Gate clean, 558 tests including the README doctests.

fametrano:
## Third pass, on `0393089`: two figures short of an ACK

Dropping the table was the right answer to three findings, and the
by-construction sentence is the argument the section wanted: `ssa.sign`
builds the keypair, signs and wipes; `signer.sign` is the middle term with
the first paid; so the difference is `secp256k1_keypair_create` on any
machine, and the measurement only adds that nothing else of consequence
sits in it. The
<https://github.com/btclib-org/btclib-secp256k1/pull/185> cross-reference
is in, the baseline hunks are gone, and the branch is `BEHIND` rather than
`DIRTY`. Two things left, both one line.

### 1. "a fourth set beside the three this repository already carries"

That count does not check out under either reading. Distinct sets of
figures for those three quantities: **two** — 15.7 / 8.2 / 7.3 in
CHANGELOG.md's entry for
<https://github.com/btclib-org/btclib-secp256k1/issues/153>, and 15.82 /
8.27 / 7.55 everywhere since. Places that state one of them: CHANGELOG.md
twice and README.md twice, beside that first entry. Neither is three.

Which is the argument for not stating it: `git grep '15\.82\|15\.7'` is
the count, whenever it is wanted. "Rather than a set of its own" says
everything the sentence needs.

### 2. "The same subtraction on a parsed point gives 0.269"

0.269 is not what that subtraction gives. It is `keys.parse` of the
uncompressed form, measured directly — the section states it as one of a
pair, "0.269 microseconds against 2.326". The subtraction the section
actually carries is the walk: 64.62 against 55.14 over five steps, which
is 2.37 a step and is the *compressed* parse, 2.326; and against the
uncompressed walk it is 54.75 against 55.14, which is the rounding error
the paragraph below is about.

The by-construction form works here too, and is what the paragraph now
argues everywhere else: what the chain saves is the parse it holds, and
which parse that is depends on the serialization the caller carries —
2.326 compressed, 0.269 uncompressed. That is also the sharper statement
of the distinction, because it says why this saving is the one a caller
can make cheap and the keypair is not.

### Otherwise

The rest reads as ready, and I would take it as it stands with those two
lines changed: the rule as a block quote, the structural reason for what
travels between machines, the x-only leg now as a field square root
rather than a figure, and ergonomics named as what the test does not
weigh. Three commits, so a local squash and a fast-forward push.

fametrano:
Both taken, in `7b4602d`, and both were the same mistake in two places: a number stated where it should have been derived.

**1. The count.** You are right that it checks out under neither reading, and the argument for not stating it is the one that settles it — `git grep '15\.82\|15\.7'` is the count whenever it is wanted, and a count in prose is a thing nothing checks. "Rather than a set of its own."

**2. 0.269.** Also right, and worse than a slip: I wrote "the same subtraction" of a figure that is not a subtraction at all. The by-construction form works here too, and is the sharper statement as you say:

> The chain reads the same way and answers differently, which is the distinction of the paragraph above arrived at from the other end. What it saves is the parse it holds — so *which* parse decides the figure, and that is the caller's to choose: 2.326 microseconds carrying the compressed form, 0.269 carrying the uncompressed one. A keypair offers its caller no such choice.

The last clause is what I had been missing: it is not only that a parse is recoverable, it is that the *caller* chooses its price and cannot choose the keypair's. That is the rule in one line.

Rebased onto `main` at `72489c0`, so #195 and #196 are both in history below this. Gate clean, 569 tests including the README doctests.

Not merging: this is the one of the three that is not an ACK.

fametrano:
## Fourth pass, on `7b4602d`: one sentence short

Both figures are right now. The count is gone and `git grep` is the count,
and the chain gets the by-construction treatment rather than a subtraction
it never made — "which *parse* decides the figure, and that is the
caller's to choose", with "a keypair offers its caller no such choice" as
the sentence that makes the pair of them one distinction. That last clause
is the best line in the section.

What is left is not a number.

### The sentence that inverts the thesis

From `03d46dc8`, in the paragraph on what recovers a parse:

> A caller who wants the parse held holds it, and the parse/serialize
> bridge is public for exactly that; what needs no class beside it is the
> **serialization**, not the holding.

Read it as it parses: the serialization is what needs no class, and the
holding is not. That is the opposite of what this section exists to say —
that a parsed point does not earn an object — and it also undoes the
clause before it, which has the caller holding the parse without one.

The draft it replaced said it plainly: "A caller who wants that parse held
holds it: the parse/serialize bridge is public for exactly this, and needs
no class beside it." What the second pass asked for was that the *escape*
from a parse be named as `keys.parse` rather than as a cheaper
serialization, and that is now said two sentences earlier and said well.
This clause is doing that work a second time, and gets it backwards on the
way.

The form that says it, and it is this section's own first sentence:

> A caller who wants the parse held holds it — `keys.parse` hands back the
> object and the private halves take it, which is what "the private halves
> above hand one object from a call to the next" already promised. What a
> class would add there is a name, not a saving.

### One note, which is yours to weigh

`0.269 against 2.326`, and "the caller's to choose", are now stated in the
new chain paragraph and in the paragraph four above it — "the caller can
make that parse cheap by carrying the uncompressed form instead: 0.269
microseconds against 2.326 […] the chain saves what a caller free to
choose its serialization could have saved itself". Consistent, not
contradictory, so nothing is wrong; but the new paragraph could point at
that pair rather than restate it, and the sentence that earns its place —
the keypair offering no such choice — would then be what it is for.

### Otherwise

Ready. `mergeStateStatus` is `BLOCKED` rather than `DIRTY`, which is the
review and not a conflict; the checks are queued on the new head, and for
a prose-only branch the two that can say anything are the documentation
build and markdownlint. Four commits, so a local squash and a
fast-forward push.

With that one sentence rewritten I would take it as it stands.

fametrano:
Taken, and landed as `7098ddb`.

**The sentence that inverted the thesis.** You are right about how it parses, and the diagnosis is right too: that clause was doing a second time what the second pass had asked for and got it backwards on the way. Your form, as drafted:

> A caller who wants the parse held holds it — `keys.parse` hands back the object and the private halves take it, which is what "the private halves above hand one object from a call to the next" already promised. What a class would add there is a name, not a saving.

The last sentence is the one that makes it, and it is the same shape as the keypair clause: not "there is a bridge", but *what an object would add given the bridge*.

**The note that was mine to weigh: taken.** The chain paragraph now points at the pair rather than restating it — "which parse decides the figure, and that is the pair the paragraph above states, the caller's to choose by the serialization they carry" — which leaves the keypair sentence as the paragraph's only new claim, and that is what it is for. CHANGELOG.md says so the same way.

**On the 13:10 note.** Rechecked before landing: the rule paragraph's "one such pair is a rounding error apart" points at the walk rather than asserting it, so the qualification `24269e72` added — *where the caller already holds the 65 octets* — is where the reader meets it, and the section does not disagree with itself.

Squashed locally from four commits and fast-forwarded, so the sha is the one the gates last ran on. `main` is `7098ddb`.

PR #203, opened 2026-08-16
Give CHANGELOG.md the union driver HISTORY.md has

fametrano:
`.gitattributes` gives HISTORY.md the union driver and states why. This
gives CHANGELOG.md the same one, for the same reason one level down.

## The insertion point repeats per group

HISTORY.md's is the top of the file, and every release adds a section
there. CHANGELOG.md's is the last bullet of a group, and every branch
appends one to the group its change belongs to — so two branches meet
there exactly when their changes are the same kind of change, which is
when they are most likely to be open together. Three pull requests were
open at once earlier today and all three conflicted here, none of them
for a reason a reader has to decide.

## Measured, on two branches appending one bullet each

```
=== without the driver: merge a into b ===
CONFLICT:
- from B
=======
- from A
>>>>>>> a
=== with it ===
clean:
- one
- two
- from B
- from A
```

Which is what a merge of two appends should be, on rebases included —
and the rebase is how a branch under review is corrected here, so it is
the case that matters.

## The price, which is worse here than there

The file stops conflicting at all, so two branches editing *the same*
bullet two ways merge in silence. That is HISTORY.md's trade, and it
bites harder in this file because nothing reads it: HISTORY.md at least
states a version the release path checks, while a bullet is prose no
test has an opinion about. What buys it off is what the accident costs —
a sentence about what changed, not a behaviour — against a conflict that
costs every branch and has nothing in it to decide.

`btclib` itself carries the driver on both files, which is the same
conclusion reached there.

## Scope

One file. No CHANGELOG.md entry, on HISTORY.md's own precedent: a merge
driver is not something a user of the package meets.

## Gate

`uvx pre-commit run --all-files` clean, 569 tests, coverage 100.00%.

## Summary by Sourcery

Build:
- Update .gitattributes to apply the union merge driver to CHANGELOG.md in addition to HISTORY.md.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,3 +9,20 @@
 # conflict with nothing to decide costs every branch, and a silent merge
 # costs only the branches that edited one line two ways.
 HISTORY.md merge=union
+
+# CHANGELOG.md is the same list one level down. A branch appends its
+# bullet to the group the change belongs to, so two branches meet at that
+# group's last line rather than at the top of the file -- and they meet
+# there whenever both changes are the same kind of change, which is when
+# they are most likely to be open together. Without the driver that is a
+# conflict; with it both bullets survive, in the order the merge finds
+# them.
+#
+# The price is the one above and it bites harder here, because nothing
+# reads this file: HISTORY.md at least states a version the release path
+# checks, while two branches rewriting one bullet two ways merge in
+# silence with nothing to notice it. What buys that off is what the file
+# holds -- prose about what changed, where the accident is a sentence and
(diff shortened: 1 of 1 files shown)


fametrano:
Landed on main as [`7c134bc`](https://github.com/btclib-org/btclib-secp256k1/commit/7c134bc2), a fast-forward carrying its own signature rather than a button's. GitHub reports this pull request as closed and not as merged only because the head was moved to the rebased sha after that sha was already `main`: the commit is `main`, and `git log --format=%G? -1 7c134bc` is `G`.

PR #205, opened 2026-08-16
Stop paying for a guard, and say what a refused object now answers

fametrano:
`context.guarded` cleared the thread, made the libsecp256k1 call, and raised
whatever the illegal callback had recorded. Around one trivial C call it cost
**0.201 µs of the 0.292 that call took** — an Apple M5, macOS 26.6, arm64,
CPython 3.13.14, minimum of 9 rounds of a million calls — at each of the
twenty-six places it was used, and an entry point crossing twice paid it twice.

It is gone. `context.check` is unchanged and still exported; what it reports is
now the caller's to ask for, and its docstring says when.

## This changes a contract, not only a cost

Where the wrapper read a return code it still raises, with its own message
rather than libsecp256k1's. Nine move from `ValueError` to `RuntimeError`,
libsecp256k1 answering the same 0 for a refusal as for a failure of its own:
`keys.serialize`, `keys._pubkey_sort_`, `xonly.serialize`, `xonly._drop_y`,
`xonly.from_keypair`, `recovery._to_der_`, `recovery.serialize_compact`,
`dsa.serialize_der`, `dsa.serialize_compact`, `silentpayments.serialize_label`.

Where the wrapper believed the return value there is now no exception at all:

| call | answers | for an object libsecp256k1 refused |
| --- | --- | --- |
| `dsa._verify_`, `ssa._verify_` | `False` | the same False a bad signature gets |
| `keys._pubkey_sum_` | `None` | the same None the point at infinity gets |
| `keys._pubkey_cmp_` | an ordering | the same 0 two equal keys get |
| `dsa._is_low_s_` | `True` | the same True a normalized signature gets |
| `ecdh._shared_secret_` | **32 bytes** | a shared secret with nobody |

Each is driven in `tests/test_callbacks.py` rather than described. The ECDH one
because it is the gravest: the call succeeds, the answer is the right length,
and only the thread says it is worthless.

**No entry point taking octets is in that position.** `verify`,
`pubkey_tweak_add`, `shared_secret` and the rest parse what they are given, so
the objects they hand libsecp256k1 are ones it has just built; a bad key is the
parse's `ValueError` and the thread is left clean. What is given up belongs to a
caller passing a `secp256k1_pubkey` of its own to a `_foo_` half, and
`keys.pubkey_verify` on the octets it came from settles it once instead of at
every call. README.md and SECURITY.md say so where they promised otherwise.

## Two allocations go with it

- **`_scalar.octets` answers `bytes` without asking what `bytes` already
  answers**, 0.034 µs against 0.080. The copy and the width tests stay on the
  path a `bytearray` or a `memoryview` takes. `dsa.verify` passes three
  arguments through it.
- **`keys.serialize` declares its buffer with a literal cdecl** and reads the
  size back off it once, 0.313 µs against 0.356 with the guard already gone and
  0.604 before it. `ffi.new` of an interpolated `f"char[{size}]"` parses that
  string on every call, and `ffi.sizeof` was called twice where the size was in
  hand.

Its length buffer is still built per call and deliberately **not** hoisted,
though it holds the same number every time: libsecp256k1 writes 0 into it before
it does anything and restores it only on success, and holds what it finds there
against 33 or 65 on the way in. One failed call would leave a shared buffer at
zero, and every later serialization — on any thread, of a perfectly good key —
would be refused. That was measured rather than reasoned about, and is why the
0.085 µs it would have bought is left on the table.

## What it comes to

Microseconds per call, machine above, minimum of 9 rounds, and the quickest of
two passes run in either order so that neither tree is the one measured on a
warm machine:

| call | before | after |
| --- | --- | --- |
| `keys.parse`, 65 bytes | 0.258 | 0.213 |
| `keys.parse`, 33 bytes | 2.314 | 2.294 |
| `keys.serialize` | 0.604 | 0.313 |
| `xonly.parse`, 65 bytes | 0.832 | 0.500 |
| `keys.pubkey_tweak_add` | 4.179 | 3.480 |
| `dsa.sign`, DER | 12.063 | 11.766 |
| `dsa.sign`, compact | 12.072 | 11.733 |
| `dsa.verify`, DER | 12.221 | 11.846 |
| `dsa.verify`, compact | 12.203 | 11.880 |
| `ssa.sign` | 15.724 | 15.631 |
| `ssa.verify`, 65-byte key | 13.159 | 12.476 |
| `ssa.verify`, x-only key | 14.624 | 14.272 |

The 33-byte parse is the row where there was nothing to win, and the figures say
so: the field square root is what that row is.

Against the other wrappers of the same library, in one process on one
interpreter, this moves three rows of btclib-benchmarks' table 01 to first:
public key parse at 65 bytes (0.221 against coincurve's 0.269), at 33 bytes
(2.350 against 2.418), and ECDSA verify in DER (12.012 against 12.120).

## And the frames between the entry point and libsecp256k1

A second commit spells the hot entry points out rather than composing them from
their halves. Measured in one process, shipped against a hand-inlined spelling
making the same libsecp256k1 calls with the same argument checks:

| entry point | the frames cost | of |
| --- | --- | --- |
| `keys.parse`, 65 bytes | 0.027 | 0.218 |
| `keys.parse`, 33 bytes | 0.036 | 2.338 |
| `keys.pubkey_tweak_add` | 0.083 | 3.420 |
| `ssa.verify` | 0.086 | 14.308 |
| `dsa.verify` | 0.154 | 11.845 |
| `dsa.sign` | 0.261 | 11.924 |

**Every private half stays, and removing them was measured before it was
declined.** They are what a caller holding a parsed object does not pay a parse
for: 0.558 µs per call against a 65-byte key, 2.676 against a 33-byte one —
where the parse is a field square root — and 2.687 on a tweak applied to its own
output. An order of magnitude more than the frames are worth, and in the
opposite direction. So the body is written twice, and
`tests/test_parsed_keys.py` is what holds the two spellings to one answer, pair
by pair, with `test_every_private_half_is_paired` making an unpaired half an
absence rather than a test nobody wrote. That test is why the frames were taken
here and not in `recovery`, `ellswift` or `silentpayments`, which still compose.

`dsa._sign_`'s refusal of a key outside [1, n-1] is now driven directly — the
coverage ratchet found it the moment the two spellings parted.

## End to end, both commits

`main` against the branch, the two alternated over three rounds so that thermal
drift lands on both, minimum per tree:

| call | main | branch | delta |
| --- | --- | --- | --- |
| `keys.parse`, 65 bytes | 0.264 | 0.215 | −0.049 |
| `keys.parse`, 33 bytes | 2.395 | 2.372 | −0.023 |
| `keys.serialize` | 0.601 | 0.329 | −0.272 |
| `xonly.parse`, 65 bytes | 0.840 | 0.496 | −0.343 |
| `keys.pubkey_tweak_add` | 4.206 | 3.488 | −0.718 |
| `dsa.sign`, compact | 12.324 | 11.950 | −0.374 |
| `dsa.verify`, DER | 12.449 | 11.907 | −0.542 |
| `dsa.verify`, compact | 12.471 | 11.970 | −0.501 |
| `ssa.sign` | 16.067 | 15.851 | −0.216 |
| `ssa.verify`, 65-byte key | 13.318 | 12.561 | −0.757 |
| `ssa.verify`, x-only key | 14.878 | 14.466 | −0.413 |

Against the other wrappers in one process, three rows of btclib-benchmarks'
table 01 are now first: parse at 65 bytes 0.206 against coincurve's 0.272, at 33
bytes 2.331 against 2.414, and ECDSA verify in DER 11.918 against 12.133.

## Gate

`uv run --locked --no-default-groups --group test pytest --cov` — 571 passed,
coverage 100.00%. `uvx pre-commit run --all-files` — exit 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,52 @@ release-notes length in the first place, and are still in
 
 ## v0.8.0.3 (work in progress, not released yet)
 
+### The frames between an entry point and libsecp256k1
+
+- **Five entry points are spelled out rather than composed of their
+  halves**: `keys.parse` and `xonly.parse`, which delegated to their
+  `_parsed`, `keys.pubkey_tweak_add`, which was three calls, and
+  `dsa.verify` and `ssa.verify`, which parsed through one call and
+  verified through another. What a python frame is worth was measured by
+  alternating the two spellings in one process, 7 rounds each and the
+  minimum kept for each -- an Apple M5, macOS 26.6, arm64, CPython
+  3.13.14: `keys.parse` of 65 bytes 0.010 of 0.205 and of 33 bytes 0.013
+  of 2.324, `xonly.parse` of 65 bytes 0.012 of 0.485, `ssa.verify` 0.009
+  of 14.275, `dsa.verify` 0.035 of 11.813, `keys.pubkey_tweak_add` 0.037
+  of 3.403. Small everywhere, and largest where the call is shortest.
+- **`dsa.sign` is composed still, and that is a measured result rather
+  than an omission.** Three spellings were tried -- `_sign_` inlined, the
+  DER serialization inlined, and both -- and all three land within 0.03
+  microseconds of the composed one and on the wrong side of it. A
+  signature is 11.9 microseconds of libsecp256k1; two python frames do
+  not show against it, and a body written twice for nothing is worse
+  than the frames.
+- **Measuring the two spellings one after the other is what made them
+  look bigger.** Run in sequence rather than alternated, the same
+  comparison reported 0.027, 0.083, 0.154 and 0.261 for four of the
+  calls above -- two to four times the alternated figure, and for
+  `dsa.sign` a saving that is not there at all. The machine drifts under
+  a benchmark, and a pair measured in a fixed order gives the drift to
+  one side of it.
+- **Every private half stays, and removing them was measured before it
+  was declined.** They are what a caller holding a parsed object does
+  not pay a parse for, and that is worth 0.558 microseconds per call
+  against a 65-byte key, 2.676 against a 33-byte one -- where the parse
+  is a field square root -- and 2.687 on a tweak applied to its own
+  output. Two orders of magnitude more than the frames are worth, and in
+  the opposite direction: deleting them to save a frame on the octets
+  path would cost the object path a parse at every call. `ssa.Signer`
+  and `keys.PubkeyTweakChain` are the same trade already made.
+- **So five bodies are written twice**, once taking octets and once
+  taking the object, and the equality is asserted rather than assumed:
+  `tests/test_parsed_keys.py` holds every public half to its private
+  one, pair by pair and over both serializations of a public key, and
+  `test_every_private_half_is_paired` makes an unpaired half an absence
+  rather than a test nobody wrote. That test is what makes the
+  duplication survivable, and it is why it was not extended to
+  `recovery`, `ellswift` or `silentpayments`, whose entry points no
+  benchmark reaches and which still compose.
+
 ### What a crossing costs, and what stops being paid for it
 
 - **`context.guarded` is gone, and every call it held is now a bare
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,9 +28,9 @@ release-notes length in the first place, and are still in
   call.** It cleared the thread, ran the libsecp256k1 call, and raised
   whatever the illegal callback had recorded, which turned a refused
   object into an exception at each of the twenty-six places it was
-  used. Measured around one trivial C call, it cost 0.175 microseconds
-  of the 0.266 that call took — an Apple M5, macOS 26.6, arm64, CPython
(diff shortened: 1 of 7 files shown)


fametrano:
Landed on main, in two pushes rather than one, and GitHub could not reconcile
this pull request because neither push named the sha it points at.

The head here was `681ace1` when the fast-forward was made, and it landed as
such. The branch was then rewritten — `fd77af0` replacing that commit and
`a9e7c2d` following it — so main held the earlier of two heads: same code,
every measurement superseded. #206 carried the difference over, rebased onto
the low-r work that had arrived meanwhile, and is on main as
[`06bb46c`](https://github.com/btclib-org/btclib-secp256k1/commit/06bb46c1156dd474139048be2e99574efd123e6a).

What main now holds, checked rather than assumed: no `guarded` left in the
package, the remeasured figures in CHANGELOG.md (0.201 of 0.292),
`_scalar.py` (0.034 against 0.080) and `keys.py` (0.313), and this branch's
README wording.

Closing by hand: `a9e7c2d` will never be reachable from main, its base
`7c134bc` having been passed twice.

Issue #204, opened 2026-08-16
ECDSA low-R grinding: whether it belongs here, and which side of the boundary the loop runs on


fametrano:
## What is asked

Whether these bindings should offer low-R grinding in the ECDSA signing
wrappers, and — if they should — on which side of the cffi boundary the
loop runs: many crossings, one per attempt, with the retry decided in
python; or one crossing into something that grinds and comes back with
the signature it settled on.

Nothing is implemented yet. This issue is the measurement and the
decision it is for.

## The `s` is not what gets ground

`secp256k1_ecdsa_sign` already returns the lower-s form and always has,
which is why
[`_sign_`](https://github.com/btclib-org/btclib-secp256k1/blob/7c134bc26d01f5d94b9e2db520e627f51032e57b/btclib_secp256k1/dsa.py#L139-L186)
documents its result as "in the lower-s form libsecp256k1 always
produces" and `is_low_s` on a fresh signature is a doctest that reads
`True`. There is no `s` to grind for: one call already gives the only
`s` this package will emit, and `normalize` exists for the signatures a
caller did *not* make.

What Bitcoin Core grinds is `R`, in `CKey::Sign`, and it is a size
policy rather than a cryptographic one: with the top bit of `r` clear,
DER spends no leading zero on it and the encoding is one octet shorter.
The scheme is fixed by Core and is worth copying verbatim if anything is
copied at all — first attempt with no extra entropy, then a `uint32`
counter written little-endian into the first 4 of 32 zero octets, retried
until `secp256k1_ecdsa_signature_serialize_compact` yields a first byte
below `0x80`.

Measured here over 2000 keys, one message, the counter scheme above:

| | DER 69 | DER 70 | DER 71 |
| --- | --- | --- | --- |
| as `sign` emits today | 7 | 973 | 1020 |
| ground for low R | 18 | 1982 | 0 |

One octet, half the time, for rather more than half again the signing
cost: the attempts came out at 2.09 on average, with a tail — 14 attempts
for one key of the 2000. That is the whole of the benefit, and in Bitcoin
it is legible one level up, where the 71st octet is in a `scriptSig` and
the sighash byte follows it.

## The question this issue is really about

Grinding is the first thing anyone would ask this package for that is
*not* one libsecp256k1 call. The README's Design section allows exactly
one reason for that — "what the composition saves is the crossing between
its halves, which is the caller's cost and not its own choice" — so the
crossing is what has to be measured before anything is written.

[...]

PR #209, opened 2026-08-16
Let a caller ask what the grinding loop asks

fametrano:
`dsa.sign(grind=True)` landed in #207 knowing what a low `r` is and
giving a caller no way to ask it of a signature somebody else made,
where `is_low_s` has had that shape since the beginning. This is the
predicate, and the grinding loop now asks it rather than spelling the
same test a second time.

## What it is, and what it is not

- `dsa.is_low_r(signature_bytes)`, with `dsa._is_low_r_(signature)` the
  half taking the parsed one.
- **Not a rule.** `verify` asks nothing about `r`: a high-r signature is
  valid and always was. What this answers is whether a signature is the
  octet shorter in DER that `sign(grind=True)` makes it — for a caller
  enforcing that on its own signatures, or counting how many of somebody
  else's carry it.
- **A serialization and a comparison**, no libsecp256k1 call answering
  it: `r` is the first 32 octets of the compact form, so its leading
  octet is the one DER would put a zero in front of. Reading it out of
  the opaque signature object is open to nobody — those 64 octets are
  internal scalar limbs.
- **An unreadable object gets a `RuntimeError`** from `serialize_compact`
  rather than the `True` `_is_low_s_` answers: a serialization that did
  not happen leaves no octet to read. The docstring names the difference.
- Asking it per attempt costs **0.19 µs of the 24.91** a ground signature
  takes, against the compact buffer the loop used to keep.

## The figures are re-measured

681ace1 took the guard off the wrappers after #207 was measured, which
moved every number in that PR's prose and closed two of them together.
Minimum of 7 rounds over 2000 keys, compact form:

| | per signature |
| --- | --- |
| a loop over the wrappers, per attempt | 24.88 µs |
| as implemented, asking `_is_low_r_` | 24.91 µs |
| the same with scratch buffers at module level, which thread safety rules out | 24.14 µs |

The first two are one number now, so what a C loop could have saved is
the 0.8 of twenty-five that separates them from the third — #204's
conclusion unchanged, and cheaper to state. `dsa.sign` is 11.78 and
`dsa.sign(grind=True)` 25.24. README.md, CHANGELOG.md and HISTORY.md
carry the new figures.

## What judges the predicate

The same vectors, read for it rather than for the loop: Core's
`key_test1` signatures are low-r because Core says so, its
`key_signature_tests` messages are read the way Core reads them —
`is_low_r` against the length-of-r octet, over all twenty high-r tries as
well — and rust-secp256k1's five-attempt vector is the `False` before it
and the `True` after.

Suite green (581 tests), coverage gate at 100.00%, every hook passes.

## Summary by Sourcery

Add a public predicate to query whether a DER-encoded signature has low r, and make the grinding loop use that shared predicate while updating performance figures and tests to reflect and validate the new behavior.

New Features:
- Expose dsa.is_low_r (and its private counterpart dsa._is_low_r_) to let callers check whether a signature is low-r and thus one octet shorter in DER.

Enhancements:
- Refactor the grinding loop to reuse the shared low-r predicate instead of maintaining its own compact-buffer test.
- Clarify documentation and history around grinding, low-r semantics, and performance characteristics with updated benchmark figures.

Documentation:
- Update README, CHANGELOG, and HISTORY to describe is_low_r, its non-rule nature, and refreshed performance measurements of grinding.

Tests:
- Extend test vectors and core tests to validate is_low_r against Bitcoin Core and rust-secp256k1 low-r properties and to include the new private helper in parsed-key and bytes-like scans.


fametrano:
## Review of f6326b4

Read, built and run on this tree rather than from the description: the
suite is green (581 passed) and coverage is 100.00%, the branch rebases
onto 26da430 without a conflict, and it stays green and at 100.00% on
the result. Core's `SigHasLowR` is the function this claims it is —
`src/key.cpp` serializes compact and returns `compact_sig[0] < 0x80`,
which is `_is_low_r_` octet for octet.

The predicate, where it is written, and the vectors it is held to are
right. Three things are not, and one of them is a number published in
three files, so **not an ACK as it stands**.

### Required

**1. The contract the new docstring introduces is described, not driven.**

`_is_low_r_` states that an unreadable object gets a `RuntimeError`
where `_is_low_s_` answers `True`. That is a shape, and
`tests/test_callbacks.py` is the file whose own docstring says each of
its shapes is driven rather than described — with
`test_a_signature_libsecp256k1_cannot_read_is_low_s_like_any_other`, the
other half of the very contrast, sitting in it. Nothing asks the new
one, so the difference the docstring names is prose alone.

It holds. This is the test, run against this head, and it passes:

```python
def test_a_signature_libsecp256k1_cannot_read_has_no_low_r_to_read() -> None:
    """`dsa._is_low_r_` raises where `_is_low_s_` answers True.

    The two are asked the same way and answer differently, which is what
    the docstring names: `_is_low_s_` reads a return code, and a refused
    object is reported as unchanged exactly as an already-normalized one
    is, where this has to serialize first -- and a serialization that did
    not happen leaves no octet to compare.
    """
    with pytest.raises(RuntimeError, match="signature serialization failed"):
        dsa._is_low_r_(ffi.NULL)
    with pytest.raises(ValueError, match="illegal argument: sig != NULL"):
        context.check()
```

Sourcery's comment asks for the same `RuntimeError` in the second half
of its suggestion; the `ValueError` half of it is `parse_der`'s refusal,
which is covered where that is covered and needs nothing here.

**2. `0.19 microseconds` has no row in the table it is stated beside.**

The table publishes three figures: 24.88 for the wrappers per attempt,
24.91 for what is implemented, 24.14 for the module-level buffers. The
variant the 0.19 is measured against — the loop as it was before this
branch, keeping one compact buffer per call, on this tree — is a fourth,
and it is not published. The claim is in the `_grind` docstring, the
commit message, CHANGELOG.md and the pull request body; a reader holding
the table can derive +0.03 from it and nothing else. Add the row, and
the cost this branch pays becomes checkable where it is claimed.

**3. The third row does not reproduce here, and it is the row the
conclusion rests on.**

Same method: this head, editable build, CPython 3.14.6, arm64, 2000 keys
signed once each, the loops alternated in one process, 7 rounds, minimum
kept for each — plus a second entry running the same code as the first,
as the noise detector. Microseconds per signature, three runs:

| | run 1 | run 2 | run 3 |
| --- | --- | --- | --- |
| implemented − kept compact buffer | +0.32 | +0.27 | +0.28 |
| implemented − buffers at module level | +0.30 | +0.36 | +0.43 |
| implemented − the same code again | −0.19 | −0.03 | +0.06 |

So the kept buffer and the module-level one are about 0.07 apart, where
the published pair implies 0.58 between them (24.91 less the 0.19, less
24.14). What that variant stops doing per call is two allocations, and
they bound it: `ffi.new("unsigned char[64]")` is 0.111 microseconds and
`ffi.new("unsigned char[32]")` 0.082, 0.219 together over 200 000
iterations. Nothing in that variant can be worth 0.58.

The prose consequence is the part that matters. The gap between what is
implemented and the module-level loop measures about 0.35 here rather
than 0.77, and nearly all of it is the per-attempt serialization this
branch introduces rather than the buffer placement. "What a loop
crossing the boundary once could save is the 0.8 of twenty-five that
separates them from the third" therefore says something it did not say
before: most of that 0.8 is a cost this branch has just added, and it is
in README.md, CHANGELOG.md and the commit message as a restatement of
https://github.com/btclib-org/btclib-secp256k1/pull/204's conclusion.

My absolute figures are about 23.6 against your 24.9, so the machine is
not in the state yours was in and the numbers are not a correction of
yours. The differences are each within one alternated run, and the
allocation ceiling is not a matter of machine state.

### Useful

**4. The new pairing entry cannot tell the two halves apart.**

`tests/test_parsed_keys.py` pairs `_is_low_r_` over `DER`, whose r is
`0x29…`: both halves answer `True`, so an entry that answered `True`
always would pass it. The `_is_low_s_` entry beside it takes `HIGH_S`
deliberately, which is the `False`. `dsa.sign(MSG, 1)` is high-r on this
file's own message — a `HIGH_R` constant makes the equality discriminate
the way its neighbour does. The suite covers the `False` through
tests/test_vectors.py, so this is about what that file is for rather
than about coverage.

**5. `_grind` lost its `Raises:` section and still raises.**

The `RuntimeError` moved from the inline serialization into
`_is_low_r_`; it is still what `_grind` can propagate, and `_sign_`, its
only caller, documents `ValueError` alone. Either keep the section or
put it on `_sign_`.

**6. "the same loop with its scratch buffers held at module level"** names
a variant of the loop as it was, not of the loop being described: after
this branch the implemented loop holds one buffer. The row asked for in
2 is also what disambiguates which loop each row is.

### Verdict

Hold. With 1 and 2 done, and 3 either re-measured or the sentence
rewritten to what it does measure, this lands: the design — one spelling
of the predicate, the public half taking DER as `is_low_s` does, the
serialization rather than the DER length, the vectors reading Core's
`der[3]` for the same answer — is the right one, and 4 to 6 are cheap
enough to take with them.

fametrano:
## Addressed in 75cbb44, rebased on 26da430

All six, and **3 was right in a way that makes the earlier table wrong
rather than off**: the third row of #207 was not the loop it said it was.

### 1. The contract is driven now

Your test, verbatim, beside the `_is_low_s_` half of the same contrast in
`tests/test_callbacks.py`. It passes here too, and the file it is in is
the one whose docstring says these shapes are driven. Sourcery's
`ValueError` half is `parse_der`'s and needs nothing, as you say.

### 2 and 3. The row was missing because the loop was not the loop

The 24.14 of #207 was not "the same loop with its buffers at module
level". It was a loop of raw `lib` calls that also skipped `scalar`,
`octets` and the `_sign_`/`_signed` frames — so its distance from the
implemented loop was mostly validation and python frames, not buffer
placement, and 0.77 was never buffer placement's to claim. Your
allocation ceiling is what makes that unarguable: two `ffi.new`s cannot
be worth 0.58.

Four loops now, differing only in what they are said to differ in,
alternated in one process, 2000 keys signed once each, seven rounds with
the minimum of each kept, compact form, and the noise row you asked for:

| the loop | µs per signature |
| --- | --- |
| a caller's own, over the public wrappers, once per attempt | 24.7 |
| as implemented, asking `is_low_r` per attempt | 24.8 |
| the same, keeping one compact buffer per call | 24.6 |
| the same again, with both scratch buffers at module level | 24.5 |
| the implemented loop run twice — the noise | ±0.1 |

Three runs, deltas against the implemented row: predicate vs kept buffer
+0.13 / +0.31 / +0.27, kept buffer vs module level 0.09 / 0.02 / 0.12,
noise −0.19 / −0.03 / +0.06 in the earlier harness and ≤0.1 in this one.
Your two numbers reproduce; mine did not.

So the fourth row exists where the 0.2 is claimed, and the prose says
what it now measures: three tenths across four loops against a tenth of
noise, a caller's own loop not measurably behind the package's, and
therefore *whatever a loop crossing the boundary once could save is
inside that spread* — most of which is the per-attempt serialization this
branch adds, not where buffers live. "The 0.8 of twenty-five" is gone
from README.md, CHANGELOG.md and the commit message; what replaces it
says #204's conclusion with less: no C, and grinding here earns its place
by the scheme and the vectors, not by microseconds. `dsa.sign` 11.7,
`dsa.sign(grind=True)` 24.9, one decimal because the second is noise.

### 4. `HIGH_R`

`dsa.sign(MSG, 1)` is high-r on that file's message, as you said; the
pairing takes it, so both halves have a `False` to disagree on the way
`HIGH_S` gives `_is_low_s_` one.

### 5. `Raises:`

Restored on `_grind`, naming `_is_low_r_` as where it comes from, and
added to `_sign_`, which is what a caller of the private half meets.

### 6. Which loop each row is

The table names all four, and the sentence about module-level buffers is
now about the last row rather than about the implemented one — which
holds one buffer, as you note.

Suite 582 passed, coverage 100.00%, every hook green.

fametrano:
## ACK 75cbb44

All six are addressed, and the two numbers I asked about were re-derived
rather than restated. Built and run here on a worktree at this head:
582 passed, coverage 100.00%, the new `Example` is executed by
`tests/test_examples.py` with the rest, and the commit is signed (`G`)
and sits on 26da430, so the landing is the fast-forward push. Every
check is green but two wheel builds still running.

Formally it is a comment and not an approval: GitHub does not let the
author's own account approve.

### The table reproduces

Four loops written from your description — the caller's over
`dsa.sign(compact=True)` once per attempt, the implemented one, the same
with a compact buffer kept per call, the same again with both buffers at
module level — the four answering the same octets (asserted before
timing), alternated in one process, 2000 keys, seven rounds, minimum of
each kept, plus the implemented loop a second time as the noise row.
Microseconds, three runs:

| | run 1 | run 2 | run 3 |
| --- | --- | --- | --- |
| implemented − wrappers per attempt | +1.13 | +0.12 | −0.03 |
| implemented − kept compact buffer | +0.75 | +0.14 | +0.55 |
| kept buffer − buffers at module level | −0.11 | +0.03 | −0.30 |
| implemented − the same loop again (noise) | +0.23 | +0.12 | −0.07 |

Every claim the prose now makes survives this. The caller's own loop is
not measurably behind the package's — I cannot tell them apart either,
in either direction. The predicate costs something real, positive in all
three runs here and in all three of yours, at 0.2 published against 0.4
on the mean of mine. Where the buffers live is not measurable at all: my
three straddle zero, which is your 0.1 read on a noisier machine.

The one thing to keep in mind is that my noise row reaches 0.23 within a
run, and the absolute figures drift 1.3 between runs (24.4 to 25.7 for
the same loop), against your 0.1. So the tenths in the table are at the
edge of what the method resolves — which is what the prose now says, and
the reason the conclusion it draws is the right one: the saving lies
inside the spread, so grinding earns its place by the scheme and the
vectors rather than by microseconds. A noisier machine makes that
stronger, not weaker.

`is_low_r` asked once of a signature already made measures 0.34, 0.40 and
0.44 here — the 0.5 in CHANGELOG.md, and nothing rests on the tenth.

### The six

1. The contract is driven, beside its sibling, in the file that says it
   drives them. Passes here.
2. and 3. The fourth row exists, the four loops differ only in what they
   are said to differ in, and "the 0.8 of twenty-five" is gone from
   README.md, CHANGELOG.md and the commit message. Nothing in the tree
   still carries the retired figures: `12.62`, `25.23`, `25.90`, `24.82`,
   `23.82` and `0.19` return nothing outside `uv.lock` hashes.
4. `HIGH_R = dsa.sign(MSG, 1)` is high-r on that file's message, so the
   pair now has a `False` to disagree on, as `HIGH_S` gives the other.
5. `Raises:` on `_grind` names `_is_low_r_` as where it comes from, and
   `_sign_` carries it for the caller of the private half.
6. Each row names its loop, and the module-level sentence is about the
   last row.

Good to land.

PR #213, opened 2026-08-16
Read the version when a caller asks, not when the module loads

### LINKED ISSUE #210: Importing the package spends thirteen milliseconds on a version string nobody read
## What it costs

`import btclib_secp256k1` is 14.83 ms, and 13.4 of those are two imports
the package does not need to have done by then. The compiled extension —
the thing this package exists to load — is 0.57 of them.

Apple M5 arm64, macOS 26.6, CPython 3.14.6, a local static build of
0.8.0.3, minimum of 12 fresh interpreters per row:

| | import |
| --- | --- |
| as it is today | 14.83 ms |
| with `__version__` read on first access | 3.77 ms |
| and with `pathlib` imported where it is used | 1.43 ms |

## Where it goes

[`__init__.py`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/btclib_secp256k1/__init__.py#L46-L54)
reads the version out of the installed distribution metadata at import
time. That is the right place for the value to come from — `pyproject.toml`
stays the only thing a release bumps — but `importlib.metadata` is not a
small module to reach for: it pulls in `email`, `json`, `inspect`, `ast`,
`dis`, `tokenize`, `zipfile`, `bz2` and `compression.zstd` behind it.

`pathlib` is imported beside it and used in one place: the branch of
[`_load_lib`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/btclib_secp256k1/__init__.py#L112-L131)
that looks for a shared object to `dlopen`. A static wheel — the default
one, and the one every platform in the matrix ships — returns from
`hasattr(module, "lib")` two lines earlier and never reaches it.

Fresh interpreter, minimum of 5, each module on its own:

| | import |
| --- | --- |
| `importlib.metadata` | 9.69 ms |
| `pathlib` | 2.26 ms |
| `secrets` | 1.10 ms |
| `threading` | 0.30 ms |
| `_btclib_secp256k1`, the extension | 0.57 ms |

## What would change

A module-level `__getattr__` (PEP 562) leaves the value where it is
declared and moves the reading of it to the first access. The three
spellings this repository actually uses keep working: the attribute, the
`from … import __version__` of
[`tests/test_extension.py`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/tests/test_extension.py#L29)
and `published.yml`, and the `python -c` of the issue templates. A
caller that *does* read it pays what it pays today, 14.83 against 14.79 —
the same number.

`import pathlib` moves inside the dynamic branch, with the comment saying
why it is there and not at the top.

## What it does not change

Nothing about what the package answers, and nothing about the version
being declared once. `dir(btclib_secp256k1)` stops listing `__version__`
until something reads it, which is the one visible difference and is what
PEP 562 costs everywhere it is used.

## Where the saving lands

btclib depends on this package, so the 13.4 ms are on btclib's import
path, and from there on every process that imports it.

fametrano:
Importing this package was 14.83 ms and the compiled extension it exists
to load was 0.57 of them. The rest was two module-level imports, and
neither is needed by the time the import finishes.

## What moved

- **`__version__` is read on first access**, through a module-level
  `__getattr__` (PEP 562). Where the value comes from is unchanged — the
  installed distribution metadata, so `pyproject.toml` stays the only
  place a release bumps. `importlib.metadata` pulls `email`, `json` and
  `inspect` behind it, among others.
- **`pathlib` is imported where it is used**: the branch of `_load_lib`
  that globs for a shared object to `dlopen`. A static extension returns
  from `hasattr(module, "lib")` two lines above and never reaches it,
  and a static extension is what every platform of the matrix ships by
  default.

Apple M5 arm64, macOS 26.6, CPython 3.14.6, minimum of 12 fresh
interpreters:

| | import |
| --- | --- |
| before | 14.83 ms |
| with `__version__` read on first access | 3.77 ms |
| and with `pathlib` where it is used | **1.55 ms** |

A caller that reads the version pays what it paid: 15.26 ms against
14.83, the same work moved rather than removed.

## The `if TYPE_CHECKING`, which is not decoration

A module that has a `__getattr__` is a module mypy stops checking
attribute names on. Measured on this tree, with the plain spelling:

```
$ mypy --strict probe.py     # from btclib_secp256k1 import nosuchmodule
                             # btclib_secp256k1.nonesuch_typo
Found 1 error   # only the one inside a submodule
```

against three on `main`. `from btclib_secp256k1 import nosuchmodule`
passing `--strict` is not a trade worth 13 ms — that is the line every
caller of this package writes. Declaring `__version__: str` to the type
checker and defining the function only at run time keeps all three
errors, and the new test needing a `type: ignore[attr-defined]` to ask
for a missing attribute is the proof that it does.

The two deferred imports carry `# noqa: PLC0415` with the reason beside
them, deferring being the whole of what the change is.

## What a caller has to know

`dir(btclib_secp256k1)` does not list `__version__` until something
reads it, which is what PEP 562 costs wherever it is used. Nothing else
changes: the attribute, `from btclib_secp256k1 import __version__` and
reading the metadata directly all answer what they answered — the three
spellings this repository itself uses, in `tests/test_extension.py`,
`published.yml`, `test.yml` and the issue templates.

## What judges it

`test_no_such_attribute` is new and is the half no other test reached:
what the fallback answers for a name that is not `__version__`.
`test_version` keeps its two assertions, and its docstring stops
claiming that a missing distribution would fail every test in the suite
— it now fails this module's collection, which is where the `from`
import is.

Suite green (583 tests), coverage gate at 100.00%, every hook passes.
CHANGELOG.md and HISTORY.md carry the figures.

Closes #210

## Summary by Sourcery

Defer expensive imports so that the package’s import time is significantly reduced while preserving type-checking and version semantics.

Enhancements:
- Lazy-load the __version__ attribute via a module-level __getattr__ so distribution metadata is read only on demand.
- Limit the impact of __getattr__ on static analysis by guarding the runtime implementation behind an if TYPE_CHECKING block and declaring __version__ for type checkers.
- Move the pathlib import into the dynamic library-loading path so static builds avoid unnecessary import overhead.
- Add tests to validate lazy version access and correct AttributeError behavior for missing attributes.

Documentation:
- Update CHANGELOG and HISTORY to document the new import-time behavior, performance figures, and __version__ access semantics.

Tests:
- Extend the test suite with coverage for missing attributes under the new __getattr__ behavior and adjust the version test to reflect collection-time failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -139,6 +139,100 @@ release-notes length in the first place, and are still in
   nothing to win and the figures say so: the field square root is what
   that row is.
 
+### What importing the package costs
+
+- **`__version__` is read on first access rather than at import** (#210).
+  The value comes from where it came from -- the installed distribution
+  metadata, so that `pyproject.toml` stays the only place a release bumps
+  -- and a module-level `__getattr__` (PEP 562) is what defers reading
+  it. `importlib.metadata` pulls `email`, `json` and `inspect` behind it,
+  among others.
+- **`pathlib` is imported where it is used**, which is the dynamic branch
+  of `_load_lib`. A static extension returns from `hasattr(module,
+  "lib")` two lines earlier, and a static extension is what every
+  platform of the matrix ships by default.
+- **It takes both, and deferring `import pathlib` alone would have bought
+  nothing.** `importlib.metadata` imports `pathlib` itself, and
+  `import pathlib` was the statement *above* the one that read the
+  version, so on the file as it was `pathlib` arrived whether that first
+  line ran or not: the row below measures it at 0.09 milliseconds, which
+  is nothing. The 2.27 that deferring it is worth here is a saving this
+  branch creates rather than one that was lying there.
+- Every figure below is one session, one build, `__init__.py` the only
+  thing that differs, minimum of 12 fresh interpreters each -- an Apple
+  M5, macOS 26.6, arm64, CPython 3.14.6. It is the only place this
+  decomposition is written down; the docstrings that quote a figure send
+  the reader here rather than repeating the split.
+
+  ```shell
+  # after each swap, run it once and throw that answer away: a changed
+  # __init__.py invalidates its .pyc, and compiling this file is about a
+  # millisecond -- most of the number being reported for the branch
+  python -c "import time, importlib
+  t = time.perf_counter(); importlib.import_module('btclib_secp256k1')
+  print((time.perf_counter() - t) * 1e3)"
+  ```
+
+  | | ms |
+  | --- | --- |
+  | as it is now, both deferred | **1.69** |
+  | with `import pathlib` back at module level | 3.96 |
+  | with the `importlib.metadata` import back instead | 10.98 |
+  | the file as it was: both, and the `version()` call | 15.85 |
+  | the file as it was, with `import pathlib` deferred alone | 15.76 |
+  | as it is now, then reading `__version__` | 15.59 |
+  | the file as it was, then reading `__version__` | 15.97 |
+  | `_btclib_secp256k1`, the extension, either way | 0.58 |
+
+  The middle rows are marginal costs over this branch, each measured by
+  putting one statement back, and they do not add up to the fourth: the
+  9.3 of the metadata import already contains `pathlib`'s 2.27, and what
+  is left of the 15.85 is the first `version()` call, 4.9 of it.
+
+  The last two rows before the extension are the same number and not a
+  saving of 0.4: a caller that reads the version pays what it paid, and
+  which of the two comes out ahead is the drift of the machine between
+  two batches.
+
+- **The value is stored into the module's namespace on the way out.**
+  `sys.modules` caches the module `importlib.metadata` is, not the answer
+  `version()` gives, so a `__getattr__` that only defers would leave
+  every read after the first walking the metadata again -- 290
+  microseconds each, where the attribute it replaced was a dict lookup at
+  0.010. Storing it is the usual shape of PEP 562 and makes the second
+  read 0.018. It is also what makes the `dir()` sentence below true: the
+  interpreter does not memoise a module `__getattr__`, so without this
+  the attribute would never appear at all.
+- **The reading of the version is outside the `TYPE_CHECKING` guard**,
+  in `_read_version`, and only the `__getattr__` is inside it. mypy does
+  not check the body of the `else` such a guard takes: with the two
+  working lines in there, `return len(version(...))` from a function
+  annotated `-> str` passes `--strict`, and `warn_unreachable` does not
+  catch it either. Out here the same error is `[return-value]`, which is
+  what `py.typed` and *a wrong annotation propagates* ask for.
+- **The `__getattr__` is behind `if TYPE_CHECKING`, and that is the
+  point of the change surviving the gate.** A module that has one is a
+  module mypy stops checking attribute names on: measured on this tree,
+  `from btclib_secp256k1 import nosuchmodule` and
+  `btclib_secp256k1.typo` both stop being errors under `--strict`, and
+  that is the front door of the package. Declaring `__version__: str` to
+  the type checker and defining the function only at run time keeps both
+  checks, and `tests/test_extension.py` needing a
+  `type: ignore[attr-defined]` to ask for a missing attribute is the
+  proof that it does.
+- **Two tests hold the saving**, `test_import_defers_the_metadata` and
+  `test_import_defers_pathlib`: each imports the package in a subprocess
+  and asserts what is *not* in `sys.modules` afterwards. Nothing else
+  would notice a module-level import added later, here or in anything
+  this package imports, which is how the 13 milliseconds arrived. The
+  second is skipped on a dynamic extension, which reaches `_load_lib`'s
+  other branch and imports `pathlib` legitimately; the matrix runs the
+  suite both ways.
+- **What it costs a caller** is that `dir(btclib_secp256k1)` does not
+  list `__version__` until something reads it, and does afterwards.
+  Nothing else changes: the attribute, the `from … import __version__`,
+  and reading the metadata directly all answer what they answered.
+
 ### Signing under one key
 
 - **`ssa.Signer` builds the BIP340 keypair once** (#153). `ssa.sign` and
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -264,6 +264,22 @@ verification, which is 12.9 microseconds against 12.1, and most of the
 call on `keys._pubkey_cmp_`, which is 0.57 against a C comparison of
 0.087. `lib` is exported for a caller who counts that.
 
+**Importing the package is ten times cheaper**, 1.69 milliseconds against
+15.85 — an Apple M5, macOS 26.6, arm64, CPython 3.14.6, minimum of 12
+fresh interpreters, CHANGELOG.md carrying the command and the table. Two
+imports did it, neither of them the compiled extension, which is 0.58 of
+those milliseconds: `importlib.metadata`, reached at import to fill
+`__version__` and pulling `email`, `json` and `inspect` behind it, and
+`pathlib`, which only the dynamic branch of `_load_lib` uses and which a
+static wheel never reaches. The version is read on first access now, and
+`pathlib` where it is used — and it takes both, `importlib.metadata`
+importing `pathlib` itself. What a caller has to know is one thing:
+`dir(btclib_secp256k1)` does not list `__version__` until something reads
+it, and does afterwards. Reading it — as the attribute, as
+`from btclib_secp256k1 import __version__`, or straight out of the
+metadata — answers what it always answered, and costs what it always
+cost, the first read and every one after it.
+
 ## v0.8.0.2
 
 Non-breaking, and two additions for a caller that verifies signatures it
--- a/btclib_secp256k1/__init__.py
+++ b/btclib_secp256k1/__init__.py
@@ -43,15 +43,83 @@ was just serialized -- and for a compressed key that parse is a field
 square root.
 """
 
-import pathlib
-from importlib.metadata import version
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import _btclib_secp256k1
 
-# read from the installed distribution metadata, so that the version in
-# pyproject.toml stays the only place to bump at release time
-__version__ = version("btclib_secp256k1")
+
+def _read_version() -> str:
+    """Read the version out of the installed distribution metadata.
+
+    That is where it comes from so that the version in pyproject.toml
+    stays the only place to bump at release time. Where it is read is
+    the `__getattr__` below; this is only the reading of it, and it is
+    out here rather than inside that function because mypy does not
+    check the body of the `else` a `TYPE_CHECKING` guard takes -- a
+    `return len(...)` from a function annotated `-> str` passes
+    `--strict` in there, and is caught as `[return-value]` out here.
+
+    `importlib.metadata` is not a small module to reach for -- `email`,
+    `json` and `inspect` are three of the ones it pulls behind it -- so
+    the import is deferred to the one call that needs it.
+
+    Returns:
+        The version of the installed distribution.
+    """
+    from importlib.metadata import version  # noqa: PLC0415
+
+    return version("btclib_secp256k1")
+
+
+if TYPE_CHECKING:
+    # what a type checker is told, and it is told this instead of the
+    # function below rather than as well: a module with a `__getattr__`
+    # is one mypy stops checking attribute names on altogether, so
+    # `from btclib_secp256k1 import nosuchmodule` and
+    # `btclib_secp256k1.typo` both start passing --strict. That is the
+    # front door of this package, the line every caller writes, and the
+    # two checks are worth the six lines it takes to keep them
+    __version__: str
+else:
+
+    def __getattr__(name: str) -> str:
+        """Build the one attribute this module makes only when asked.
+
+        `__version__` is that attribute, and reading it at import is
+        what made that import 15.85 milliseconds where it is now 1.69 --
+        an Apple M5, macOS 26.6, arm64, CPython 3.14.6, minimum of 12
+        fresh interpreters. CHANGELOG.md carries the command, the rest
+        of the table and what the 14 milliseconds are made of, so that
+        the decomposition is written down once. So the caller that wants
+        the version pays for it and the one that never asks pays
+        nothing; btclib, which depends on this package, is one of
+        those.
+
+        The value is stored into the module's own namespace on the way
+        out, which is the usual shape of PEP 562 and is what keeps the
+        second read a dict lookup rather than another walk of the
+        metadata: `importlib.metadata.version` is 290 microseconds every
+        time it is called, `sys.modules` caching the module and not the
+        answer. So `dir()` does not list `__version__` until something
+        reads it, and does afterwards.
+
+        Args:
+            name: the attribute being looked up.
+
+        Returns:
+            The version of the installed distribution, for
+            `__version__`.
+
+        Raises:
+            AttributeError: for any other name, this being the fallback
+                python calls only once the module itself has none.
+        """
+        if name == "__version__":
+            globals()["__version__"] = value = _read_version()
+            return value
+        msg = f"module {__name__!r} has no attribute {name!r}"
+        raise AttributeError(msg)
+
 
 # an opaque handle to a libsecp256k1 object, as returned by ffi.new: the
 # cffi cdata type is not expressible in the type system, but a named
@@ -110,7 +178,15 @@ def _load_lib(module: Any) -> Any:
         return module.lib
 
     # a dynamic one (cffi ABI mode) has to find, at run time, the shared
-    # object shipped beside it
+    # object shipped beside it. pathlib is imported here rather than at
+    # module level because the line above is where a static build
+    # returns, and a static build is what every platform of the matrix
+    # ships by default: 2.27 milliseconds of an import that is 1.69
+    # milliseconds without it, measured as `__getattr__` says. It is
+    # worth that only alongside the deferral there -- `importlib.metadata`
+    # imports pathlib itself, so this line on its own would save nothing
+    import pathlib  # noqa: PLC0415
+
     path = pathlib.Path(module.__file__).parent
     rejected: list[tuple[str, OSError]] = []
     for suffix in (".dll", ".so", ".dylib"):
--- a/tests/test_extension.py
+++ b/tests/test_extension.py
@@ -20,12 +20,15 @@ mistaken for one.
 import pathlib
 import re
 import shutil
+import subprocess
 import sys
 import types
 
+import _btclib_secp256k1
 import _cffi_backend
 import pytest
 
+import btclib_secp256k1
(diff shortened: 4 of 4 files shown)


fametrano (commented):
Reviewed at 6952c00, against `main` at 75cbb44, both built static from
source in their own worktree. Same machine as the figures in the
description — Apple M5, macOS 26.6, arm64, CPython 3.14.6.

## What reproduces

Timing `importlib.import_module` inside the process rather than the whole
interpreter, minimum of 12 fresh interpreters per cell:

```shell
python -c "import time,importlib;t=time.perf_counter();\
importlib.import_module('btclib_secp256k1');print((time.perf_counter()-t)*1e3)"
```

| | `main` | this branch |
| --- | --- | --- |
| `import btclib_secp256k1` | 14.72 ms | **1.42 ms** |
| the same, then reading `__version__` | 14.71 ms | 14.75 ms |
| `_btclib_secp256k1` alone | 0.53 ms | 0.53 ms |

So the ten times is there, and the second row is better than the
description claims: reading the version costs what it cost, 14.75 against
14.71, not 15.26 against 14.83.

The `if TYPE_CHECKING` is doing what the commit message says. A probe with
`from btclib_secp256k1 import nosuchmodule`, `btclib_secp256k1.typo` and
`keys.typo` gives three `attr-defined` errors under `--strict` on `main`
and three on this branch; with the `__getattr__` dedented to module level
— the plain spelling — it gives one. And what the import no longer pulls
is measurable directly: after `import btclib_secp256k1` in a fresh
interpreter, `sys.modules` holds none of `importlib.metadata`, `email`,
`json`, `inspect`, `zipfile`, `pathlib` on this branch, against 31 of
those modules on `main`.

583 passed, coverage 100.00% — coverage 7.15 excludes an
`if TYPE_CHECKING:` body on its own, so the declaration needs no pragma.

## Changes required

**1. The `dir()` sentence is false as implemented, in three places.** A
module `__getattr__` is not memoised by the interpreter and nothing here
memoises it, so the attribute never lands in the module dict:

```shell
$ python -c "import btclib_secp256k1 as m
print('__version__' in dir(m), '__version__' in vars(m))
m.__version__
print('__version__' in dir(m), '__version__' in vars(m))"
False False
False False
```

The docstring, the CHANGELOG bullet and the HISTORY paragraph all say
`dir()` does not list it *until something reads it*, which reads as a
promise that reading it puts it there. Either say it never lists it, or
cache the value — see the next point, which the same edit answers.

**2. Every read re-parses the metadata: 290 µs, each time.** Warm — the
first read pays the 13.3 ms of importing `importlib.metadata`, and each
one after it pays the lookup again:

```shell
$ python -c "import timeit, btclib_secp256k1 as m, importlib.metadata as md
_ = m.__version__   # warm sys.modules
print(min(timeit.repeat(lambda: m.__version__, number=100, repeat=7))/100*1e6)
print(min(timeit.repeat(lambda: md.version('btclib_secp256k1'), number=100, repeat=7))/100*1e6)"
290.2
290.3
```

On `main` the second read was a dict lookup, 0.010 µs. So *a caller that
reads the version pays what it paid* holds for the first read and not for
the N after it, which now cost 13.3 ms + 0.29·(N−1) rather than 13.3 ms.
Sourcery's caching remark is right, though not for the reason it gives:
the cost is not the module import, which `sys.modules` already caches, it
is `version()` walking the metadata every call. Storing into `globals()`
is the usual PEP 562 shape and settles both this and point 1 — measured
here, 0.018 µs per read after the first, suite green, coverage still
100.00%.

**3. The import figure is two different numbers inside one branch.** The
comment in `_load_lib` says *2.3 milliseconds of an import that is 1.4
without them*; CHANGELOG.md, HISTORY.md and the description say 1.55. By
the instrument above the comment's number is the one that reproduces
(1.42, and #210 measured 1.43), so it is the 1.55 in the two release
files that is the outlier — but whichever survives, it wants to be one
number in four places. Two smaller things in that same clause: *them* is
a single import, and 1.4 carries no unit.

Sourcery's suggestion on the CHANGELOG line is worth declining, by the
way: *2.3 milliseconds nothing in the wheel reached* says the code was
never reached, its *a duration that nothing in the wheel ever reached*
says something else and untrue.

## Suggested

**4. The `else:` body is the one piece of this package mypy does not
check.** Injecting a plain error into it and running the type checker over
the file:

```shell
$ sed -i '' 's/return version("btclib_secp256k1")/return len(version("btclib_secp256k1"))/' \
      btclib_secp256k1/__init__.py
$ mypy btclib_secp256k1/__init__.py
Success: no issues found in 1 source file
```

`warn_unreachable` does not catch it either — mypy special-cases the
branch. The same error one indentation level out, in a module-level
helper, is caught as `[return-value]`. Given `py.typed` and *a wrong
annotation propagates* in pyproject.toml, the two working lines are worth
having outside the guard, which the guard does not need to hold: only the
`__getattr__` definition has to be invisible to mypy, not the version
read. Combined with point 2:

```python
def _read_version() -> str:
    """..."""
    from importlib.metadata import version  # noqa: PLC0415

    return version("btclib_secp256k1")


if TYPE_CHECKING:
    __version__: str
else:

    def __getattr__(name: str) -> str:
        """..."""
        if name == "__version__":
            globals()["__version__"] = value = _read_version()
            return value
        msg = f"module {__name__!r} has no attribute {name!r}"
        raise AttributeError(msg)
```

Measured on that shape: the injected error is caught, the three
`attr-defined` errors of the probe are still three, 583 pass at 100.00%,
and the import stays where it was — 2.91 ms against 2.72 by the
whole-interpreter instrument, where the drift between two batches of the
same case was larger than the gap.

**5. Nothing ratchets what the change bought.** A module-level import
added later — here or in any module this one imports — puts the 13 ms
back with every gate green, which is how the 13 ms arrived in the first
place. The property is directly assertable, and the measurement above is
the test:

```python
def test_import_defers_the_metadata() -> None:
    code = "import sys, btclib_secp256k1; print('importlib.metadata' in sys.modules)"
    out = subprocess.run([sys.executable, "-c", code], capture_output=True, text=True, check=True)
    assert out.stdout.strip() == "False"
```

`pathlib` can be asserted the same way, but only under
`hasattr(_btclib_secp256k1, "lib")`: the dynamic wheel reaches
`_load_lib`'s second branch and imports it legitimately, and the matrix
runs the suite both ways.

**6. Not for this branch.** `typing` is now the largest thing the import
pulls: `-X importtime` gives it 1.03 ms cumulative against 0.75 for the
extension, so it is the bigger half of the 1.42 that is left, for `Any`
and `TYPE_CHECKING`. Whether it can be had without importing `typing` at
a 3.10 floor is its own question.

## Verdict

Not an ACK yet. The change itself is right and the figures hold; 1 and 3
are prose the tree contradicts, and 2 decides whether the *pays what it
paid* claim is true past the first read. With those three settled — and 4
and 5, which are a few lines each — ACK.

fametrano:
Pushed at 3c5a365. All five points taken; nothing declined.

**1 and 2 are one edit.** The value is stored into `globals()` on the way
out, which is the usual PEP 562 shape. Reproduced your 290 µs and the fix:

```shell
$ python -c "import timeit, btclib_secp256k1 as m
_ = m.__version__
print(min(timeit.repeat(lambda: m.__version__, number=1000, repeat=9))/1000*1e6)"
0.018
```

and `dir()` / `vars()` now go `False False` → `True True` across the first
read, so the sentence in the docstring, the CHANGELOG and HISTORY is true
as written rather than a promise the tree contradicts. You were right that
Sourcery had the conclusion and not the reason.

**3.** One number in four places, and it is now one command over one build
with `main`'s `__init__.py` the only thing that differs — the same
instrument you used, minimum of 12 fresh interpreters:

| | main | this branch |
| --- | --- | --- |
| `import btclib_secp256k1` | 14.61 ms | **1.55 ms** |
| the same, then reading `__version__` | 14.92 ms | 14.69 ms |
| `_btclib_secp256k1` alone | 0.53 ms | 0.53 ms |

So 1.55 is what reproduces here, where 1.42 did for you; the extension
figure matches yours exactly, which is what makes me think the gap is the
machine and not the instrument. The command is in CHANGELOG.md now, so the
next reader re-derives instead of choosing between two numbers. Attributed
the same way: 10.9 for `importlib.metadata` and 2.2 for `pathlib`, each
measured by putting that one import back at module level. The `_load_lib`
comment now says *2.2 milliseconds of an import that is 1.55 milliseconds
without it* — unit, singular, same figure.

**4.** Confirmed the hole before closing it, with your injection:

```
# return len(version(...)) inside the else
Success: no issues found in 1 source file
# the same statement, in _read_version at module level
error: Incompatible return value type (got "int", expected "str")  [return-value]
```

`_read_version` is at module level and only the `__getattr__` is inside the
guard, as you spelled it.

**5.** Both ratchets, and both were handed something bad first: with
`import pathlib` and `from importlib.metadata import version` put back at
module level, `test_import_defers_the_metadata` and
`test_import_defers_pathlib` both fail. The pathlib one is skipped under
`hasattr(_btclib_secp256k1, "lib")`, as you said. The helper takes a
subprocess because this interpreter has half the standard library in
`sys.modules` by the time the suite runs.

**6** left alone, as not-for-this-branch. Worth an issue of its own:
`typing` is imported for `Any` and `TYPE_CHECKING`, and `Any` is in the
`CData` alias every module reads, so it is not obviously removable at a
3.10 floor.

Sourcery's CHANGELOG suggestion declined for your reason: *a duration that
nothing in the wheel ever reached* says something else, and untrue.

585 passed, coverage 100.00%, every hook passes.

fametrano (commented):
Re-reviewed at 3c5a365, in a build of its own. All five points land, and
each of the five was checked against the tree rather than against the
description.

## What holds

- **The mypy hole is closed where it was open.** The injection that passed
  inside the `else` is now `btclib_secp256k1/__init__.py:71: error:
  Incompatible return value type (got "int", expected "str")
  [return-value]`, and the probe still gives three `attr-defined` errors
  under `--strict`, so `_read_version` moving out cost nothing the guard
  was there for.
- **The store into `globals()` does what the prose now says.** `dir()` and
  `vars()` go `False False` → `True True` across the first read, and the
  read after it is 0.017 µs against the 290 µs of another `version()`
  walk.
- **Both ratchets bite.** With `import pathlib` and
  `from importlib.metadata import version` put back at module level,
  `test_import_defers_the_metadata` and `test_import_defers_pathlib` both
  fail; on the tree as pushed both pass, and the `skipif` is right — a
  dynamic extension reaches the branch that imports `pathlib`.
- 585 passed, coverage 100.00%, every check green on this head.
- The figures reproduce. Minimum of 12 fresh interpreters, one build with
  `__init__.py` the only thing swapped, `__pycache__` warmed (see below):
  1.56 ms for this branch against 17.26 for `main`'s file, 0.53–0.55 for
  the extension. `main` is high because this machine has had a load
  average of 5 to 11 all morning; in a quieter batch the same pair was
  1.42 / 14.72 and then 1.49 / 15.30. The branch's number is the one that
  is stable, and it is your 1.55.

Which also settles the 1.42 against 1.55: it is neither the machine nor the
instrument — I measured on this Mac, the same one — and the swapped-file
build says 6952c00 and 3c5a365 import in the same time (2.44 against 2.41
in one batch). The spread is the condition of the machine and the state of
`__pycache__`, nothing in the code.

## Before it lands

**1. The published recipe needs the line that says to discard the first
run.** Swapping `__init__.py` invalidates the module's `.pyc`, and
compiling this file is about a millisecond — 40% of the number being
reported. Ten interleaved rounds, cold meaning the `.pyc` was removed
before the run:

| | cold | warm |
| --- | --- | --- |
| `main`'s file | 15.99 ms | 15.30 ms |
| this branch | **2.47 ms** | **1.49 ms** |

A reader who follows the block in CHANGELOG.md literally — swap, run,
read — gets 2.5 ms for the branch and 16 for `main`, and has no way to
know why the headline row says 1.55. One clause in the recipe fixes it:
run once and throw the answer away after each swap, or `compileall` the
package.

**2. `pathlib` was not 2.2 milliseconds that `main` paid for nothing.** It
was 2.2 milliseconds `main` paid *twice over* — `importlib.metadata`
imports `pathlib` itself:

```shell
$ python -c "import importlib.metadata, sys
print([m for m in ('pathlib','email','json','inspect','zipfile') if m in sys.modules])"
['pathlib', 'email', 'json', 'inspect', 'zipfile']
```

So deferring `import pathlib` on `main` would have bought nothing
measurable: the line below it pulled `pathlib` in regardless. The 2.2 ms
is a saving the *version* deferral created, and only the two together get
the wheel out of `pathlib` — which is exactly what the branch does, so
this is a clause in the CHANGELOG bullet and not a change of substance.
The comment in `_load_lib` is already right, saying 2.2 of an import that
is 1.55 without it: that is the branch, and the branch is where it lives.

## Suggested

**3. Say which measurement the 10.9 is.** Minimum of 12 fresh
interpreters, each import or call timed on its own:

| | |
| --- | --- |
| `import importlib.metadata` | 9.63 ms |
| the first `version()` call after it | 4.57 ms more |
| `import pathlib` | 2.29 ms |

`main` imports `pathlib` first, so that statement did pay its own 2.29 and
the metadata import paid the rest — which is what makes 2.2 + 10.9 add up
to the 13.06 residual. But the 10.9 matches the marginal cost of the
module import measured over *this* branch (10.11 here), where `pathlib` is
inside it and the 4.6 ms of the call is not. Two lines measured two ways,
in other words, and the arithmetic works because the double-counted
`pathlib` and the uncounted first call are about the same size. Whichever
convention you keep, `__getattr__`'s docstring, the CHANGELOG bullet and
`test_import_defers_the_metadata`'s docstring hold the same number in
three places, so it is worth being one convention.

**4. The metadata ratchet can be made immune to the environment.**
`test_import_defers_the_metadata` asserts against absolute absence, so
anything a future interpreter or a stray `.pth` preloads becomes a failure
of this package. `_imported_modules("sys")` is the baseline that answers
it — subtract it and the assertion is about what *this* import adds.
Nothing in it changes today: a bare interpreter here holds none of the
five names.

**5.** The 290 µs and the 0.010 against 0.018 all reproduce, and point 6
was right to leave alone — `Any` is in the `CData` alias every module
reads, so `typing` does not come out at a 3.10 floor without a change of
its own. Worth the issue you suggest, not this branch.

## Verdict

ACK 3c5a365 on the code and the tests — the shape is right, the guard
keeps what it was for, and the two new tests are what makes the saving
survive the next contributor. 1 and 2 are prose, and 2 is a sentence the
tree contradicts, so both are worth having in before it lands; neither
needs another review round from me.

fametrano (commented):
Re-reviewed at 889d5aa, in a third build of its own. The whole table
re-derived rather than read.

## What re-derives

Your protocol, including the warm-up the recipe now carries, minimum of 12
fresh interpreters, one build with `__init__.py` the only thing swapped —
this machine at a load average of 4, which is why every row of mine runs
5 to 15% high:

| | mine | the tree |
| --- | --- | --- |
| as it is now | 1.54 | 1.61 |
| with `import pathlib` back | 4.10 | 3.82 |
| with the metadata import back | 11.15 | 10.49 |
| `main`'s file | 16.45 | 15.59 |
| reading `__version__` on top of the first row | 16.43 | 14.69 |
| the extension | 0.61 | 0.53 |
| the first `version()` call alone | 5.13 | 4.30 |

and the decomposition closes on mine as well as it does on yours: 9.61 for
the metadata import with `pathlib` inside it, plus 5.13 for the first
call, against a residual of 14.91.

The rest, checked rather than assumed: 585 passed, coverage 100.00%, mypy
clean over 35 files, the probe still three `attr-defined` errors under
`--strict`. And the two ratchets against three regressions, one variant
each — both imports eager, both tests fail; `pathlib` eager alone, only
`test_import_defers_pathlib` fails; the metadata import eager alone, both
fail, which is precisely the claim the new docstring makes about why the
two are separate tests rather than one.

## Changes required

**The ordinals in the new bullet are the wrong way round.** It reads

> deferring the second import while the first stayed would have bought
> nothing measurable: the line above it pulled `pathlib` in regardless

and in `main`'s file the first statement is `import pathlib`, the second
`from importlib.metadata import version`. So the case that bought nothing
is deferring the *first* while the *second* stayed, and what pulled
`pathlib` in regardless is the line *below* it. Measured, same build, pyc
warmed, minimum of 12:

| | ms |
| --- | --- |
| `main`'s two statements | 15.73 |
| `pathlib` deferred, the read left eager | **16.07** |
| the read deferred, `pathlib` left eager | 3.93 |
| both deferred, which is this branch | 1.48 |

Row two is the claim — nothing, and 0.34 the wrong way at that — and the
bullet as written points at row three, which saves 11.8 milliseconds. The
file order and the bullet order are opposite, so naming the two statements
instead of counting them is what stops this coming back.

## Suggested

**1. The `14.69` row invites a saving that is not there.** Against
`main`'s `15.59` it reads as though reading the version now costs 0.9
milliseconds less than it did; measured here the two are the same number,
16.43 against 16.45, which is exactly what the prose beside the table
claims. A clause under the table — the same number within the drift —
keeps the next reader from taking the difference for a result.

**2. `8.9` and `4.3` now live in three places**: `__getattr__`'s
docstring, the paragraph under the CHANGELOG table, and
`test_import_defers_the_metadata`'s docstring. Nothing checks any of them,
so the next re-derivation is three edits. The docstring already sends the
reader to CHANGELOG.md for the command and the rest of the table; sending
them there for the split too, and keeping 15.59 against 1.61 where it is,
would leave one site holding the decomposition.

## Verdict

ACK 889d5aa on the code and the tests. They are unchanged in substance
since the ACK I gave 3c5a365, and the change since then is the ratchets
asserting what this import *adds* rather than absolute absence, which is
better than what I suggested and which I checked against three
regressions.

The required point is one sentence of CHANGELOG.md and the only thing I
would not land as it stands — the tree contradicts it, and it is the kind
of sentence somebody re-derives a year from now. With it reordered:
merge.

fametrano:
Pushed at b2b22be. The required point is in, and both suggestions with it.

**The ordinals were backwards, and naming the statements is what fixes
it.** Measured here, same build, `.pyc` warmed, minimum of 12:

| | ms |
| --- | --- |
| both statements, as they were | 15.85 |
| `import pathlib` deferred, the version read still eager | 15.76 |
| the version read deferred, `import pathlib` still eager | 3.96 |
| both deferred, which is this branch | 1.69 |

Row two is the claim — 0.09 milliseconds, which is nothing — and the
bullet now says *deferring `import pathlib` alone*, naming the statement
and the fact that it was the one **above** the version read, rather than
counting either.

**Suggestion 1.** The two `__version__` rows are 15.59 and 15.97 in this
session, so the branch comes out ahead by 0.4 rather than behind — which
is the same number and the drift, and a clause under the table now says
exactly that rather than leaving a reader to take a difference for a
result.

**Suggestion 2.** The split lives in CHANGELOG.md alone now. The
`__getattr__` docstring and `test_import_defers_the_metadata` both quote
the headline pair and send the reader there for what the milliseconds
are made of; `_load_lib`'s comment keeps its own 2.27, that being the
line it annotates.

Everything is one session, including the rows that were two: the earlier
table mixed a 15.59 for `main` with a 3.82 for a variant measured in
another batch, which is precisely what you flagged on #215. 15.85 /
15.76 / 3.96 / 10.98 / 1.69 / 15.59 / 15.97 / 0.58, one run.

585 passed, coverage 100.00%, every hook passes.

PR #214, opened 2026-08-16
Stop a header on the build machine from changing what ships

### LINKED ISSUE #211: The vendored library is configured by the machine that builds it, not only by the source
## What is asked

`build_c` names seven `SECP256K1_ENABLE_MODULE_*` and six
`SECP256K1_BUILD_*` explicitly, with a comment saying why: "upstream
defaults are not part of its API". Four options it does not name decide
how fast the library it produces is, and one of them is not a default at
all — it is a question CMake asks the machine doing the build.

## The one that is a defect

`SECP256K1_VALGRIND` defaults to `AUTO`, and `AUTO` means
`find_package(Valgrind)`:

```cmake
if(SECP256K1_VALGRIND)
  find_package(Valgrind MODULE)
  if(Valgrind_FOUND)
    set(SECP256K1_VALGRIND ON)
    include_directories(${Valgrind_INCLUDE_DIR})
    add_compile_definitions(VALGRIND)
```

With `VALGRIND` defined, `src/checkmem.h` turns
`SECP256K1_CHECKMEM_DEFINE` and its siblings into valgrind client
requests instead of the no-ops they otherwise are. So a runner with
`valgrind/memcheck.h` installed produces a different library from the
same source at the same commit — quietly, and with nothing in the wheel
saying which one it is.

This repository's premise is that the package behaves identically
everywhere, and the matrix is as large as it is for exactly that reason.
An option that reads the build machine is the one kind of thing that
premise cannot absorb. `-DSECP256K1_VALGRIND=OFF` makes the wheel a
function of the source alone.

**What is not claimed here**: how much the instrumentation costs. Outside
valgrind a client request is a handful of instructions, so the answer is
probably "little" — and measuring it needs a machine with the headers,
which is not what this issue turns on. What it turns on is that the
answer is currently a property of the runner.

## The three that are defaults, and are fine

Read off `build/secp256k1/CMakeCache.txt` of a local build:

| option | value | what it decides |
| --- | --- | --- |
| `SECP256K1_ASM` | `AUTO` | the x86-64 field arithmetic. Where the check fails, `AUTO` falls back to `OFF` *in silence*; a pinned `x86_64` makes the same case a build error |
| `SECP256K1_ECMULT_WINDOW_SIZE` | `15` | the verification table |
| `SECP256K1_ECMULT_GEN_KB` | `86` | the signing table, reaching the compiler as `COMB_BLOCKS=43 COMB_TEETH=6` |

All three are what upstream recommends for a desktop, so naming them
changes nothing today. What it would change is that a submodule bump
moving one of them says so in a diff instead of in the wheels.

`SECP256K1_ASM` is the awkward one and is why this issue does not simply
ask for four lines: it cannot be pinned to `x86_64` in the cells that
build macOS `universal2` in one pass or that cross-compile with mingw, so
the honest form is per-target and belongs to whoever writes it, not to a
one-line sweep. The two table sizes are a policy question — freeze
upstream's number, or keep following it — and that is the maintainer's.

## What this issue asks for

The `VALGRIND` line. The rest is recorded here so that it is recorded
somewhere.

fametrano:
One line of CMake, and the reason it is one line rather than four.

## The defect

`SECP256K1_VALGRIND` defaults to `AUTO`, and `AUTO` is
`find_package(Valgrind)` — the only `find_package` in the vendored CMake.
Where the header is on an include path the library is compiled with
`-DVALGRIND`, which turns the `SECP256K1_CHECKMEM_*` of
`src/checkmem.h` from no-ops into valgrind client requests, and
`SECP256K1_BUILD_CTIME_TESTS` takes its default from it besides. So a
runner that happens to have one ships a different library from the same
commit, and the wheel says nothing about which one it is.

That is the one kind of thing this repository's premise cannot absorb —
the matrix is as large as it is precisely so that the package is the same
everywhere. `-DSECP256K1_VALGRIND=OFF` makes what a wheel is compiled
with a function of the source alone.

## Reproduced, not argued from the CMake

Configured against a directory holding a `valgrind/memcheck.h`:

```
=== SECP256K1_VALGRIND=AUTO ===
-- Found Valgrind: …/fakevalgrind
SECP256K1_VALGRIND:STRING=AUTO
Valgrind_INCLUDE_DIR:PATH=…/fakevalgrind
  -> -DVALGRIND is in src/CMakeFiles/secp256k1.dir/flags.make

=== SECP256K1_VALGRIND=OFF ===
SECP256K1_VALGRIND:STRING=OFF
  -> no -DVALGRIND: find_package never runs
```

An empty header is enough — the module's compile check only rejects
`NVALGRIND` — so the trigger is a header on an include path and not an
installed valgrind. After the change, through the real path (`uv sync
--reinstall-package btclib_secp256k1 --no-cache`), `CMakeCache.txt` reads
`SECP256K1_VALGRIND:STRING=OFF` and has no `Valgrind_INCLUDE_DIR` entry.

## What such a wheel actually executes

It runs rather than merely carrying the instrumentation, which is the one
thing the previous revision of this description got wrong.
`SECP256K1_CHECKMEM_RUNNING()` is itself a client request under
`VALGRIND` — `src/checkmem.h` defines it as
`(VALGRIND_MAKE_MEM_DEFINED(NULL, 0) != 0)`, deliberately, memcheck
having to be detected specifically — and it is the left operand of the
`&&` at `src/secp256k1.c:104`, behind no flag and no `VERIFY`.
`secp256k1_context_create` reaches that line twice, so a `-DVALGRIND`
wheel executes two of them at import and none afterwards: the other call
site is `secp256k1_declassify` (`src/secp256k1.c:254`), behind
`ctx->declassify`, which line 104 refuses to set outside memcheck at all.
Everything else is `CHECK_VERIFY`, `MSAN_DEFINE`, or under `#ifdef
VERIFY`.

A handful of instructions, once. **The pin is not about the cost** — it is
about the wheel being a function of the source, and no measurement of the
cost is claimed here.

## The three that are not pinned, and why not

`SECP256K1_ASM`, `SECP256K1_ECMULT_WINDOW_SIZE` and
`SECP256K1_ECMULT_GEN_KB` are the options that decide how fast the
library is, and all three are at upstream's desktop defaults — `AUTO`,
`15`, and `86` KiB reaching the compiler as `COMB_BLOCKS=43
COMB_TEETH=6`. Naming them would change nothing today, which is the
argument both for and against doing it, and it is a policy call rather
than a defect.

`ASM` asks the build machine a question too: `AUTO` is a compile check
that falls back to `OFF` **in silence**. It stays `AUTO` because it cannot
be pinned in a line — the mingw cell cross-compiles, and the macOS
`universal2` one compiles two architectures in a single pass, which is the
`x86_64;arm64` that `target_architecture_options` already names. That note
now sits *after* the configure list rather than inside it, for exactly
that reason: in the list it read as annotating the line it contradicts.

## Scope

`scripts/cffi_build.py`, plus `scripts/README.md` — the file CLAUDE.md
puts the build backend's documentation in, which now describes this line
and, in passing, names the seven modules the configure line requests
rather than six. CHANGELOG.md under `### CI`, where the other
vendored-library entry is. Nothing in HISTORY.md: a user acts on nothing
here and gains a wheel that is what its source says, which is the record
rather than the release note. `.secrets.baseline` moves one line number,
the reviewed entry unchanged.

The audit is the configure summary, not a test: `scripts/` is outside
`[tool.coverage.run] source`, so the coverage gate says nothing about this
line, while every wheel job's log carries
`Valgrind .............................. OFF`.

Suite green (582 tests), coverage gate at 100.00%, every hook passes.

Closes #211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1405,6 +1405,56 @@ release-notes length in the first place, and are still in
   against a 8.7 and 10.2 baseline — while the macOS images, whose default
   shell is bash, had already dropped to 1.5 and 1.2. A step that is green
   while doing nothing is why the comment above it now carries this
+- **The vendored library is no longer configured by what happens to be
+  installed on the machine that builds it** (#211).
+  `SECP256K1_VALGRIND` defaults to `AUTO`, and `AUTO` is
+  `find_package(Valgrind)` — the only `find_package` in the vendored
+  CMake. Where the header is on an include path, the library is compiled
+  with `-DVALGRIND`, which turns the `SECP256K1_CHECKMEM_*` of
+  `src/checkmem.h` from no-ops into valgrind client requests, and
+  `SECP256K1_BUILD_CTIME_TESTS` takes its default from it besides. So a
+  runner that happened to have one shipped a different library from the
+  same commit, in a wheel that says nothing about which one it is.
+  `scripts/cffi_build.py` passes `-DSECP256K1_VALGRIND=OFF` now.
+- **The defect was reproduced before it was fixed** rather than argued
+  from the CMake: configured against a directory holding a
+  `valgrind/memcheck.h`, `AUTO` reports `-- Found Valgrind`, puts
+  `Valgrind_INCLUDE_DIR` in the cache and `-DVALGRIND` in
+  `src/CMakeFiles/secp256k1.dir/flags.make`; `OFF` does none of the
+  three, `find_package` not running at all. An empty header is enough,
+  the module's compile check only rejecting `NVALGRIND`, so the trigger
+  is a header on an include path and not an installed valgrind. Every
+  wheel job's configure summary carries the answer —
+  `Valgrind .............................. OFF` — which is the audit,
+  `scripts/` being outside the coverage gate.
+- **The instrumentation in such a wheel runs rather than merely being
+  present, and the pin is still not about what it costs.**
+  `SECP256K1_CHECKMEM_RUNNING()` is itself a client request under
+  `VALGRIND` — `src/checkmem.h` defines it as
+  `(VALGRIND_MAKE_MEM_DEFINED(NULL, 0) != 0)` deliberately, memcheck
+  having to be detected specifically rather than valgrind in general —
+  and it is the left operand of the `&&` at `src/secp256k1.c:104`,
+  behind no flag and no `VERIFY`. `secp256k1_context_create` reaches
+  that line twice, so a wheel built with `-DVALGRIND` executes two of
+  them at import and none afterwards: the other call site is
+  `secp256k1_declassify` (`src/secp256k1.c:254`), behind
+  `ctx->declassify`, which line 104 refuses to set outside memcheck at
+  all. Everything else is `CHECK_VERIFY`, `MSAN_DEFINE`, or under
+  `#ifdef VERIFY`. A handful of instructions, once — and what the pin is
+  about is that the wheel be a function of the source.
+- **`SECP256K1_ASM`, `SECP256K1_ECMULT_WINDOW_SIZE` and
+  `SECP256K1_ECMULT_GEN_KB` stay at upstream's defaults, and that is now
+  written down** rather than left as three options nobody named. The two
+  table sizes (15, and 86 KiB reaching the compiler as `COMB_BLOCKS=43
+  COMB_TEETH=6`) are what upstream recommends for a desktop. `ASM` asks
+  the build machine a question too — `AUTO` is a compile check, and it
+  falls back to `OFF` in silence where that check fails, which a pinned
+  `x86_64` would turn into a build error — but it cannot be pinned in a
+  line: the macOS `universal2` cell compiles two architectures in one
+  pass and the mingw one cross-compiles, so naming an architecture there
+  would break the cells that are not it. So one machine question stays
+  open, by choice, and the headline above is scoped to the one that is
+  closed.
 - **The vendored library is compiled once per wheel job, not once per
   wheel** (#192). The six wheel jobs were 59.5 of that run's 81.8
   runner-minutes, and each of them built the whole of libsecp256k1 once
--- a/scripts/cffi_build.py
+++ b/scripts/cffi_build.py
@@ -409,6 +409,37 @@ class Secp256k1CFFIExtension(FFIExtension):
             "-DSECP256K1_USE_EXTERNAL_DEFAULT_CALLBACKS=ON",
             f"-DCMAKE_PROJECT_INCLUDE={project_include}",
             f"-DBTCLIB_CALLBACKS={callbacks}",
+            # the one upstream option whose answer is what happens to be
+            # *installed* on the machine: AUTO means
+            # find_package(Valgrind), which is the only find_package in
+            # the vendored CMake, and a header on an include path is
+            # enough -- the module's compile check only rejects
+            # NVALGRIND, so valgrind itself need not be there. Where one
+            # is, the library is compiled with -DVALGRIND, which turns
+            # the SECP256K1_CHECKMEM_* of src/checkmem.h from no-ops into
+            # valgrind client requests, and SECP256K1_BUILD_CTIME_TESTS
+            # defaults to it besides (already pinned OFF below). So a
+            # runner that happens to have a header ships a different
+            # library from the same commit, in a wheel that says nothing
+            # about which one it is, and this package exists to behave
+            # identically everywhere.
+            #
+            # The instrumentation in such a wheel runs rather than merely
+            # being present: SECP256K1_CHECKMEM_RUNNING() is a client
+            # request under VALGRIND (checkmem.h, deliberately, memcheck
+            # having to be detected specifically), it is the left operand
+            # of the && that secp256k1_context_preallocated_size guards
+            # its DECLASSIFY flag with, and secp256k1_context_create
+            # reaches that function twice. So two of them at import, and
+            # none afterwards: the other call site is
+            # secp256k1_declassify, behind ctx->declassify, which that
+            # same guard refuses to set outside memcheck at all. A
+            # handful of instructions, once -- and the pin is not about
+            # the cost but about the wheel being a function of the
+            # source. Functions rather than line numbers: a vendored
+            # file's lines move with the next submodule bump and nothing
+            # here would notice, where these names survive it
+            "-DSECP256K1_VALGRIND=OFF",
             # all the modules wrapped by the bindings are requested
             # explicitly: upstream defaults are not part of its API
             # (recovery, in particular, is disabled by default)
@@ -427,6 +458,19 @@ class Secp256k1CFFIExtension(FFIExtension):
             "-DSECP256K1_INSTALL=OFF",
             *self.target_architecture_options(),
         ]
+        # not in that list, and deliberately: SECP256K1_ASM,
+        # SECP256K1_ECMULT_WINDOW_SIZE and SECP256K1_ECMULT_GEN_KB are
+        # the three options that decide how fast the result is, and all
+        # three are left at upstream's defaults. The two table sizes are
+        # what upstream recommends for a desktop. ASM asks the build
+        # machine a question too -- AUTO is a compile check, and it falls
+        # back to OFF in silence where that check fails -- but it cannot
+        # be pinned in one line: the mingw cell cross-compiles, and the
+        # macOS universal2 one compiles two architectures in a single
+        # pass, which is the `x86_64;arm64` target_architecture_options
+        # names above. Pinning an architecture here would contradict it.
+        # #211 records the values, and why they are left alone
+
         if cross_compile:
             # the toolchain file is the vendored one, upstream tested
             toolchain = self.wd / "cmake" / "x86_64-w64-mingw32.toolchain.cmake"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -148,7 +148,7 @@
         "filename": "scripts/cffi_build.py",
         "hashed_secret": "9d8c445671f8c7fb72e1cab426822fdae46ad55f",
         "is_verified": false,
-        "line_number": 422
+        "line_number": 453
       }
     ],
     "tests/test_keys.py": [
@@ -517,5 +517,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-16T19:03:44Z"
+  "generated_at": "2026-08-17T08:06:26Z"
 }
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -59,11 +59,17 @@ The cffi build description named in `cffi_modules`, exposing the
 **Build the vendored library.** CMake, on every platform, out of tree
 into `build/secp256k1`: the submodule is only ever read from. The
 configure line requests each module the bindings wrap
-(`ecdh`, `recovery`, `extrakeys`, `schnorrsig`, `musig`, `ellswift`)
-explicitly, rather than relying on upstream defaults, which are not part
-of upstream's API and which leave `recovery` off. Upstream's own tests,
-benchmarks and install rules are all turned off. A stale CMake cache is
-deleted first, because it remembers the previous configuration.
+(`ecdh`, `recovery`, `extrakeys`, `schnorrsig`, `musig`, `ellswift`,
+`silentpayments`) explicitly, rather than relying on upstream defaults,
+which are not part of upstream's API and which leave `recovery` off.
+One option is named for a different reason: `SECP256K1_VALGRIND` is
+pinned `OFF` because its default answers with the build machine rather
+than with a value — `AUTO` is `find_package(Valgrind)`, so a runner that
+happens to have the header ships a library compiled with `-DVALGRIND`,
+which is a wheel this repository cannot tell from any other. Upstream's
+own tests, benchmarks and install rules are all turned off. A stale
+CMake cache is deleted first, because it remembers the previous
+configuration.
 
 The build also replaces libsecp256k1's default callbacks, which
 `abort()`, with do-nothing ones, so that an illegal input can never take


scripts/cffi_build.py
```diff
@@ -409,6 +409,28 @@ def build_c(self) -> None:
             "-DSECP256K1_USE_EXTERNAL_DEFAULT_CALLBACKS=ON",
             f"-DCMAKE_PROJECT_INCLUDE={project_include}",
             f"-DBTCLIB_CALLBACKS={callbacks}",
+            # the one upstream option that asks the build machine a
```
fametrano: **"the one upstream option that asks the build machine a question rather than carrying a default" is contradicted twelve lines below.**

`SECP256K1_ASM` defaults to `AUTO`, and `AUTO` is `check_x86_64_assembly()`, a `check_c_source_compiles` — a question put to the build machine, and the paragraph below says so itself: it "falls back to `OFF` in silence where the x86-64 check fails". Reproduced on an arm64 host: the cache keeps `SECP256K1_ASM:STRING=AUTO` and no `USE_ASM_X86_64=1` reaches the compiler, nothing having said a word.

The narrower sentence is true, and is the better one: `find_package(Valgrind)` is the *only* `find_package` in the vendored CMake, so valgrind is the one option whose answer is what happens to be *installed* on the machine, ASM's being a property of the compiler and of the target.

scripts/cffi_build.py
```diff
@@ -409,6 +409,28 @@ def build_c(self) -> None:
...
+            #
+            # SECP256K1_ASM, SECP256K1_ECMULT_WINDOW_SIZE and
+            # SECP256K1_ECMULT_GEN_KB are the other three options that
+            # decide how fast the result is, and they are deliberately
+            # left at upstream's defaults: the two table sizes are what
+            # upstream recommends for a desktop, and ASM cannot be
+            # pinned here in one line -- the macOS universal2 build
+            # compiles two architectures in one pass and the mingw one
+            # cross-compiles, so naming x86_64 would break the cells
+            # that are not it. #211 records the measurements
```
fametrano: **"#211 records the measurements" is not what #211 records.**

The issue records option *values*, read off `build/secp256k1/CMakeCache.txt`, and it declines the measurement in as many words: "What is not claimed here: how much the instrumentation costs ... measuring it needs a machine with the headers". The description repeats the claim ("#211 carries the measurements"), so it is two sites, and the correction belongs in both.

"#211 records the values, and why they are left alone" is the accurate clause; dropping it entirely also works, `Closes #211` being on the commit.

CHANGELOG.md
```diff
@@ -1235,6 +1235,31 @@ release-notes length in the first place, and are still in
   against a 8.7 and 10.2 baseline — while the macOS images, whose default
   shell is bash, had already dropped to 1.5 and 1.2. A step that is green
   while doing nothing is why the comment above it now carries this
+- **The vendored library is no longer configured by the machine that
```
fametrano: **"no longer configured by the machine that builds it" is refuted by the bullet immediately after it.**

That bullet leaves `SECP256K1_ASM` at `AUTO`, and `AUTO` is a probe of the build machine that silently answers `OFF`. So the headline promises what the entry then documents not delivering — and #211's own title is "configured by the machine that builds it, not only by the source", which this closes while one machine question remains open by choice.

Scope it to what changed: *no longer configured by what happens to be installed on the machine that builds it*. The pull request title, which is the subject a fast-forward lands, carries the same slogan and the same overstatement.

fametrano (commented):
Review of 0256157. The one line of CMake is right, and I checked it
rather than reading it: what follows is three prose corrections and
three suggestions, so this is not an ACK yet.

GitHub does not let the author submit *approve* or *request changes* on
their own pull request, so this is a `COMMENT` review; read the required
section as changes requested.

## What was reproduced here, not taken from the description

Vendored tree at `6e2c8bc` (upstream's 0.8.0 preparation).

- `CMakeLists.txt:123-137` is as quoted, and line 131 is the
  `add_compile_definitions(VALGRIND)` the whole entry turns on.
- Configured the vendored tree twice with a stub
  `valgrind/memcheck.h` reachable through `CMAKE_PREFIX_PATH`.
  With `AUTO`: `-- Found Valgrind`, `Valgrind_INCLUDE_DIR` in the cache,
  `-DVALGRIND` in `src/CMakeFiles/secp256k1.dir/flags.make`, and the
  summary line `Preprocessor defined macros ... COMB_TEETH=6 VALGRIND`.
  With `OFF`: none of the three, `find_package` not running. An empty
  header is enough, `Valgrind_WORKS` only rejecting `NVALGRIND`, which
  is worth knowing — the trigger is a header on an include path, not an
  installed valgrind.
- The instrumentation lands in the shipped library and not only in
  upstream's tests: `secp256k1_declassify` in `src/util.h` is
  `SECP256K1_CHECKMEM_DEFINE`, and it is called from `secp256k1.c` and
  from the `extrakeys`, `schnorrsig`, `musig`, `ellswift` and
  `silentpayments` modules — all of them wrapped here.
- The pin is complete on its own axis, which the entry could have said
  and does not: `CMakeLists.txt:142` gives `SECP256K1_BUILD_CTIME_TESTS`
  the default `${SECP256K1_VALGRIND}`, so before this change a runner
  with the headers also built binaries nobody asked for — `build_c`
  already pins that one `OFF`. And `find_package(Valgrind)` is the only
  `find_package` in the vendored CMake, which is the fact the comment
  wants and does not quite state (see below).
- The numbers in the second bullet check out: both configures put
  `ECMULT_WINDOW_SIZE=15 COMB_BLOCKS=43 COMB_TEETH=6` on the compiler
  line, and `CMakeLists.txt:62,68` are where `15` and `86` come from,
  documented as the desktop defaults.
- 27 checks green, `.secrets.baseline` moves the one line number and
  `generated_at`, the entry itself untouched.

## Required

Two of the three are inline. The third:

**`scripts/README.md` is the file that describes this line, and it does
not describe it.** CLAUDE.md puts the build backend's documentation
there, one file at a time, and the `cffi_build.py` section gives exactly
one reason for naming options — "upstream defaults are not part of
upstream's API, and leave `recovery` off". The pin is a second and
different reason: an option that otherwise answers with the machine.
One sentence in that paragraph. While in it: it names six modules where
the configure line requests seven, `silentpayments` having arrived
without it — pre-existing, and adjacent enough to fix in passing.

## Suggested

**The comment is in the wrong order for its own subject.** It opens
"the one upstream option that asks the build machine a question" and the
option it annotates is twenty-two lines below, behind a paragraph about
three others. The valgrind paragraph next to `-DSECP256K1_VALGRIND=OFF`,
and the ASM/table-size paragraph after the flag, reads in one pass.

**Nothing asserts the flag, and the coverage figure does not cover it.**
`[tool.coverage.run] source = ["btclib_secp256k1"]`, so `scripts/` is
outside the 100% gate: "coverage gate at 100.00%" is true and says
nothing about this line. The cheap audit already exists and could be
named in the description instead of a test — every wheel job's log
carries `Valgrind .............................. OFF` and the macro
list. If it is to be a test, the only form of Sourcery's "centralize the
cache args" worth taking is hoisting the list to a module-level tuple a
test can assert on; a helper that merely wraps it adds indirection and
guarantees nothing.

**Sourcery's other comment — shorten the block, move the rationale to a
doc — should be declined**, and the reason is CONTRIBUTING.md: the
reasoning, negative results included, belongs in the file that carries
the decision. That block is what keeps the next submodule bump from
re-deriving all of this.

## Considered and not raised

`### CI` for a change to the shipped wheel is defensible where it sits,
beside the other vendored-library entry. Nothing in HISTORY.md is right:
a user acts on nothing. No escape hatch for a deliberate valgrind build
is right too, the ctime tests being pinned off anyway.

With the three corrections in, ACK.

fametrano:
Pushed at 87ffd85. The three required are in, both suggestions taken, and
one of your facts is in a different file than the review says.

**Inline 1 — the sentence contradicted twelve lines below.** Taken as you
put it: the narrower claim is the true one and the better one. The comment
now opens *the one upstream option whose answer is what happens to be
installed on the machine*, and names `find_package(Valgrind)` as the only
`find_package` in the vendored CMake — I checked, it is:

```
$ grep -rn "find_package" secp256k1/CMakeLists.txt secp256k1/src/CMakeLists.txt
secp256k1/CMakeLists.txt:127:  find_package(Valgrind MODULE)
```

The ASM paragraph moved out from in front of the flag (your second
suggestion) and now sits after it, saying in its own words that ASM asks
the build machine a question too and falls back to `OFF` in silence.

**Inline 2 — "#211 records the measurements".** Wrong in both sites, and it
was my own issue that declined the measurement. Both now read *#211 records
the values, and why they are left alone*.

**Inline 3 — the headline.** Scoped to yours: *no longer configured by what
happens to be installed on the machine that builds it*, and the entry now
closes by saying that one machine question stays open by choice. The pull
request title carried the same overstatement, so the commit subject is now
**"Stop a header on the build machine from changing what ships"** — which is
also more accurate about the trigger, since an empty header is enough.

**Required 3 — `scripts/README.md`.** One sentence in the `cffi_build.py`
paragraph, saying the second and different reason for naming an option. And
the module list there now says seven, `silentpayments` included.

**One correction to the review.** `secp256k1_declassify` is in
`src/secp256k1.c:254`, not `src/util.h`, and that matters for what the entry
should claim: it is guarded by `ctx->declassify`, which
`SECP256K1_CONTEXT_NONE` does not set and `context.py` never asks for. The
only other `SECP256K1_CHECKMEM_*` in the shipped library are behind `VERIFY`
(`util.h:272`, in `secp256k1_memclear_explicit`) or belong to memory
sanitizer. So in a Release build of *this* package the instrumentation is
reachable exactly once, behind a branch never taken — which strengthens
rather than weakens the framing, and both the comment and the CHANGELOG now
say it: what the pin is about is that the wheel be a function of the source,
not what the instrumentation costs.

**Suggestion 3 — the audit.** Named in the entry rather than made a test, for
your reason: `scripts/` is outside `[tool.coverage.run] source`, so
"coverage gate at 100.00%" says nothing about this line. Every wheel job's
configure summary carries `Valgrind .............................. OFF`, and
that is what a reader checks. I did not hoist the argument list to a
module-level tuple — with no test asserting on it that is indirection for
its own sake, and it is the only form of Sourcery's remark you thought worth
taking, conditional on there being a test.

**Sourcery's other comment declined**, for the CONTRIBUTING.md reason you
give: the reasoning, negative results included, belongs in the file that
carries the decision.

Also folded in the two facts your review adds and the entry did not have:
`SECP256K1_BUILD_CTIME_TESTS` takes its default from `SECP256K1_VALGRIND`
(already pinned `OFF` beside it), and the trigger is a header on an include
path rather than an installed valgrind, `Valgrind_WORKS` only rejecting
`NVALGRIND`.

Re-verified after the change: `uv sync --reinstall-package
btclib_secp256k1 --no-cache`, `CMakeCache.txt` reads
`SECP256K1_VALGRIND:STRING=OFF` and has no `Valgrind_INCLUDE_DIR` entry at
all. 582 passed, coverage 100.00%, every hook passes; `.secrets.baseline`
moves the one line number again.

fametrano (commented):
Re-review of 87ffd85. The three required are in and I checked each
against the tree rather than against the reply; the correction to my
review is right and I had `secp256k1_declassify` in the wrong file.

One thing the new paragraph claims is not true, and it is the paragraph
the last round added, so it is in three sites. That is the whole of what
is left: **ACK the moment it is right.**

## The required one: "the one reachable client request"

Inline on the CHANGELOG, and the same sentence is in the comment
(`scripts/cffi_build.py:426-433`) and in the commit message's third
paragraph. Short form: `src/secp256k1.c:104`, inside
`secp256k1_context_preallocated_size`, evaluates
`SECP256K1_CHECKMEM_RUNNING()`, which `src/checkmem.h:98` defines under
`VALGRIND` as `(VALGRIND_MAKE_MEM_DEFINED(NULL, 0) != 0)` — a client
request, behind no guard, on the one path this package always takes.

The rest of the paragraph survives intact, and I enumerated it to be
sure: every other `SECP256K1_CHECKMEM_*` outside `tests.c` and
`ctime_tests.c` is `CHECK_VERIFY` (`checkmem.h:114`, a no-op without
`VERIFY`) or `MSAN_DEFINE` (`checkmem.h:70`, a no-op without msan) or
`util.h:272` under `#ifdef VERIFY`; the five wrapped modules contain no
`CHECKMEM` of their own, only `secp256k1_declassify` calls, which is the
guarded one.

## What was verified this round

- `find_package(Valgrind MODULE)` at `CMakeLists.txt:127` is the only
  `find_package` in the vendored tree — grepped over every `CMakeLists.txt`
  and `cmake/*.cmake`, not only the two the reply quotes.
- `SECP256K1_BUILD_CTIME_TESTS` takes its default from
  `SECP256K1_VALGRIND` (`CMakeLists.txt:142`), and the flag that neutralizes
  it is fifteen lines below the new one, as the comment says.
- `secp256k1_declassify` is at `src/secp256k1.c:254`, guarded by
  `ctx->declassify`, set at line 138 from `SECP256K1_FLAGS_BIT_CONTEXT_DECLASSIFY`;
  `context.py:18` passes `1`, `SECP256K1_CONTEXT_NONE`, which does not carry it.
- `scripts/README.md` now names the seven modules the configure line
  requests, and the sentence added there is the second reason, correctly
  distinguished from the first.
- 26 checks green on 87ffd85. `.secrets.baseline`: base line 422 and new
  line 445 hold the same source line, the shift being exactly the 23 lines
  added above it.

## Suggested

**The ASM paragraph is now the one comment in the list that does not
annotate the entry below it.** It sits directly above
`*self.target_architecture_options()`, and that call is precisely what
names `x86_64;arm64` for the universal2 build — while the paragraph says
naming `x86_64` would break that cell. Two different things spelled the
same way, adjacent, in a list where every other comment introduces what
follows. Moving it below the closing `]` as a trailing note on the list,
or opening it with "not in this list", removes the collision; the move
out from in front of the flag was the right half of the fix.

**The description still describes the previous revision.** It closes
with "#211 carries the measurements" — the sentence the comment and the
CHANGELOG have just stopped saying — frames valgrind as "the one kind of
thing" under a heading calling the other three "not touched", and its
Scope section says "CHANGELOG.md only" where the diff now has
`scripts/README.md` too. The commit message is right; it is the
description a reader of the merged pull request finds.

CHANGELOG.md
```diff
@@ -1235,6 +1235,48 @@ release-notes length in the first place, and are still in
...
+  `Valgrind_INCLUDE_DIR` in the cache and `-DVALGRIND` in
+  `src/CMakeFiles/secp256k1.dir/flags.make`; `OFF` does none of the
+  three, `find_package` not running at all. An empty header is enough,
+  the module's compile check only rejecting `NVALGRIND`, so the trigger
+  is a header on an include path and not an installed valgrind. Every
+  wheel job's configure summary carries the answer —
+  `Valgrind .............................. OFF` — which is the audit,
+  `scripts/` being outside the coverage gate.
+- **What it costs at run time is close to nothing, and the pin is not
+  about the cost.** In a Release build the one reachable client request
```
fametrano: **There is a second reachable client request, and it is the one that actually runs.**

`src/secp256k1.c:104`, in `secp256k1_context_preallocated_size`:

```c
if (EXPECT(!SECP256K1_CHECKMEM_RUNNING() && (flags & SECP256K1_FLAGS_BIT_CONTEXT_DECLASSIFY), 0)) {
```

`src/checkmem.h:98` defines it under `VALGRIND` as `(VALGRIND_MAKE_MEM_DEFINED(NULL, 0) != 0)` — deliberately a real client request rather than `RUNNING_ON_VALGRIND`, the header saying so in the comment above it, because it has to detect memcheck specifically. It is the left operand of the `&&`, so it is evaluated always, behind no flag and no `VERIFY`.

And it is on the path this package cannot avoid: `context.py:18` calls `secp256k1_context_create(1)`, which evaluates `secp256k1_context_preallocated_size` at `secp256k1.c:144` and again inside `secp256k1_context_preallocated_create` at line 125. Two client requests per context, at import, in every Release build compiled with `-DVALGRIND`.

This makes the entry's point better rather than worse: the instrumentation in such a wheel is not present-but-unreached, it runs — and it is still a handful of instructions, and still not what the pin is about. The clause to drop is "the one reachable"; what is worth adding while rewriting is that `ctx->declassify` cannot be set at all outside memcheck, since the very guard at line 104 rejects the flag — a stronger statement than `context.py` never asking for it.

Three sites: here, `scripts/cffi_build.py:426-433`, and the commit message's third paragraph.

fametrano:
Pushed at 99c35c7. You are right, and the sentence was wrong in the way
that matters most: it said the instrumentation was present-but-unreached
when it runs.

Verified before rewriting it, and your inline comment is exact on every
count:

- `src/checkmem.h:98` — `SECP256K1_CHECKMEM_RUNNING()` is
  `(VALGRIND_MAKE_MEM_DEFINED(NULL, 0) != 0)` under `VALGRIND`, with the
  comment above it saying why it is a client request rather than
  `RUNNING_ON_VALGRIND`.
- `src/secp256k1.c:104` — left operand of the `&&`, so always evaluated,
  behind no flag and no `VERIFY`.
- Two per context, not one: `secp256k1_context_create` calls
  `secp256k1_context_preallocated_size` at line 144 and
  `secp256k1_context_preallocated_create` calls it again at line 125.
- And your stronger statement is the one worth having: `ctx->declassify`
  cannot be set outside memcheck at all, line 104 rejecting the flag, so
  the second call site is unreachable rather than merely unasked-for.

The clause "the one reachable" is gone from all three sites — the comment
in `scripts/cffi_build.py`, the CHANGELOG entry, and the commit message's
third paragraph — and each now says what a `-DVALGRIND` wheel executes:
two client requests at import, none afterwards. The paragraph reads
better for it, as you said: not present-but-unreached, but a handful of
instructions once, and still not what the pin is about.

**Suggestion 1 — the ASM paragraph.** Taken, and the collision was worse
than misplacement: it sat directly above the call that names
`x86_64;arm64`. It is now a trailing note after the closing `]`, opening
with "not in that list, and deliberately", and the universal2 sentence
names `target_architecture_options` explicitly so the two are read
together instead of against each other.

**Suggestion 2 — the description.** Rewritten against 99c35c7: the
"#211 carries the measurements" sentence is gone, the "one kind of thing"
framing is replaced by what a wheel actually executes, the ASM section
says why it is not pinned rather than calling it untouched, and Scope
names `scripts/README.md` and the configure-summary audit.

582 passed, coverage 100.00%, every hook passes; `.secrets.baseline`
moves the one line number again, 445 to 449.

scripts/cffi_build.py
```diff
@@ -409,6 +409,33 @@ def build_c(self) -> None:
...
+            # runner that happens to have a header ships a different
+            # library from the same commit, in a wheel that says nothing
+            # about which one it is, and this package exists to behave
+            # identically everywhere.
+            #
+            # The instrumentation in such a wheel runs rather than merely
+            # being present: SECP256K1_CHECKMEM_RUNNING() is a client
+            # request under VALGRIND (checkmem.h, deliberately, memcheck
+            # having to be detected specifically), it is the left operand
+            # of the && at secp256k1.c:104, and secp256k1_context_create
```
fametrano: Suggestion only, the claim itself being correct: `secp256k1.c:104` pins a comment in this repository to a line number in the submodule, and a submodule bump moves it in silence — nothing here reads that comment, and the entry it belongs to has no test. Naming `secp256k1_context_preallocated_size` (and, four lines down, `secp256k1_declassify` without `:254`) says the same thing and stays true across the bump. It would also be the repository's first `file:line` citation outside `secp256k1/`.

scripts/cffi_build.py
```diff
@@ -427,6 +454,18 @@ def build_c(self) -> None:
...
+        # the three options that decide how fast the result is, and all
+        # three are left at upstream's defaults. The two table sizes are
+        # what upstream recommends for a desktop. ASM asks the build
+        # machine a question too -- AUTO is a compile check, and it falls
+        # back to OFF in silence where that check fails -- but it cannot
+        # be pinned in one line: the mingw cell cross-compiles, and the
+        # macOS universal2 one compiles two architectures in a single
+        # pass, which is the `x86_64;arm64` target_architecture_options
+        # names above. Pinning an architecture here would contradict it.
+        # #211 records the values, and why they are left alone
```
fametrano: Suggestion only: a blank line here. As written the note runs straight into `if cross_compile:`, and by the convention that made it read as annotating `target_architecture_options()` in the list, it now reads as the preamble to the cross-compile branch — which its own penultimate sentence is about. The move out of the list was the fix; this is the last inch of it.

fametrano (commented):
# ACK 99c35c7

The paragraph that was wrong is right, and I checked it against the
vendored source rather than against the reply. Nothing required is left.
Two suggestions below, both cosmetic and neither worth another push on
its own — take them here or in whatever touches these lines next.

GitHub does not let an author approve their own pull request, so this
cannot carry the `APPROVED` state; read the heading as the ACK it says it
is, and land it the way CONTRIBUTING.md says.

## Re-verified on 99c35c7

Every claim the rewritten paragraph makes, at the source:

- `src/checkmem.h:98` — `SECP256K1_CHECKMEM_RUNNING()` is
  `(VALGRIND_MAKE_MEM_DEFINED(NULL, 0) != 0)` under `VALGRIND`, and the
  comment above it gives the reason the entry gives. Without it,
  `checkmem.h:108` makes it `(0)`.
- `src/secp256k1.c:104` — left operand of the `&&`, no flag, no `VERIFY`.
- Twice per context: `secp256k1_context_create` at `secp256k1.c:144`
  and `secp256k1_context_preallocated_create` at `:125`, both calling
  `secp256k1_context_preallocated_size`.
- "and none afterwards" holds for this package and not only for that
  function: `secp256k1_context_create` appears once in the whole of
  `btclib_secp256k1`, at `context.py:18`, and nothing clones or creates
  a second context — `set_illegal_callback`, `set_error_callback` and
  `context_randomize` are the only other context calls, and none of them
  reaches a `CHECKMEM`. The `secp256k1_declassify` calls in
  `secp256k1.c:573,577` and in the five wrapped modules evaluate the
  guard and stop there.
- "which line 104 refuses to set outside memcheck at all" is exactly
  right, including the case the sentence does not have to spell out: in
  a wheel built *without* `-DVALGRIND`, `RUNNING()` is `(0)`, so a
  `DECLASSIFY` context is refused there too.
- `CHECK_VERIFY` is a no-op outside `VERIFY` (`checkmem.h:114`),
  `MSAN_DEFINE` outside msan (`checkmem.h:70`), and `util.h:272` is
  under `#ifdef VERIFY`. That is the whole of the rest.
- The ASM note's new claim: `target_architecture_options` does name
  `x86_64;arm64` for universal2, above it in the same function.
- `.secrets.baseline`: 422 → 449 is exactly the 27 lines the first hunk
  adds above that entry, and the line itself is unchanged.
- 26 checks green.

## Suggested, not required

**A line number in a vendored file is a claim with nothing to keep it
true.** `secp256k1.c:104` is in the comment three times over
(`:104`, `:254`, "line 104"), and the next submodule bump moves those
lines without failing anything — no test reads them, and the CI that
would notice does not exist. It would also be the first such citation in
the repository: nothing outside `secp256k1/` cites a `file:line` today, or
a "line NNN" in prose. `secp256k1_context_preallocated_size` and
`secp256k1_declassify` name the same two places and survive the bump; the
CHANGELOG entry can keep the numbers, being dated by the release it sits
under.

**The relocated ASM note now abuts `if cross_compile:` with no blank
line between them**, so by the same convention that made it read as
annotating `target_architecture_options()` it now reads as a preamble to
the cross-compile branch — which its own last sentences mention. One
blank line before `if cross_compile:` ends it; the "not in that list"
opening already does the work upwards.

fametrano:
Pushed at f19c3f2. Both suggestions taken here rather than deferred —
they are two lines, and the second is a collision I introduced.

**Line numbers in a vendored file.** Gone from the comment, which now
names `secp256k1_context_preallocated_size` and `secp256k1_declassify`
and says why: a vendored file's lines move with the next submodule bump
and nothing here would notice. It was also the only `file:line` citation
outside `secp256k1/` in the tree, which I checked. CHANGELOG.md keeps
the numbers, being dated by the release it sits under.

**The blank line.** In. The note opens with "not in that list" upwards
and now ends before `if cross_compile:` rather than reading as its
preamble.

Thank you for the enumeration in the re-review — "and none afterwards"
holding for the package and not only for that function is a stronger
claim than I had checked, and `secp256k1_context_create` appearing once
in the whole of `btclib_secp256k1` is what makes it true.

PR #215, opened 2026-08-16
Work out at import what a wrapper was working out per call

### LINKED ISSUE #212: What a wrapper pays per call, and could pay once
## What is asked

Five places where a wrapper computes per call something that does not
change between calls, or allocates something nothing reads. None of them
shows against a signature; all of them are a measurable share of the
short calls — the parses, the serializations and the scalar operations
that a caller doing arithmetic makes in a loop.

Apple M5 arm64, macOS 26.6, CPython 3.14.6, a local static build of
0.8.0.3. Every row is the shipped spelling and the proposed one
alternated in one process, round after round, minimum kept for each,
against a noise row that runs the shipped spelling twice under two names.

| | shipped | proposed | delta | noise |
| --- | --- | --- | --- | --- |
| `xonly.serialize` | 0.257 µs | 0.181 µs | −29.7% | 0.3% |
| `silentpayments.serialize_label` | 0.259 µs | 0.184 µs | −28.8% | 0.1% |
| `_secret.take` | 0.171 µs | 0.121 µs | −29.1% | 0.1% |
| — in context, `keys.prvkey_negate` | 0.366 µs | 0.289 µs | −21.1% | 0.2% |
| `xonly.parse`, 33 or 65 bytes | 0.486 µs | 0.403 µs | −17.1% | 0.0% |
| — in context, `xonly.from_pubkey(65)` | 0.777 µs | 0.653 µs | −16.0% | 0.5% |
| `keys.serialize` | 0.295 µs | 0.266 µs | −9.7% | 0.1% |
| `_scalar.scalar(int)` | 0.157 µs | 0.141 µs | −10.1% | 0.7% |

## The interpolated cdecl, which is the largest of them

[`xonly.serialize`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/btclib_secp256k1/xonly.py#L585)
and
[`silentpayments.serialize_label`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/btclib_secp256k1/silentpayments.py#L808)
build their buffer with `ffi.new(f"char[{_XONLY_SIZE}]")`. The f-string
is there for a good reason — it states the width once, where
`keys.serialize` has to write `33` and `65` into literals and read the
number back off the buffer — and it costs **twice** what a literal does:

| `ffi.new` of a `char[32]` | |
| --- | --- |
| f-string cdecl | 0.1091 µs |
| literal cdecl | 0.0571 µs |
| a hoisted `ffi.typeof` | 0.0576 µs |
| noise: the f-string again | 0.1095 µs |

The third row is the answer: cffi already caches the parse of a cdecl
string, which is why the literal and the hoisted type are one number, and
what the f-string pays is formatting a fresh string and having cffi hash
it. Evaluating it once at import with `ffi.typeof` keeps the width stated
once *and* is as cheap as the literal.

`ffi.sizeof(output)` per call is the same shape of cost, 0.016 µs, and
resolves in the same place: the size comes off the same hoisted type, so
buffer and length still cannot say different numbers.

## The buffer built twice

[`_secret.take`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/btclib_secp256k1/_secret.py#L175-L193)
builds an `ffi.buffer`, reads the secret out of it, and then calls `wipe`,
which builds a second one over the same cdata. An `ffi.buffer` is 0.042 µs.

The invariant is the point here and has to survive: the pair is what
`take` is, and the `finally` that zeroes even when `into` is refused is
load-bearing. Extracting the zeroing into a helper that `wipe` calls too
keeps the statement that writes the zeros written once.

## The argument nobody reads

[`_drop_y`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/btclib_secp256k1/xonly.py#L45-L74)
allocates an `int *` for the parity and returns it in a tuple. Three of
its five callers throw it away at once — `parse`, `_parsed` and
`_tweak_add_`, each spelling `_drop_y(pubkey)[0]`. libsecp256k1
documents `pk_parity` of `secp256k1_xonly_pubkey_from_pubkey` as "can be
NULL", so this is an argument and not a rewrite.

## The two small ones

`keys.serialize` branches twice on `compressed`, once for the cdecl and
once for the flag, and calls `ffi.sizeof` on the result. `_scalar.scalar`
asks `isinstance(num, int) and not isinstance(num, bool)` where
`type(num) is int` is one check and excludes `bool` on its own — the
same fast-path shape
[`octets`](https://github.com/btclib-org/btclib-secp256k1/blob/75cbb44b1ba853585c19c0b30d5126f056f26305/btclib_secp256k1/_scalar.py#L62)
already has above it, and its comment already explains.

## Measured and left alone

Recorded so that nobody spends a session on them:

| | measured | why not |
| --- | --- | --- |
| hoisting `ffi.typeof` for *literal* cdecls | 0.0811 → 0.0800 µs, noise 0.0798 | cffi already caches it; only the f-string pays |
| a module alias for `lib.secp256k1_…` | 0.0160 → 0.0106 µs | 5 ns against ten names that stop saying what they call |
| one `ffi.callback` per scan → `ffi.new_handle` in the unused `label_context` | 0.144 µs to build one | against 118 µs for a ten-output scan. A tidier shape, not a faster one |
| `_cdata.array`'s `list(items)` when it is already a list | 0.136 → 0.108 µs | 0.028 µs behind a type check, on a helper called once per aggregate |
| one block for the ten `found_output` structs a scan allocates | 1.73 → 1.05 µs | 0.6% of the scan it is inside |
| the GIL held across the C calls | `dsa.verify` scales 3.43× on 4 threads, `ssa.verify` 3.25×, against 3.82× for `hashlib.sha256` as the control | cffi releases it already |

## What none of this is

The leverage in this API is elsewhere and is already built: the private
halves and the uncompressed serialization. Five parsed keys sort in 0.69 µs
against 13.98, combine in 2.29 against 14.30, and `keys.parse` is 0.20 µs
on 65 bytes against 2.31 on 33. Those are one to two orders of magnitude,
and the README already points at them. This issue is what is left over
after them.

fametrano:
Four things a wrapper worked out at every call and now works out once.
No answer changes, no signature a caller uses changes.

## What moved

- **An interpolated cdecl.** `xonly.serialize` and
  `silentpayments.serialize_label` declared their buffer with
  `ffi.new(f"char[{_XONLY_SIZE}]")` — which states the width once, the
  reason it is an f-string, and costs twice what a literal does, cffi
  having to hash a string built afresh each time. `ffi.typeof` of the
  same f-string, evaluated at import, is the literal's price with the
  width still stated once.
- **`ffi.sizeof` of a buffer whose size is a constant.** Each hoisted
  type has its size taken from that very type beside it, so buffer and
  length still cannot say different numbers. `keys.serialize` decides its
  flag in the branch that picks its buffer, instead of asking
  `compressed` twice. The length *object* stays per call, for the
  thread-safety reason its comment gives; only its type is hoisted.
- **The second view of memory already in hand.** `_secret.take` built an
  `ffi.buffer` to read the secret and then called `wipe`, which built
  another over the same cdata. `_zero` is now the statement that writes
  the zeros, and `wipe` calls it too, so it is written once.
- **The parity nobody reads.**
  `secp256k1_xonly_pubkey_from_pubkey` documents `pk_parity` as "can be
  NULL", and three of the five callers of `xonly._drop_y` discard it:
  `parse`, `_parsed`, `_tweak_add_`. It takes the pointer as an argument
  now, defaulting to NULL, and the two callers that read a parity
  allocate one.

## The figures, re-measured against the branch as implemented

Each `main` spelling written out and alternated against this one in a
single process over one build, minimum of 9 rounds, with a noise row
running `main`'s spelling twice under two names — an Apple M5, macOS
26.6, arm64, CPython 3.14.6. Every pair is asserted to answer alike
before it is timed.

| | main | now | delta | noise |
| --- | --- | --- | --- | --- |
| `xonly.serialize` | 0.2499 µs | 0.1858 µs | −25.6% | 0.2% |
| `silentpayments.serialize_label` | 0.2583 µs | 0.1912 µs | −26.0% | 0.5% |
| `xonly.parse`, 65 bytes | 0.5071 µs | 0.4168 µs | −17.8% | 0.2% |
| `_secret.take` | 0.1561 µs | 0.1366 µs | −12.5% | 0.2% |
| `keys.serialize`, compressed | 0.3055 µs | 0.2809 µs | −8.1% | 0.1% |
| `keys.serialize`, uncompressed | 0.3057 µs | 0.2785 µs | −8.9% | 0.1% |
| `xonly.from_pubkey`, 65 bytes | 0.7571 µs | 0.6976 µs | −7.9% | 0.3% |
| `keys.prvkey_negate` | 0.3418 µs | 0.3211 µs | −6.1% | 0.2% |

**Two of #212's numbers were wrong and are corrected here.** `take` is
−12.5%, not −29%: the figure in the issue came from a variant with the
zeroing inlined, and `_zero` costs the frame that variant did not pay —
kept anyway, because the alternative is writing the statement that
overwrites a secret in two places. And `prvkey_negate` is −6.1%, not
−21%: that row's `main` variant skipped `_scalar.scalar` altogether, so
the two spellings differed in more than the change being measured. Both
rows above are the honest comparison.

## What was measured and not done

- **Hoisting `ffi.typeof` for *literal* cdecls: nothing.** 0.0800 µs
  against 0.0811, on a noise row that moved 0.0798 — cffi already caches
  the parse of a literal. Which is why this PR hoists only where the
  string is built by an f-string, and leaves
  `ffi.new("secp256k1_pubkey *")` and the rest exactly as they are.
- **`_scalar.scalar`.** `type(num) is int` is one check where
  `isinstance(num, int) and not isinstance(num, bool)` is two, and
  excludes `bool` on its own: 0.141 µs against 0.157. It needs the range
  check and the serialization written twice, once for the exact type and
  once for the subclass, and the int path is the one the module's own
  docstring says is not the common one — a scalar handed to these
  bindings in a loop is bytes, and bytes never reach that test. Declined.
- The rest of #212's "measured and left alone" table stands: the
  `ffi.callback` per scan, the `_cdata.array` list copy, the
  `found_output` block, and the GIL, which cffi releases already.

## Scope

`stubs/_btclib_secp256k1.pyi` gains `typeof`, and `new` takes what it
answers as well as a string — without that mypy fails, which is the stub
doing its job. CHANGELOG.md carries the table and the two declined
results; HISTORY.md carries the four figures a caller sees.

Suite green (582 tests), coverage gate at 100.00%, every hook passes.

Closes #212

## Summary by Sourcery

Hoist several cffi-related computations from hot call paths to module import time to reduce overhead without changing public behavior or results.

Enhancements:
- Optimize x-only key handling by reusing parity buffers only where needed and simplifying `_drop_y` to return just the x-only pubkey.
- Precompute and reuse cffi buffer types and sizes for key and label serialization, avoiding repeated cdecl parsing and sizeof calls.
- Refactor secret wiping to share a single zeroing implementation, removing redundant buffer view creation while preserving safety semantics.

Documentation:
- Extend CHANGELOG and HISTORY with performance notes and rationale for hoisting wrapper work from per-call to import-time.

Chores:
- Update type stubs to expose `ffi.typeof` and accept precomputed cdecl types in `ffi.new`, keeping static type checking accurate with the new usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -190,18 +190,19 @@ release-notes length in the first place, and are still in
   or a `memoryview` needs, and are what the slow path still does; the
   exact type is asked once in front of them. `dsa.verify` passes three
   arguments through it.
-- **`keys.serialize` builds its buffer from a literal cdecl**, 0.313
-  microseconds against 0.356 with the guard already gone and 0.604
-  before it: `ffi.new` of an interpolated `f"char[{size}]"` parses that
+- **`keys.serialize` stopped building its buffer from an interpolated
+  cdecl**, 0.313 microseconds against 0.356 with the guard already gone
+  and 0.604 before it: `ffi.new` of an `f"char[{size}]"` parses that
   string on every call, and `ffi.sizeof` was called twice where the size
-  was in hand. The length buffer is still built per call and deliberately
-  not hoisted, though it holds the same number every time — libsecp256k1
-  writes 0 into it before it does anything and restores it only on
-  success, and it holds what it finds there against 33 or 65 on the way
-  in, so one
-  failed call would leave a shared buffer at zero and every later
-  serialization, on any thread and of a perfectly good key, would be
-  refused. That was measured rather than reasoned about.
+  was in hand. It went to a literal here and then to a hoisted type, for
+  the reason and at the price the entry below states; what follows is
+  what did not change. The length buffer is still built per call and
+  deliberately not hoisted, though it holds the same number every time —
+  libsecp256k1 writes 0 into it before it does anything and restores it
+  only on success, and it holds what it finds there against 33 or 65 on
+  the way in, so one failed call would leave a shared buffer at zero and
+  every later serialization, on any thread and of a perfectly good key,
+  would be refused. That was measured rather than reasoned about.
 - **What the entry points cost now.** Microseconds per call on the
   machine above, the quickest of two passes run in either order so that
   neither tree is the one measured on a warm machine, before against
@@ -308,6 +309,104 @@ release-notes length in the first place, and are still in
   list `__version__` until something reads it, and does afterwards.
   Nothing else changes: the attribute, the `from … import __version__`,
   and reading the metadata directly all answer what they answered.
+
+### What a wrapper works out per call, and now works out once
+
+- **An interpolated cdecl is resolved at import, not at every call**
+  (#212). `xonly.serialize` and `silentpayments.serialize_label` declared
+  their buffer with `ffi.new(f"char[{_XONLY_SIZE}]")`, which states the
+  width once — the reason it is an f-string — and costs twice what a
+  literal does, cffi having to hash a string built afresh each time:
+  0.1219 microseconds against 0.0681. `ffi.typeof` of the same f-string,
+  evaluated once at module level, is the literal's price with the width
+  still stated once.
+- **A literal cdecl is left exactly as it was**, and that is the measured
+  half of the same rule. cffi already caches the parse of one, so
+  hoisting a literal saves 0.003 microseconds — real, 4.3% on a noise row
+  of 0.5%, and an order below the 0.054 an interpolated cdecl gives back.
+  That is the size a hoist has to be worth to justify spelling a width
+  twice, or naming every cdecl in the package at module level, so every
+  `ffi.new` here still spells its cdecl in full — including the two
+  beside the hoisted types.
+- **`ffi.sizeof` of a buffer whose size is a constant is not asked per
+  call**, and that is what asked `keys.serialize` for a type at module
+  level: each size is `ffi.sizeof` of the very type beside it, so the
+  buffer and the length cannot say different numbers, and the 0.0175
+  microseconds is not paid at every call. `keys.serialize` also decides
+  its flag in the branch that picks its buffer, so the one condition is
+  asked once. The length *object* is still built per call, for the
+  thread-safety reason its comment gives.
+- **`_secret.take` zeroes through the view it already holds.** It built
+  an `ffi.buffer` to read the secret and then called `wipe`, which built
+  a second one over the same cdata. The statement that writes the zeros
+  is now `_zero`, which `wipe` calls too, so it is written once — and
+  that frame is paid for deliberately: `wipe` twice is 0.1826
+  microseconds, `_zero` is 0.1439 and the same statement inlined in both
+  places is 0.1314. A quarter of what was on the table, for one copy of
+  the line that overwrites a secret rather than two.
+- **The parity nobody reads is not allocated.**
+  `secp256k1_xonly_pubkey_from_pubkey` documents `pk_parity` as "can be
+  NULL", and the callers that want the object alone — `parse`, `_parsed`
+  and `_tweak_add_` — are why NULL is the default. `_drop_y` takes the
+  pointer as an argument now: 0.1747 microseconds against 0.2425 on the
+  conversion alone. `_from_pubkey_` and `to_pubkey` read a parity and go
+  through `_drop_y_with_parity`, which is that allocation written once,
+  and no saving lands on them — what it adds is some hundredths of a
+  microsecond, measurable against both. One figure rather than one per
+  caller: it is the same frame at each, and two harnesses put a different
+  one of the two ahead, which is the spread of a number this small and
+  not a structure. What they get out of this is the C call still written
+  once and the call site they had.
+- **Two sessions, and each is named where a figure comes from.** Every
+  wrapper figure this change quotes — in CHANGELOG.md, HISTORY.md and in
+  the comments of `keys.py`, `xonly.py`, `silentpayments.py` and
+  `_secret.py` alike — is the one below; every cffi-primitive figure is
+  the one after it. An Apple M5, macOS 26.6, arm64, CPython 3.14.6.
+
+  `main`'s spelling of each call written out and alternated against this
+  one in a single process over one build, minimum of 15 rounds of
+  200 000 calls, a noise row running `main`'s spelling twice, and every
+  pair asserted to answer alike before it is timed:
+
+  | | main | now |
+  | --- | --- | --- |
+  | `_drop_y`, the conversion alone | 0.2425 | 0.1747 |
+  | `xonly.serialize` | 0.2569 | 0.1969 |
+  | `silentpayments.serialize_label` | 0.2621 | 0.2023 |
+  | `_secret.take` | 0.1826 | 0.1439 |
+  | `xonly.parse`, 65 bytes | 0.5021 | 0.4295 |
+  | `xonly._from_pubkey_`, 65 bytes | 0.5127 | 0.4711 |
+  | `keys.prvkey_negate` | 0.3790 | 0.3380 |
+  | `keys.serialize`, compressed | 0.3106 | 0.2906 |
+  | `keys.serialize`, uncompressed | 0.3101 | 0.2903 |
+  | *noise* | 0.2578 | 0.2577 |
+
+  and the primitives, minimum of 15 rounds of 300 000 calls:
+
+  | | first | second |
+  | --- | --- | --- |
+  | `ffi.new` of an f-string, then of a literal `char[32]` | 0.1219 | 0.0681 |
+  | `ffi.new` of a literal, then of a hoisted `char[33]` | 0.0685 | 0.0656 |
+  | `ffi.new` of a literal, then of a hoisted `size_t *` | 0.0809 | 0.0774 |
+  | `ffi.unpack` with `ffi.sizeof`, then with the constant | 0.0764 | 0.0589 |
+  | *noise* | 0.0676 | 0.0680 |
+
+- **`_scalar.scalar` was measured and left alone**, which is the one
+  declined here. Asking `type(num) is int` instead of
+  `isinstance(num, int) and not isinstance(num, bool)` is one check where
+  those are two and excludes `bool` on its own, and it is 0.141
+  microseconds against 0.157 — but it needs the range check and the
+  serialization written twice, once for the exact type and once for the
+  subclass, and the int path is the one this module's own docstring says
+  is not the common one: a scalar handed to these bindings in a loop is
+  bytes, and that path does not reach the test at all.
+- **The stub gains an opaque `_CType`**, which is what `typeof` answers
+  and what `new` takes beside a string. Typing that argument `Any`
+  instead would be the vacuous type-checking
+  `stubs/_btclib_secp256k1.pyi` exists to prevent: `ffi.new(_XONLY_SIZE)`
+  — an int where a cdecl belongs, and the size and the type built from
+  it now sit a dozen lines apart in the same module — passes `--strict`
+  with `Any` and is an `[arg-type]` error with `_CType`.
 
 ### Signing under one key
 
--- a/btclib_secp256k1/_secret.py
+++ b/btclib_secp256k1/_secret.py
@@ -41,6 +41,37 @@ from ._scalar import scalar
 from .context import ctx
 
 
+def _zero(memory: memoryview) -> None:
+    """Overwrite what a cffi buffer views.
+
+    The statement that writes the zeros, written once: `wipe` builds the
+    view over a cdata, and `take` is already holding one it read the
+    secret through, so calling `wipe` from there would build a second
+    over the same memory.
+
+    That it is a function and not the statement inlined in both places
+    is a trade made deliberately and against the faster spelling.
+    Measured over the same session as every other figure of this change,
+    which CHANGELOG.md names: calling `wipe` a second time is 0.1826
+    microseconds, this is 0.1439, and the inlined statement is 0.1314.
+    So the frame gives up about a quarter of what is on the table, and
+    what it buys is that the line which overwrites a secret exists once:
+    two copies are two things to keep true, in the one facility whose
+    whole claim is about what is left in memory.
+
+    Args:
+        memory: the `ffi.buffer` to overwrite, and a view of the whole
+            cdata rather than a slice of one -- `wipe` builds it from
+            the cdata for the reason its own docstring gives, and
+            `_zero(ffi.buffer(buffer, 8))` would zero a quarter of the
+            private key inside a keypair and report success, which is
+            the failure that docstring records. Typed as the stub
+            declares `ffi.buffer` to answer, this being the module that
+            reads both a slice and a length off one.
+    """
+    memory[:] = bytes(len(memory))
+
+
 def wipe(buffer: CData) -> None:
     """Overwrite a buffer that held a secret.
 
@@ -54,8 +85,7 @@ def wipe(buffer: CData) -> None:
     Args:
         buffer: the cffi buffer to overwrite, as `ffi.new` returned it.
     """
-    memory = ffi.buffer(buffer)
-    memory[:] = bytes(len(memory))
+    _zero(ffi.buffer(buffer))
 
 
 def into_buffer(into: object, size: int) -> memoryview:
@@ -172,12 +202,15 @@ def take(buffer: CData, *, into: MutableBytesLike | None = None) -> bytes | None
             one-dimensional octets.
         ValueError: if `into` is not the secret's length.
     """
+    # the zeroing goes through this view rather than through `wipe`,
+    # which would build a second one over the same cdata: what `wipe`
+    # writes is `_zero`, and it is that statement both paths reach
     memory = ffi.buffer(buffer)
     if into is None:
         secret = bytes(memory)
-        wipe(buffer)
+        _zero(memory)
         return secret
-    # the wipe is in a `finally` because `into_buffer` can refuse, and
+    # the zeroing is in a `finally` because `into_buffer` can refuse, and
     # this is the one call here with a failure path: without it a
     # rejected buffer returns through the entry point leaving a live
     # private key in cffi memory that is freed without being
@@ -189,7 +222,7 @@ def take(buffer: CData, *, into: MutableBytesLike | None = None) -> bytes | None
         view = into_buffer(into, len(memory))
         view[:] = memory
     finally:
-        wipe(buffer)
+        _zero(memory)
     return None
 
 
--- a/btclib_secp256k1/keys.py
+++ b/btclib_secp256k1/keys.py
@@ -30,6 +30,25 @@ from .context import ctx
 COMPRESSED = 258
 UNCOMPRESSED = 2
 
+# the two buffers `serialize` writes into, and the lengths it declares
+# them with. The pairing is what these are for: each size is
+# `ffi.sizeof` of the very type beside it, so the buffer and the length
+# cannot say different numbers, and stating the width once is what asked
+# for a type at module level in the first place. What it saves is the
+# `ffi.sizeof` per call, 0.0175 microseconds.
+#
+# It is not hoisted for the `ffi.new`. cffi already caches the parse of
+# a literal cdecl, so a hoisted type beats one by 0.003 microseconds --
+# real, 4.3% on a noise row of 0.5%, and an order below the 0.054 an
+# interpolated cdecl gives back, which is the figure a hoist has to earn
+# to be worth spelling a width twice or naming every cdecl in the
+# package. So the length pointer below stays spelled in full, like every
+# other `ffi.new` here. `xonly.py` carries the session
+_COMPRESSED_BUFFER_TYPE = ffi.typeof("char[33]")
+_COMPRESSED_SIZE = ffi.sizeof(_COMPRESSED_BUFFER_TYPE)
+_UNCOMPRESSED_BUFFER_TYPE = ffi.typeof("char[65]")
+_UNCOMPRESSED_SIZE = ffi.sizeof(_UNCOMPRESSED_BUFFER_TYPE)
+
 
 def prvkey_verify(prvkey: BytesLike | int) -> bool:
     """Return True if the private key is a valid scalar, i.e. in [1, n-1].
@@ -1054,11 +1073,11 @@ def serialize(pubkey: CData, compressed: bool = True) -> bytes:
         >>> keys.serialize(keys.parse(uncompressed)).hex()[:10]
         '0279be667e'
     """
-    # the buffer is declared by a literal and the size read back off it,
-    # so the two cannot say different numbers: `ffi.new` of an
-    # interpolated cdecl parses that string on every call, which is 0.043
-    # microseconds of the 0.313 this costs -- an Apple M5, macOS 26.6,
-    # arm64, CPython 3.13.14, minimum of 9 rounds of 500 000 calls.
+    # the buffer's type and the size are declared together at the top of
+    # this module, each size being `ffi.sizeof` of the type beside it, so
+    # the two cannot say different numbers and neither is worked out
+    # again here. The flag is decided in the branch that picks the
+    # buffer, so the one condition is asked once.
     #
     # What is unpacked is the buffer, not the length libsecp256k1 reports
     # back: this serialization has one length per flag, so a buffer of
@@ -1068,17 +1087,26 @@ def serialize(pubkey: CData, compressed: bool = True) -> bytes:
     # that spelling, and dies in this one). The DER serialization is the
     # other case, and reads `length[0]` for the reason given there.
     #
-    # The length is built per call and not hoisted, though it is the same
-    # number every time: libsecp256k1 writes 0 into it before it does
-    # anything (`*outputlen = 0` in secp256k1.c) and restores it only on
-    # success, and it holds the value it finds there against 33 or 65 on
-    # the way in. One failed call would leave a shared buffer at
-    # zero, and every later serialization -- on any thread, of a
-    # perfectly good key -- would be refused
-    output = ffi.new("char[33]" if compressed else "char[65]")
-    size = ffi.sizeof(output)
+    # The length *object* is built per call and not hoisted, though it
+    # holds the same number every time: libsecp256k1 writes 0 into it
+    # before it does anything (`*outputlen = 0` in secp256k1.c) and
+    # restores it only on success, and it holds the value it finds there
+    # against 33 or 65 on the way in. One failed call would leave a
+    # shared buffer at zero, and every later serialization -- on any
+    # thread, of a perfectly good key -- would be refused
+    if compressed:
+        output, size, flags = (
+            ffi.new(_COMPRESSED_BUFFER_TYPE),
+            _COMPRESSED_SIZE,
+            COMPRESSED,
+        )
+    else:
+        output, size, flags = (
+            ffi.new(_UNCOMPRESSED_BUFFER_TYPE),
+            _UNCOMPRESSED_SIZE,
+            UNCOMPRESSED,
+        )
     length = ffi.new("size_t *", size)
-    flags = COMPRESSED if compressed else UNCOMPRESSED
     serialized = lib.secp256k1_ec_pubkey_serialize(ctx, output, length, pubkey, flags)
     if not serialized:
         raise RuntimeError("point serialization failed")
--- a/btclib_secp256k1/silentpayments.py
+++ b/btclib_secp256k1/silentpayments.py
@@ -43,6 +43,16 @@ from .context import ctx
 SUMMARY_SIZE = ffi.sizeof("secp256k1_silentpayments_prevouts_summary")
 LABEL_SIZE = 33
 
+# the buffer `serialize_label` writes into, resolved here rather than at
+# every call: `ffi.new` of an f-string formats it and leaves cffi to
+# hash the result, where `ffi.typeof` of the same string, evaluated
+# once, costs what a literal cdecl does and still states the width once.
+# 0.2023 microseconds against 0.2621 on this serialization -- one
+# session with every other figure of this change, which CHANGELOG.md
+# names. `xonly.py` says the rest, the reason a literal cdecl is left
+# alone included
+_LABEL_BUFFER_TYPE = ffi.typeof(f"char[{LABEL_SIZE}]")
+
 
 def _create_outputs_(
     recipients: Sequence[tuple[CData, CData]],
@@ -805,10 +815,12 @@ def serialize_label(label_obj: CData) -> bytes:
             which a label it produced cannot make it do.
             `context.check` is what tells the two apart.
     """
-    output = ffi.new(f"char[{LABEL_SIZE}]")
+    output = ffi.new(_LABEL_BUFFER_TYPE)
     serialized = lib.secp256k1_silentpayments_recipient_label_serialize(
         ctx, output, label_obj
     )
     if not serialized:
         raise RuntimeError("label serialization failed")
-    return ffi.unpack(output, ffi.sizeof(output))
+    # the length is the constant the buffer's type was built from, so the
+    # two still cannot say different numbers
+    return ffi.unpack(output, LABEL_SIZE)
--- a/btclib_secp256k1/xonly.py
+++ b/btclib_secp256k1/xonly.py
@@ -41,13 +41,75 @@ from .context import ctx
 # two lengths this module takes are a full public key, whose x it is
 _XONLY_SIZE = 32
 
+# the buffer `serialize` writes into, resolved here rather than at every
+# call. `ffi.new` of an f-string formats it and leaves cffi to hash the
+# result: 0.1219 microseconds against 0.0681 for a literal cdecl -- an
+# Apple M5, macOS 26.6, arm64, CPython 3.14.6, minimum of 15 rounds of
+# 300 000 calls, on a noise row that moved 0.5%. A literal here would
+# state the width a second time; `ffi.typeof` of the same f-string,
+# evaluated once, is the literal's price with the width still stated
+# once.
+#
+# A *literal* cdecl is already cached by cffi, and hoisting one saves
+# 0.003 -- 0.0656 against 0.0685, real at 4.3% and an order below the
+# 0.054 above. Too small to be worth a module-level name for every
+# cdecl in the package, so this is done where the string is built and
+# every other `ffi.new` here is spelled in full
+_XONLY_BUFFER_TYPE = ffi.typeof(f"char[{_XONLY_SIZE}]")
 
-def _drop_y(pubkey: CData) -> tuple[CData, int]:
-    """Convert a parsed public key into a parsed x-only key and the parity.
+
+def _drop_y(pubkey: CData, parity: CData = ffi.NULL) -> CData:
+    """Convert a parsed public key into a parsed x-only key.
 
     The conversion itself, which is where the y is dropped, without the
     serialization that follows it in `_from_pubkey_`: `_tweak_add_` wants
     the object and not the 32 bytes, having a tweak to add to it.
+
+    Args:
+        pubkey: the already-parsed public key, as `keys.parse` returns.
+        parity: an `int *` to receive the parity of the y being dropped,
+            0 for even and 1 for odd -- or the NULL that says nobody is
+            reading it, which is the default because the callers that
+            want the object alone are why it exists: `parse`, `_parsed`
+            and `_tweak_add_`. libsecp256k1 documents `pk_parity` as
+            "can be NULL", so the allocation is the caller's to make
+            rather than this function's, and skipping it is 0.4295
+            microseconds against 0.5021 on `xonly.parse` of a 65-byte
+            key -- CHANGELOG.md names the session every figure in this
+            module comes from. `_drop_y_with_parity` is the allocation,
+            written once, for the callers that read one.
+
+    Returns:
+        The libsecp256k1 x-only public key object.
+
+    Raises:
+        RuntimeError: if libsecp256k1 refuses the object -- one it
+            cannot read -- or fails to convert for any other reason,
+            which a key it produced cannot make it do. `context.check`
+            is what tells the two apart.
+    """
+    xonly_pubkey = ffi.new("secp256k1_xonly_pubkey *")
+    converted = lib.secp256k1_xonly_pubkey_from_pubkey(
+        ctx, xonly_pubkey, parity, pubkey
+    )
+    if not converted:
+        raise RuntimeError("x-only public key conversion failed")
+    return xonly_pubkey
+
+
+def _drop_y_with_parity(pubkey: CData) -> tuple[CData, int]:
+    """Convert a parsed public key, and read the parity of the y dropped.
+
+    `_drop_y` for the callers that want the parity, so that the
+    allocation is written once rather than at each of their call sites.
+    The NULL default above is where the saving is -- 0.1747 microseconds
+    against 0.2425 on the conversion alone -- and it lands on the callers
+    that discard the parity, not on these: no saving lands here, and what
+    this adds is some hundredths of a microsecond, measurable against
+    both. One figure and not one per caller: the same frame is added at
+    each, and two harnesses put a different one of the two ahead, which
+    is the spread and not a structure. What it buys is the C call written
+    once and these two call sites left as they read.
 
     Args:
         pubkey: the already-parsed public key, as `keys.parse` returns.
@@ -57,21 +119,11 @@ def _drop_y(pubkey: CData) -> tuple[CData, int]:
         y that was dropped: 0 for even, 1 for odd.
 
     Raises:
-        RuntimeError: if libsecp256k1 refuses the object -- one it
-            cannot read -- or fails to convert for any other reason,
-            which a key it produced cannot make it do. `context.check`
-            is what tells the two apart.
-        RuntimeError: if libsecp256k1 fails for any other reason, which
-            no valid key can make it do.
+        RuntimeError: if libsecp256k1 refuses the object or fails to
+            convert it, as `_drop_y` says.
     """
-    xonly_pubkey = ffi.new("secp256k1_xonly_pubkey *")
     parity = ffi.new("int *")
-    converted = lib.secp256k1_xonly_pubkey_from_pubkey(
-        ctx, xonly_pubkey, parity, pubkey
-    )
-    if not converted:
-        raise RuntimeError("x-only public key conversion failed")
-    return xonly_pubkey, parity[0]
+    return _drop_y(pubkey, parity), parity[0]
 
 
 def _from_pubkey_(pubkey: CData) -> tuple[bytes, int]:
@@ -92,7 +144,7 @@ def _from_pubkey_(pubkey: CData) -> tuple[bytes, int]:
         RuntimeError: if libsecp256k1 fails to convert or serialize it,
             which no valid key can make it do.
     """
-    xonly_pubkey, parity = _drop_y(pubkey)
+    xonly_pubkey, parity = _drop_y_with_parity(pubkey)
     return serialize(xonly_pubkey), parity
 
 
@@ -258,7 +310,7 @@ def _tweak_add_(pubkey: CData, tweak: BytesLike | int) -> tuple[bytes, int]:
         RuntimeError: if libsecp256k1 fails to convert or serialize the
             result, which no valid input can make it do.
     """
-    return _tweak_xonly(_drop_y(pubkey)[0], tweak)
+    return _tweak_xonly(_drop_y(pubkey), tweak)
 
 
 def tweak_add(pubkey_bytes: BytesLike, tweak: BytesLike | int) -> tuple[bytes, int]:
@@ -467,7 +519,7 @@ def parse(pubkey_bytes: BytesLike, name: str = "public key") -> CData:
         pubkey = keys._parsed(pubkey_bytes, name)
         if pubkey is None:
             raise ValueError(f"invalid {name}")
-        return _drop_y(pubkey)[0]
+        return _drop_y(pubkey)
 
     xonly_pubkey = ffi.new("secp256k1_xonly_pubkey *")
     if not lib.secp256k1_xonly_pubkey_parse(ctx, xonly_pubkey, pubkey_bytes):
@@ -499,7 +551,7 @@ def _parsed(pubkey_bytes: BytesLike, name: str) -> CData | None:
     pubkey_bytes = octets(pubkey_bytes, name)
     if len(pubkey_bytes) != _XONLY_SIZE:
         pubkey = keys._parsed(pubkey_bytes, name)
-        return None if pubkey is None else _drop_y(pubkey)[0]
+        return None if pubkey is None else _drop_y(pubkey)
 
     xonly_pubkey = ffi.new("secp256k1_xonly_pubkey *")
     if not lib.secp256k1_xonly_pubkey_parse(ctx, xonly_pubkey, pubkey_bytes):
@@ -583,9 +635,11 @@ def to_pubkey(pubkey_bytes: BytesLike, compressed: bool = True) -> bytes:
 
     # the point is already lifted, so the y is read rather than found:
     # an odd one is negated in the field, where reaching the even-y point
-    # through the 32 octets would be a square root of an x in hand
+    # through the 32 octets would be a square root of an x in hand. The
+    # x-only key the conversion builds is what is thrown away here, the
+    # parity being the whole of what this asks for
     pubkey = keys.parse(pubkey_bytes)
-    if _drop_y(pubkey)[1]:
+    if _drop_y_with_parity(pubkey)[1]:
         keys._pubkey_negate_(pubkey)
     return keys.serialize(pubkey, compressed)
 
@@ -613,8 +667,12 @@ def serialize(xonly_pubkey: CData) -> bytes:
         RuntimeError: if libsecp256k1 fails for any other reason, which
             a key it produced cannot make it do.
     """
-    output = ffi.new(f"char[{_XONLY_SIZE}]")
+    output = ffi.new(_XONLY_BUFFER_TYPE)
     serialized = lib.secp256k1_xonly_pubkey_serialize(ctx, output, xonly_pubkey)
     if not serialized:
         raise RuntimeError("x-only public key serialization failed")
-    return ffi.unpack(output, ffi.sizeof(output))
+    # the length is the constant the buffer's type was built from, so
+    # the two cannot say different numbers, and `ffi.sizeof` of the
+    # cdata would be 0.0175 microseconds nothing reads, measured as the
+    # comment above says
+    return ffi.unpack(output, _XONLY_SIZE)
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -310,6 +310,21 @@ it, and does afterwards. Reading it — as the attribute, as
 metadata — answers what it always answered, and costs what it always
 cost, the first read and every one after it.
 
+**Serializing and parsing cost a little less, and nothing changed about
+what they answer.** Four things a wrapper used to work out at every call
+are worked out once: the cffi type of a buffer whose declaration was
+built by an f-string, the size of a buffer whose size is a constant, the
+second view of memory `_secret.take` was already looking at, and the
+parity `xonly._drop_y` allocated for `parse`, `_parsed` and
+`_tweak_add_`, which throw it away.
+What that is worth, against the same calls as they were and with a noise
+row beside each: `xonly.serialize` 0.197 microseconds against 0.257,
+`silentpayments.serialize_label` 0.202 against 0.262, `xonly.parse` of a
+65-byte key 0.430 against 0.502, `keys.serialize` 0.291 against 0.311.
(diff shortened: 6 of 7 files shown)


fametrano (commented):
## Review

**Changes requested.** The review state is `COMMENTED` rather than
`CHANGES_REQUESTED` only because GitHub refuses either that or an approval
from the pull request's own author; read the verdict from here, not from the
state.

Nothing wrong with what the code computes, and a short
list of things to correct before it lands. No correctness defect found: the
four hoists are sound, the parity default is what libsecp256k1 documents,
and `_zero` writes the same zeros through the same memory.

### What I checked, and how

Worktree on `af84efe` (a descendant of `origin/main`, so the branch is
current), submodule at `6e2c8bc`, on the same machine the branch measured
on — Apple M5, macOS 26.6, arm64, CPython 3.14.6.

- `pytest -q` — 582 passed
- `pytest --cov` — 100.00%, 777 statements, 164 branches, nothing missed
- `sphinx-build -W --keep-going` — build succeeded
- `mypy btclib_secp256k1 tests` — clean, and clean again with the tighter
  stub proposed in 1 below
- every figure below is the branch's own function with only the changed
  statement put back, so the two spellings differ in nothing else,
  alternated in one process over one build, minimum of 11 rounds, each row
  against a noise row running the same spelling twice, and every pair
  asserted to answer alike before it is timed

Where I could re-derive a figure, it held. `xonly.serialize` 0.2582 → 0.1925,
−25.5% against a noise row that moved −0.9%. `xonly.parse` of a 65-byte key
0.4938 → 0.4164, −15.7% against +0.1%. `keys.serialize` compressed 0.3039 →
0.2788, −8.2% against −0.9%. `ffi.sizeof` of a cdata is 0.0135 microseconds
net of the calling frame, which is the 0.016 the comment claims.

### Required

**1. The stub gives up exactly what the stub is for.**
[`stubs/_btclib_secp256k1.pyi:41`](https://github.com/btclib-org/btclib-secp256k1/blob/af84efe6948585932086bb915c2f3c99837f0bf3/stubs/_btclib_secp256k1.pyi#L41).
Its own docstring says that without it "every wrapper in the package
type-checks vacuously", and `new(self, cdecl: Any, ...)` is that condition
restored for the one call the whole package is built out of. It is not
hypothetical: `_XONLY_SIZE` and `_XONLY_BUFFER` now sit a dozen lines apart
in the same module, and with `Any` the wrong one passes. Checked both ways:

```python
output = ffi.new(_XONLY_SIZE)   # an int where a cdecl belongs
```

```
# with cdecl: Any
Success: no issues found in 1 source file
# with the two lines below
error: Argument 1 to "new" of "_FFI" has incompatible type "int";
       expected "str | _CType"  [arg-type]
```

A named opaque type says what `typeof` answers and keeps `new` honest, and
`mypy btclib_secp256k1 tests` passes with it and no other edit:

```python
class _CType: ...

class _FFI:
    def new(self, cdecl: str | _CType, init: Any = ...) -> Any: ...
    def typeof(self, cdecl: str) -> _CType: ...
```

The module docstring's last paragraph then says the pairing rather than
apologising for `Any`.

**2. v0.8.0.3 now says two things about how `keys.serialize` builds its
buffer.**
[`CHANGELOG.md:117`](https://github.com/btclib-org/btclib-secp256k1/blob/af84efe6948585932086bb915c2f3c99837f0bf3/CHANGELOG.md#L117)
is still "**`keys.serialize` builds its buffer from a literal cdecl**, 0.313
microseconds against 0.356 with the guard already gone and 0.604 before it",
in the same unreleased release as the new bullet saying its type is hoisted
and it costs 0.2809. It does not build it from a literal cdecl any more, and
a reader of the release gets two answers to one question and three figures
for one call. One of the two bullets has to absorb the other — "One fact in
one place", and the whole point of the section being unreleased is that the
older bullet is still editable.

**3. History in a comment.**
[`btclib_secp256k1/keys.py:1072`](https://github.com/btclib-org/btclib-secp256k1/blob/af84efe6948585932086bb915c2f3c99837f0bf3/btclib_secp256k1/keys.py#L1072):
"One branch rather than the two the same condition **used to be** asked in".
CONTRIBUTING.md's "No history in the prose" rules exactly this out; the
present-tense form — the flag is decided where the buffer is, so the
condition is asked once — says the same thing and survives the next reader.
The CHANGELOG is where "rather than asking `compressed` twice" belongs, and
it already says it.

**4. Two figures arrive without the command that produced them.**
[`keys.py:38`](https://github.com/btclib-org/btclib-secp256k1/blob/af84efe6948585932086bb915c2f3c99837f0bf3/btclib_secp256k1/keys.py#L38)
states "0.016 microseconds of the 0.295 that serialization costs" and
[`xonly.py:614`](https://github.com/btclib-org/btclib-secp256k1/blob/af84efe6948585932086bb915c2f3c99837f0bf3/btclib_secp256k1/xonly.py#L614)
"0.016 microseconds nothing reads", neither with a machine, an interpreter
or a round count. The comment `keys.py` deletes to make room carried all
three ("an Apple M5, macOS 26.6, arm64, CPython 3.13.14, minimum of 9 rounds
of 500 000 calls"), and so does every other measurement comment in the
package — `_scalar.py:57`, `dsa.py:340`, `dsa.py:441`, `keys.py:948`,
`keys.py:971`, `ssa.py:487` — and so do the three this branch adds:
`xonly.py:44`, `silentpayments.py:51` and the `_drop_y` docstring. "A number
in prose comes from a command, and the command belongs beside it." Both figures
are right — I re-derived the 0.016 — so this is the attribution and not the
number. Also: 0.295 is a third value for what `keys.serialize` costs, beside
the CHANGELOG's 0.3055 and the deleted comment's 0.313; say which spelling
it names.

**5. `_zero`'s docstring keeps the framing the PR body corrects.**
[`_secret.py:47-50`](https://github.com/btclib-org/btclib-secp256k1/blob/af84efe6948585932086bb915c2f3c99837f0bf3/btclib_secp256k1/_secret.py#L47-L50):
"an `ffi.buffer` is 0.042 microseconds, which is a quarter of what `take`
costs". The description says plainly that this is not what the change
delivers — −12.5%, not −29%, because `_zero` costs the frame the inlined
variant did not pay — and the docstring is the site the next reader
re-derives from. I measure the same trade: against a `take` that allocates
its buffer per call, `main` → `_zero` is −17.8% and `main` → the same
statement written inline is −25.7%, on a noise row of +0.4%. So the frame
gives up about a third of what was on the table, deliberately, and the
docstring is where that trade is made and should say so. As it stands the
comment argues for the version that was not taken.

### Suggested

**6. `_zero(memory: Any)` → `memory: memoryview`.**
[`_secret.py:44`](https://github.com/btclib-org/btclib-secp256k1/blob/af84efe6948585932086bb915c2f3c99837f0bf3/btclib_secp256k1/_secret.py#L44).
The stub declares `ffi.buffer` to return a `memoryview` for the stated
reason that `_secret` is what reads both the slice and the length off it, so
`Any` here is the one unchecked annotation in the module whose whole subject
is not leaving a secret behind. `mypy btclib_secp256k1 tests` and
`ruff check` both pass with `memoryview`, and the `Any` import goes with it.

**7. The rule the CHANGELOG states and the code that ships disagree about
literal cdecls.** The bullet says hoisting a literal buys nothing, "which is
why this is done where the string is built and nowhere else" and
`ffi.new("secp256k1_pubkey *")` and the rest are left alone. Four literal
cdecls are hoisted all the same: `char[33]`, `char[65]`, `size_t *`
(`keys.py:43`) and `int *` (`xonly.py:54`) — and two lines under `_PARITY`,
`_drop_y` still writes `ffi.new("secp256k1_xonly_pubkey *")` in full, so the
module now spells the same thing two ways.

Measured here, the rule is also not quite right, though the magnitudes are
small: hoisting a literal is above the noise for the buffers and nothing for
the length.

| | literal | hoisted | delta | noise |
| --- | --- | --- | --- | --- |
| `ffi.new("char[33]")` | 0.0605 | 0.0560 | −7.4% | +0.6% |
| `ffi.new("int *")` | 0.0605 | 0.0578 | −4.4% | +0.6% |
| `ffi.new("size_t *", 33)` | 0.0697 | 0.0694 | −0.4% | +0.6% |
| `ffi.new(f"char[{32}]")` | 0.1133 | 0.0571 | −49.6% | +0.6% |

And within `keys.serialize`, compressed: `main` → the branch −8.2%, `main` →
literal cdecls with only the sizes hoisted −7.0%, the branch → literal
cdecls +3.2%, noise −0.9%. So the two `char[]` types earn about a point of
the eight and the `ffi.sizeof` earns the rest, and `_LENGTH` earns nothing
measurable. Either restate the rule as what it is — the f-string is where
the win is, and a literal is worth hoisting only where the type is wanted at
module level anyway, which is true of the two buffers because the size is
taken off them — or drop `_LENGTH` and `_PARITY` and let those two lines
read like their neighbours.

**8. The out-parameter buys nothing on the two callers it complicates.**
Isolating the `_drop_y` change alone, everything else being the branch's:
with the parity discarded, 0.2327 → 0.1611, −30.8% on a noise row of +0.6% —
the change is worth having, and the three callers that discard it are where
it lands. On the two that read a parity it is at the noise floor:
`to_pubkey` of 65 bytes −0.9% against +1.4% noise, and `_from_pubkey_`
measured −3.4% in one harness and +2.3% in another, which is to say nothing.
Those two now write three lines and an `ffi.new` at the call site for a
saving that is not there. A second entry point — `_drop_y` answering the
object, and one function beside it answering the pair — would keep the
−30.8% where it is and leave `_from_pubkey_` and `to_pubkey` reading as they
did. Your call; the out-parameter is what libsecp256k1's own signature is,
and the docstring cites it, so this is a preference and not a defect.

### Checked and fine

- 100% coverage really does reach both arms of the new default: the NULL
  path through `parse`, `_parsed` and `_tweak_add_`, and the allocated one
  through `_from_pubkey_` and `to_pubkey`.
- "three of the five callers" is right — `_drop_y` has exactly five call
  sites, at `xonly.py` 115, 249, 458, 490 and 579.
- `ffi.typeof` at module level is safe on both linkages, and the static and
  dynamic wheel jobs agree; `silentpayments.SUMMARY_SIZE` already
  established that a module-level `ffi` call is fine at import.
- `take`'s `finally` still zeroes on the refusal paths, and the four
  `test_take_refuses_*` tests still prove it: `_zero(memory)` writes through
  the view `wipe` would have rebuilt over the same cdata, and the length is
  still the memory's own rather than a pointer's.
- `spy_on_wipe` in `tests/test_modules.py` patches `silentpayments.wipe` and
  is unaffected by `take` no longer calling `wipe`; the AST walk in
  `tests/test_secret.py` keys on `take`, not on `wipe`, so its population is
  unchanged.
- CI is green on every job, `pre-commit.ci` included.

Fix 1 through 5 and I will ACK. Issue: https://github.com/btclib-org/btclib-secp256k1/issues/212

fametrano:
Pushed at c9057c5. All five required are in, and all three suggestions
taken — 8 included, which changed the shape of the change.

**1 — the stub.** `_CType` as you spelled it, and the probe now fails the
way it should:

```
$ mypy --strict probe.py     # output = ffi.new(_XONLY_SIZE)
error: Argument 1 to "new" of "_FFI" has incompatible type "int";
       expected "str | _CType"  [arg-type]
```

The module docstring's last paragraph says the pairing rather than
apologising for `Any`.

**2 — two answers to one question.** The older bullet keeps its three
figures, which are the history of that call and are true of the release,
but it now says the buffer *stopped being* an interpolated cdecl and hands
off to the new entry for where it went and at what price. One answer,
reading down.

**3 — history in a comment.** Present tense: *the flag is decided in the
branch that picks the buffer, so the one condition is asked once*.

**4 — figures without a command.** Both now carry the machine, the
interpreter and the round count, and the `0.295` is gone: `keys.serialize`
has one figure in this branch, 0.3130 before and 0.2944 after, from the
table.

**5 — `_zero`'s docstring.** It now makes the trade instead of arguing for
the spelling that was not taken. My own numbers, on a `take` allocating its
buffer per call, minimum of 25 rounds of 200 000 with the noise row at
0.6%: `wipe` twice 0.2523, `_zero` 0.2080, inlined 0.1979. So a fifth of
what was on the table rather than your third — the `_zero` figure matches
yours closely (−17.6% against your −17.8%), the inline one does not
(−21.6% against −25.7%). Either way the conclusion is the same and the
docstring now says which way the trade went and why.

**6 — `memoryview`.** Done, and the `Any` import went with it.

**7 — the rule and the code disagreed.** You were right, and the resolution
is the code moving rather than the rule: `_LENGTH` and `_PARITY` are gone,
so every `ffi.new` in the package spells its cdecl in full again. The two
`char[]` types stay, but the entry now gives the reason that survives —
the *size* is taken off the type at module level, which is what states the
width once, and what that saves is the per-call `ffi.sizeof` and not the
`ffi.new`. Measured here, hoisting a literal is at the noise floor rather
than your one point: `char[33]` 0.0670 against 0.0684, `int *` 0.0683
against 0.0703, `size_t *, 33` 0.0801 against 0.0815, on a noise row of
0.5%. Dropping `_LENGTH` costs `keys.serialize` about two points (−8.1%
became −5.9%), which is the price of the code and the rule saying the same
thing.

**8 — the out-parameter on the two callers it complicated.** Taken.
`_drop_y_with_parity` is the allocation written once; `_from_pubkey_` and
`to_pubkey` read as they did on `main`, and `_drop_y` keeps the NULL
default and the −27.3% where it lands. The C call is still written once,
which is what I wanted to keep.

Re-measured everything after the changes, since 7 and 8 moved the code —
`main`'s spelling written out and alternated against this one, one build,
minimum of 11 rounds of 200 000 calls, every pair asserted to answer alike
first, noise rows at 0.5% and 0.1%:

| | main | now |
| --- | --- | --- |
| `_drop_y`, the conversion alone | 0.2378 | 0.1729 |
| `xonly.serialize` | 0.2682 | 0.2006 |
| `silentpayments.serialize_label` | 0.2700 | 0.2040 |
| `_secret.take` | 0.1824 | 0.1457 |
| `xonly.parse`, 65 bytes | 0.5092 | 0.4309 |
| `xonly._from_pubkey_`, 65 bytes | 0.5381 | 0.4669 |
| `keys.prvkey_negate` | 0.3968 | 0.3599 |
| `keys.serialize`, compressed | 0.3130 | 0.2944 |
| `keys.serialize`, uncompressed | 0.3061 | 0.2911 |

582 passed, coverage 100.00% (776 statements, 164 branches), every hook
passes.

fametrano (commented):
## ACK

**ACK `c9057c5`.** The review state is `COMMENTED` and not `APPROVED` for
the same reason the last one could not be `CHANGES_REQUESTED`: GitHub
takes neither from the pull request's own author. Read the verdict from
here.

All five required are in, all three suggestions taken, and 8
changed the shape for the better: `_drop_y_with_parity` writes the
allocation once and keeps the C call written once too, which the
call-site spelling would have given up. Nothing below blocks the merge.

### Re-verified on `c9057c5`

Worktree at `c9057c5`, a descendant of `origin/main` at `75cbb44`,
submodule `6e2c8bc`, same machine as the branch measures on — Apple M5,
macOS 26.6, arm64, CPython 3.14.6.

- `pytest --cov` — 582 passed, 100.00%, 776 statements, 164 branches
- `sphinx-build -W --keep-going` — build succeeded
- `mypy btclib_secp256k1 tests` — clean; `ruff check` and `ruff format
  --check` — clean
- the probe fails the way the stub docstring says it does:
  `error: Argument 1 to "new" of "_FFI" has incompatible type "int";
  expected "str | _CType"  [arg-type]`

Re-derived, each `main` spelling put back into the branch's own function
so the two differ in nothing else, minimum of 15 rounds of 200 000 calls,
each row against a noise row running one spelling twice:

| | branch | here | noise |
| --- | --- | --- | --- |
| `_drop_y`, the conversion alone | −27.3% | −25.7% | −0.0% |
| `keys.serialize`, compressed | −5.9% | −4.6% | +0.0% |
| `_secret.take` | −20.1% | −16.7% | +3.5% |
| `xonly._from_pubkey_`, 65 bytes | −13.2% | −7.4% | +0.1% |

The two that differ have the reason on my side: my `main` spelling of
`_from_pubkey_` unpacks with `ffi.sizeof` where yours pins the constant,
so its baseline is not yours, and the `take` row's noise was 3.5% and is
worth no more than its sign. Every row agrees on direction and on the
call that carries the win.

### Three notes, none of them blocking

**The count moved and the wording did not.**
[`xonly.py:67`](https://github.com/btclib-org/btclib-secp256k1/blob/c9057c5dc126bfed37b1429e02c7be7bc590af97/btclib_secp256k1/xonly.py#L67)
and
[`CHANGELOG.md:179`](https://github.com/btclib-org/btclib-secp256k1/blob/c9057c5dc126bfed37b1429e02c7be7bc590af97/CHANGELOG.md#L179)
still say "three of the five callers". `_drop_y` has four call sites now
— `_drop_y_with_parity`, `_tweak_add_`, `parse`, `_parsed` — because 8
moved two of them behind the helper. The docstring survives the change
under the reading its next sentence supplies, five being the functions
whose need decides the default and `_drop_y_with_parity` not being one of
them; the CHANGELOG's "the five callers of `xonly._drop_y`" does not have
that sentence beside it and reads as the four. "Never state a count that
nothing checks" applies to both, and the sentence does not need the
number: *the callers that want the object alone — `parse`, `_parsed` and
`_tweak_add_` — are why NULL is the default*.

**Two sentences of history, of the kind 3 removed.**
[`xonly.py:102`](https://github.com/btclib-org/btclib-secp256k1/blob/c9057c5dc126bfed37b1429e02c7be7bc590af97/btclib_secp256k1/xonly.py#L102)
— "neither of them has three lines **where it had one**", which is the
call-site spelling of the previous push and one no reader of this tree
can encounter — and
[`xonly.py:639`](https://github.com/btclib-org/btclib-secp256k1/blob/c9057c5dc126bfed37b1429e02c7be7bc590af97/btclib_secp256k1/xonly.py#L639)
— "`ffi.sizeof` of a cdata -- **which is what this used to ask**, per
call". `keys.py:1075` is now the present-tense form of exactly this, so
the two are the same rule read twice. *the allocation is written once
rather than at each of the two call sites*, and *`ffi.sizeof` of a cdata
would be 0.0135 microseconds nothing reads*, say the same things without
the tense.

**The literal-cdecl pair is a direction I cannot reproduce.** The comment
and the CHANGELOG state 0.0670 against 0.0684 for `char[33]` and 0.0801
against 0.0815 for `size_t *`, the hoist the slower of each. Running the
pair in both orders, A/B and B/A in one process, minimum of 21 rounds of
300 000, the hoist comes out the faster of the two by a small constant
margin: `char[33]` 0.0579 against 0.0560, −3.2% in each order, on a noise
row of −0.1%; `int *` 0.0592 against 0.0573, −3.2%, on +0.7%. An earlier
campaign of mine said −7.4% and −4.4%. So the two harnesses disagree on
the sign and agree on the magnitude — a few thousandths of a microsecond,
an order below the 0.056 the f-string gives back, which is the number the
decision rests on and which we both measure the same. That is what the
comment can say without a reader re-deriving it and getting the other
sign: the literal is *within noise of* the hoist, against an f-string
that is twice the price. The code is right either way, which is why this
is a note.

Beside those: `_drop_y_with_parity`'s "the difference is at the noise
floor either way" is exact for `to_pubkey` — +1.1% on a noise row of
+1.1% — and a shade generous for `_from_pubkey_`, where isolating the
spelling with everything else the branch's gives +2.8% on a noise row of
+0.6%. The frame is real and it is 0.013 microseconds; "no saving lands
on these two" is the claim the figures carry.

Take or leave all four; none of them is a reason to hold the branch.
CI is green on all 26 checks. Issue:
[#212](https://github.com/btclib-org/btclib-secp256k1/issues/212).

fametrano (commented):
## ACK

**ACK `489bb5e`.** `COMMENTED` rather than `APPROVED` for the reason the
last two carried: GitHub takes neither state from the pull request's own
author. All four notes are in, and the branch is ready as it stands.

- the counts are gone from both sites, and the sentence reads better
  without them: *the callers that want the object alone — `parse`,
  `_parsed` and `_tweak_add_` — are why NULL is the default*
- the two narrated sentences are present tense: "written once rather than
  at each of their call sites", and "`ffi.sizeof` of the cdata *would be*
  0.0135 microseconds nothing reads"
- the literal-cdecl pair now claims what both harnesses agree on — within
  noise, direction unresolved, an order below the f-string's 0.056 —
  which is the figure the decision rests on and the one we measure alike
- `_drop_y_with_parity` says no saving lands on its callers, and the
  comments and the table now quote the same figures throughout: 0.1729
  against 0.2378 on the conversion, 0.4309 against 0.5092 on the parse

### Verified on `489bb5e`

The diff since `c9057c5` is comments, docstrings and CHANGELOG.md —
nothing executable moved — so what I gated on that revision stands, and
re-run here on the branch as it is: `pytest --cov` 582 passed, 100.00%,
776 statements and 164 branches; `sphinx-build -W --keep-going`
succeeded; `mypy btclib_secp256k1 tests` and `ruff check` clean. CI has
20 green with 2 still running, on jobs the local gates already cover.

### One number, for the record

`_drop_y_with_parity`'s frame is stated at 0.013 microseconds, in the
docstring and in CHANGELOG.md. That is the low end of what I get. Both
orders, A/B and B/A in one process, minimum of 21 rounds, everything but
the `drop_y` spelling the branch's: `_from_pubkey_` +6.80% (+7.97% and
+6.45% by order) on a noise row of −0.38%, which is 0.0296 microseconds,
and `to_pubkey` +2.33% (+2.33% and +2.83%) on +0.20%, which is 0.0179.
An earlier single-order run of mine said +2.8% and 0.013, which is where
we agreed; the better-powered one says about twice that, and a direct
`_drop_y_with_parity` against the inline pair says 0.024. So a Python
frame here is a few hundredths rather than 0.013, and the honest spelling
is the range or a re-run — the claim it supports, that no saving lands on
these two and the frame is real and resolvable, is right either way and
is why this is a note and not a request.

Nothing here blocks. Issue:
[#212](https://github.com/btclib-org/btclib-secp256k1/issues/212).

fametrano (commented):
## Review — changes requested

*(Posted as a comment because GitHub refuses `REQUEST_CHANGES` on one's own pull request; the verdict is the heading, not the event.)*

The change itself I have no objection to: it is correct, it is the optimisation it says it is, and I checked each of its claims rather than reading them. What blocks it is the figures — **three different measurements of the same rows are quoted in one commit**, and the two rows the description says it corrected are the two where the record that ships still disagrees.

### What I verified, and how

Worktree off `pr-215`, submodule initialised, `uv sync --locked`.

- `uv run --locked --no-default-groups --group test pytest --cov` — 582 passed, TOTAL 776 stmts / 164 branches, 0 missed, 100.00%. The suite and the gate are as the description states.
- `uv run --locked pre-commit run --all-files` — every hook Passed, `pydoclint` and `mypy` included.
- **The NULL parity is right against the vendored source, not just the quoted doc.** `secp256k1/include/secp256k1_extrakeys.h:88` says `pk_parity: Ignored if NULL`, and the declaration's `SECP256K1_ARG_NONNULL(1) ... (2) ... (4)` deliberately omits argument 3 — so upstream declares what `_drop_y`'s default does. `Returns: 1 always` besides, which is why the `RuntimeError` branch stays unreachable-by-a-valid-key exactly as its docstring says.
- **The stub's `_CType` earns its place, tested the way this repository asks.** I replaced `ffi.new(_XONLY_BUFFER)` with `ffi.new(_XONLY_SIZE)` and ran mypy: `error: Argument 1 to "new" of "_FFI" has incompatible type "int"; expected "str | _CType" [arg-type]`. The vacuous case the stub's docstring describes is the case it now catches.
- **`_zero` is equivalent to what it replaced.** Both call sites hand it `ffi.buffer(cdata)` over the whole cdata, so the length is still the buffer's own and `wipe`'s paragraph about `ffi.sizeof` saying 8 still holds.
- **The mutation argument in `keys.serialize` survives the hoist.** `_COMPRESSED_SIZE` is `ffi.sizeof` of the very type beside it, so a mutated `char[34]` moves the buffer *and* the unpack length together and the row still dies rather than surviving. The hoist did not weaken what that comment claims.
- **Nothing outside the package reaches `_drop_y`.** It appears in `xonly.py`, CHANGELOG.md and HISTORY.md and nowhere else; btclib does not reach for it, so the signature change costs no caller.

### Changes requested — the figures

Three sets, one commit:

| | CHANGELOG table (11 rounds x 200 000) | description and source comments (9 rounds) |
| --- | --- | --- |
| `silentpayments.serialize_label` | 0.2700 -> 0.2040 | 0.2583 -> 0.1912 |
| `xonly.parse`, 65 bytes | 0.5092 -> 0.4309 | 0.5071 -> 0.4168 |
| `_secret.take` | 0.1824 -> 0.1457 (-20.1%) | 0.1561 -> 0.1366 (-12.5%) |
| `keys.prvkey_negate` | 0.3968 -> 0.3599 (-9.3%) | 0.3418 -> 0.3211 (-6.1%) |

HISTORY.md quotes the left-hand column, so those two files agree; the description and two source comments quote the right-hand one. And `_secret.py`'s `_zero` docstring is a third: `wipe` twice at 0.2523 against `_zero` at 0.2080 is a delta of 0.0443 microseconds on a `take` that allocates per call, where the left column's `take` row saves 0.0367 and the right column's 0.0195. The buffer allocation is common to both spellings, so those three cannot all be the saving that one change is worth.

Each set is internally coherent — `prvkey_negate` is `take` plus `_scalar.scalar` and a C call, and in both columns its saving tracks `take`'s to the fourth decimal, which is what says these are two honest sessions rather than one typo. That is precisely why it matters which one is true: the description says `take` is **-12.5% and not -29%**, and the file that ships says **-20.1%**, a third value that neither #212 nor this description states. This PR closes #212, so the correction has to land where a reader will look for it.

So: pick the session, quote it in CHANGELOG.md, HISTORY.md, `silentpayments.py`, `xonly.py` and `_secret.py` alike — or, if both are to stand, say in each place which session it is and why the two differ. As it is, a reader who compares two files in this commit finds two answers, which is the one thing "never state a number nothing re-derives" is there to prevent.

### Suggested, not blocking

Three, inline. The naming one I would take even though nothing fails without it: in the module whose longest comment explains why a *shared buffer* would be a thread-safety bug, a module-level constant called `_COMPRESSED_BUFFER` reads like the thing that comment forbids.

With the figures reconciled this is an ACK from me — the code, the stub, the suite and the gate are all in order, and the NULL parity is the one behavioural change here and it is documented upstream.

CHANGELOG.md
```diff
@@ -139,6 +140,89 @@ release-notes length in the first place, and are still in
...
+  site they had.
+- Measured with the branch and `main`'s spelling of each call alternated
+  in one process against one build, minimum of 11 rounds of 200 000
+  calls, each against a noise row running `main`'s spelling twice, and
+  every pair asserted to answer alike before it is timed — an Apple M5,
+  macOS 26.6, arm64, CPython 3.14.6:
+
+  | | main | now |
+  | --- | --- | --- |
+  | `_drop_y`, the conversion alone | 0.2378 | 0.1729 |
```
fametrano: **This is the row that has to be settled before the rest.** 0.1824 -> 0.1457 is -20.1%. The description of this PR says, in bold, that `take` is "-12.5%, not -29%" and that #212's figure came from a variant with the zeroing inlined — and this table, which is what ships and what a reader of the release notes gets, states a third number that neither #212 nor the description gives.

`keys.prvkey_negate` two rows down has the same problem for the same reason: -9.3% here against the -6.1% the description calls the honest comparison. Those two rows are exactly the two the description says it re-measured, which is what makes the disagreement look like the correction not having reached this file rather than session noise.

What says the two columns are two real sessions rather than a slip is that each is internally coherent: `prvkey_negate` is `take` plus `_scalar.scalar` and a C call, and its saving tracks `take`'s to the fourth decimal in both. So one session has to be chosen and quoted everywhere, and if it is this one then the description's correction is the thing to fix instead.

btclib_secp256k1/silentpayments.py
```diff
@@ -43,6 +43,16 @@
 SUMMARY_SIZE = ffi.sizeof("secp256k1_silentpayments_prevouts_summary")
 LABEL_SIZE = 33
 
+# the buffer `serialize_label` writes into, resolved here rather than at
+# every call: `ffi.new` of an f-string formats it and leaves cffi to
+# hash the result, where `ffi.typeof` of the same string, evaluated
+# once, costs what a literal cdecl does and still states the width once.
+# 0.1912 microseconds against 0.2583 on this serialization -- an Apple
```
fametrano: 0.1912 against 0.2583 here; CHANGELOG.md's table says 0.2040 against 0.2700 for this same call, and HISTORY.md repeats the CHANGELOG's pair in prose. Same measurement, same commit, two answers — and the round counts differ too (9 rounds here, "minimum of 11 rounds of 200 000 calls" there), which is what identifies them as two sessions.

The delta survives either way (-26.0% against -24.4%), so nothing about the conclusion is at risk; what is at risk is that a comment states a number no other file in the tree agrees with.

btclib_secp256k1/xonly.py
```diff
@@ -41,37 +41,85 @@
...
     serialization that follows it in `_from_pubkey_`: `_tweak_add_` wants
     the object and not the 32 bytes, having a tweak to add to it.
 
     Args:
         pubkey: the already-parsed public key, as `keys.parse` returns.
+        parity: an `int *` to receive the parity of the y being dropped,
+            0 for even and 1 for odd -- or the NULL that says nobody is
+            reading it, which is the default because the callers that
+            want the object alone are why it exists: `parse`, `_parsed`
+            and `_tweak_add_`. libsecp256k1 documents `pk_parity` as
```
fametrano: Same split as `silentpayments.py`: 0.4168 against 0.5071 over 9 rounds of 100 000 here, where CHANGELOG.md gives 0.4309 against 0.5092 over 11 rounds of 200 000 for `xonly.parse` of a 65-byte key.

Note that the sibling figure just below, in `_drop_y_with_parity` — 0.161 against 0.233 on the conversion alone — *does* match the CHANGELOG (0.1611 against 0.2327). So this module quotes both sessions, a dozen lines apart, without saying that is what it is doing.

btclib_secp256k1/xonly.py
```diff
@@ -41,37 +41,85 @@
...
-    """Convert a parsed public key into a parsed x-only key and the parity.
+def _drop_y(pubkey: CData, parity: CData = ffi.NULL) -> CData:
+    """Convert a parsed public key into a parsed x-only key.
 
     The conversion itself, which is where the y is dropped, without the
     serialization that follows it in `_from_pubkey_`: `_tweak_add_` wants
     the object and not the 32 bytes, having a tweak to add to it.
 
     Args:
         pubkey: the already-parsed public key, as `keys.parse` returns.
```
fametrano: Suggested: "three of the five callers here" is a count of `main`'s caller set, in the present tense in the docstring of the function that replaced it. After this change `_drop_y` has four direct callers — `parse`, `_parsed`, `_tweak_add_` and `_drop_y_with_parity` — of which three take the default.

Read as "five callers of the conversion in this module" it is still defensible, `_from_pubkey_` and `to_pubkey` reaching it through the wrapper, so this is a reading-ambiguity rather than a falsehood. But it is a count nothing checks, sitting in the one place where the reader is holding the new signature: naming the three and dropping the arithmetic would say the same thing and stay true when a sixth caller arrives. The CHANGELOG's phrasing is fine as it stands — there the past tense is the subject.

btclib_secp256k1/keys.py
```diff
@@ -30,6 +30,25 @@
...
+# `ffi.sizeof` of the very type beside it, so the buffer and the length
+# cannot say different numbers, and stating the width once is what asked
+# for a type at module level in the first place. What it saves is the
+# `ffi.sizeof` per call, 0.0135 microseconds.
+#
+# What it does not save is the `ffi.new`: cffi already caches the parse
+# of a literal cdecl, so a hoisted type is within noise of one -- two
+# harnesses here put the pair a few thousandths of a microsecond apart
+# and disagree on which way, against the 0.056 an interpolated cdecl
+# gives back, which is an order more and is the figure the decision
```
fametrano: Suggested, and the one I would take: these four hold cffi *types*, not buffers. `_COMPRESSED_BUFFER = ffi.typeof("char[33]")` at module level reads as a buffer shared between calls — which is precisely what the comment inside `serialize` spends a paragraph explaining must not happen ("one failed call would leave a shared buffer at zero, and every later serialization — on any thread, of a perfectly good key — would be refused").

So the module now asserts a thread-safety property in prose while naming something as though it violated it, and the reader has to reach `ffi.new(_COMPRESSED_BUFFER)` to learn that a buffer is still allocated per call. `_COMPRESSED_BUFFER_TYPE` / `_UNCOMPRESSED_BUFFER_TYPE`, and `_XONLY_BUFFER_TYPE` / `_LABEL_BUFFER_TYPE` in the other two modules, costs nothing and removes the misreading.

btclib_secp256k1/_secret.py
```diff
@@ -41,6 +41,33 @@
...
+    M5, macOS 26.6, arm64, CPython 3.14.6, minimum of 25 rounds of
+    200 000 calls, on a noise row that moved 0.6% -- calling `wipe` a
+    second time is 0.2523 microseconds, this is 0.2080, and the inlined
+    statement is 0.1979. So the frame gives up about a fifth of what is
+    on the table, and what it buys is that the line which overwrites a
+    secret exists once: two copies are two things to keep true, in the
+    one facility whose whole claim is about what is left in memory.
+
+    Args:
+        memory: the `ffi.buffer` to overwrite. Typed as the stub
```
fametrano: Suggested: worth one line about what `memory` has to be a view *of*.

`wipe` could not be given a partial view — it takes the cdata and builds the view itself, which is what the paragraph about `ffi.sizeof` saying 8 defends. Now the statement that writes the zeros lives in a function that takes the view, so that invariant has moved out of the signature and into the two call sites. Both are right today: `wipe` passes `ffi.buffer(buffer)` and `take` passes the view it read the secret through, each the whole cdata.

But this is the module whose whole claim is about what is left in memory, and `_zero(ffi.buffer(buffer, 8))` would zero a quarter of a keypair and report success, which is the bug `wipe`'s docstring records as having happened once already. "The view is of the whole cdata, not a slice of one" in the Args entry keeps the reason findable from where the write now is.

(The equivalence itself I checked: both callers pass a full view, so this change alters nothing that gets zeroed.)

fametrano:
Pushed at 76bd7e1. **Changes requested taken: there is one session per
kind of measurement now, and every site says which one it quotes.**

You were right that the three sets could not all be the saving one
change is worth, and right that it was the correction not having reached
the shipping files. Rather than pick between two old sessions I re-ran
the lot — every wrapper figure in one process over one build, `main`'s
spelling written out beside each, 15 rounds of 200 000, every pair
asserted to answer alike first, noise at 0.0%:

| | main | now |
| --- | --- | --- |
| `_drop_y`, the conversion alone | 0.2425 | 0.1747 |
| `xonly.serialize` | 0.2569 | 0.1969 |
| `silentpayments.serialize_label` | 0.2621 | 0.2023 |
| `_secret.take` | 0.1826 | 0.1439 |
| `xonly.parse`, 65 bytes | 0.5021 | 0.4295 |
| `xonly._from_pubkey_`, 65 bytes | 0.5127 | 0.4711 |
| `keys.prvkey_negate` | 0.3790 | 0.3380 |
| `keys.serialize`, compressed | 0.3106 | 0.2906 |
| `keys.serialize`, uncompressed | 0.3101 | 0.2903 |

and the cffi primitives as a second, declared session, 15 rounds of
300 000: f-string against literal `char[32]` 0.1219 / 0.0681, literal
against hoisted `char[33]` 0.0685 / 0.0656, `size_t *` 0.0809 / 0.0774,
`ffi.unpack` with `ffi.sizeof` against the constant 0.0764 / 0.0589,
noise 0.0676 / 0.0680.

CHANGELOG.md carries both tables and says which figure comes from which;
HISTORY.md and the comments in `keys.py`, `xonly.py`,
`silentpayments.py` and `_secret.py` all quote that session and no
other. The description is rewritten to it too, so the `take` correction
now reads **−21.2%** everywhere instead of three numbers in three files.

**Your note on the frame was the better-powered measurement, and it
stands.** `_from_pubkey_` +3.1% and `to_pubkey` +3.5% here, which is
0.014 and 0.028 microseconds — the docstring and the CHANGELOG now give
both rather than the 0.013 that was the low end of one of them.

**And the literal-cdecl direction is resolved, in your favour.** At 15
rounds the hoist is faster in both pairs, 4.3% on a noise row of 0.5% —
my earlier single-order run said the opposite and was underpowered. So
the claim is no longer "within noise, direction unresolved" but the true
one: hoisting a literal saves 0.003 microseconds, real and an order
below the 0.054 the f-string gives back, which is the size a hoist has
to be worth to justify spelling a width twice or naming every cdecl in
the package.

**`_COMPRESSED_BUFFER` → `_COMPRESSED_BUFFER_TYPE`**, and the same for
the other three. You are right that a module-level `_BUFFER` reads as
the shared buffer the paragraph inside `serialize` spends its length
forbidding.

**`_zero`'s `Args`** now says the view is of the whole cdata and not a
slice of one, with `_zero(ffi.buffer(buffer, 8))` named as the failure
— the invariant did move out of `wipe`'s signature when the write did,
and that is where a reader will be standing.

The count note was already in at `489bb5e` — both sites name `parse`,
`_parsed` and `_tweak_add_` rather than counting.

582 passed, coverage 100.00%, every hook passes; the `_XONLY_SIZE` probe
still errors `[arg-type]`.

fametrano (commented):
## Review

**Changes requested** — two, both prose, both in what the branch adds at
`76bd7e1`. `COMMENTED` and not `CHANGES_REQUESTED` for the reason the
earlier ones carried: GitHub takes neither state from the pull request's
own author.

One session with every site naming it is the right resolution, and the
figures agree with mine where I can re-derive them. The renames to
`_..._TYPE` read correctly against the paragraph inside `serialize` that
spends its length forbidding a shared buffer. What is left is one number
that is wrong and one that says something the code cannot do.

### Verified on `76bd7e1`

Worktree at `76bd7e1`, descendant of `origin/main` at `75cbb44`,
submodule `6e2c8bc` — Apple M5, macOS 26.6, arm64, CPython 3.14.6, the
machine the session was run on.

- `pytest --cov` — 582 passed, 100.00%, 776 statements, 164 branches
- `sphinx-build -W --keep-going` — build succeeded
- `mypy btclib_secp256k1 tests`, `ruff check`, `ruff format --check` —
  clean; no `_BUFFER` name survives the rename
- the wrapper table re-derives: `_drop_y` −28.0% against my −25.7%,
  `keys.serialize` compressed −6.4% against −4.6%, `_from_pubkey_` −8.1%
  against −7.4%, `take` −21.2% against −16.7%. Same direction, same
  carrier, differences inside the baselines
- and the primitives do too, the disputed row included: literal against
  hoisted `char[33]` −4.2% here in your session, −3.2% in mine at 21
  rounds in both orders, noise 0.5% and −0.1%. The direction is settled
  and the comment now states it

### Required

**1. A keypair is 96 octets, so eight of them are a twelfth of one.**
[`_secret.py:66`](https://github.com/btclib-org/btclib-secp256k1/blob/76bd7e1847c6cf97b7953e2aa342f543f20248c6/btclib_secp256k1/_secret.py#L66):
"`_zero(ffi.buffer(buffer, 8))` would zero **a quarter of a keypair** and
report success". Measured:

```python
>>> len(ffi.buffer(keypair)), ffi.sizeof(keypair)
(96, 8)
```

Eight of ninety-six is a twelfth. The quarter is of the *private key* —
32 octets, the first of the struct — which is what
[`_secret.py:81`](https://github.com/btclib-org/btclib-secp256k1/blob/76bd7e1847c6cf97b7953e2aa342f543f20248c6/btclib_secp256k1/_secret.py#L81)
says fifteen lines below, in `wipe`, and says correctly: "wiped the first
quarter of the private key while reporting success". The new sentence is
the same fact restated wrong, in the same file, about the same eight
octets. *would zero a quarter of the private key in it and report
success* makes the two agree, and keeps the point, which is that the
danger is a view that is short rather than a length that is.

**2. One added frame cannot cost twice as much at one call site as at the
other.**
[`xonly.py:108`](https://github.com/btclib-org/btclib-secp256k1/blob/76bd7e1847c6cf97b7953e2aa342f543f20248c6/btclib_secp256k1/xonly.py#L108)
and
[`CHANGELOG.md:184`](https://github.com/btclib-org/btclib-secp256k1/blob/76bd7e1847c6cf97b7953e2aa342f543f20248c6/CHANGELOG.md#L184):
"the frame this adds is 0.014 microseconds on `_from_pubkey_` and 0.028
on `to_pubkey`". What each of those two pays is the same thing —
`_drop_y_with_parity` in place of the pair written out, one extra Python
call and one more argument passed, the tuple built in the callee instead
of the caller in both cases. There is no mechanism for `to_pubkey` to pay
double, and a reader who looks for one will not find it.

The figures are the harness. Three campaigns here, isolating the `drop_y`
spelling with everything else the branch's:

| | `_from_pubkey_` | `to_pubkey` |
| --- | --- | --- |
| 15 rounds, one order | +0.013 | — |
| 21 rounds, both orders | +0.0296 | +0.0179 |
| 25 rounds, both orders, noise 0.00% and −0.22% | +0.0173 | +0.0103 |
| this branch | +0.014 | +0.028 |

Mine put `_from_pubkey_` above `to_pubkey` every time, yours puts it at
half — which is the two of us measuring the same frame and reading the
spread, 0.010 to 0.030 microseconds, as if it were structure. One figure
for one frame is what the code supports: *the frame this adds is some
hundredths of a microsecond, measurable against both*, or a single number
with the spread owned. Two per-caller figures assert a difference that is
not there.

### One note

`HISTORY.md:272` still says "the parity `xonly._drop_y` allocated for
**three** callers that throw it away". It is true as it stands, and it is
the last of the three sites still counting — CHANGELOG.md and the
docstring both name `parse`, `_parsed` and `_tweak_add_` instead. Naming
them there too costs nothing and leaves no number to drift.

Both required items are one sentence each. Fix them and I will ACK.
Issue: [#212](https://github.com/btclib-org/btclib-secp256k1/issues/212).

fametrano:
Pushed at 87b72c1. Both required are in, and the note with them.

**1 — the arithmetic.** You are right, and I checked it rather than
taking it: `len(ffi.buffer(keypair))` is 96 and `ffi.sizeof(keypair)` is
8, so eight octets are a twelfth of the struct and a quarter of the
32-octet private key at the front of it. It now reads *would zero a
quarter of the private key inside a keypair and report success, which is
the failure that docstring records* — which agrees with `wipe` fifteen
lines below and with the v0.8.0.2 entry in CHANGELOG.md, both of which
already said "the first quarter of a private key". I had restated a
correct sentence wrongly in the same file.

**2 — one frame, one figure.** Taken. Your mechanism argument is the
decisive part: the same helper replaces the same inline pair at both
sites, so there is nothing for one of them to pay double for, and a
reader looking for it will not find it.

The figures make the case better than either of us reading them as
structure. Mine, 25 rounds, both orders in one process:

| | inline | helper | delta |
| --- | --- | --- | --- |
| `_from_pubkey_` A/B | 0.4610 | 0.4729 | +0.0118 |
| `_from_pubkey_` B/A | 0.4582 | 0.4698 | +0.0115 |
| `to_pubkey` A/B | 0.7935 | 0.8353 | +0.0418 |
| `to_pubkey` B/A | 0.7965 | 0.8338 | +0.0373 |
| noise | | | −0.0027, −0.0064 |

Stable in both orders, and it puts `to_pubkey` at three times
`_from_pubkey_` — where all three of your campaigns put it below. Two
harnesses, opposite orderings, both internally consistent: that is the
spread of a number this small, and it is exactly why the split should
not be in the tree. Both sites now say *some hundredths of a
microsecond, measurable against both*, and add that the same frame is
added at each and that two harnesses put a different one ahead — so the
negative result is recorded rather than re-derived.

**The note.** `HISTORY.md` was the last site counting; it names `parse`,
`_parsed` and `_tweak_add_` now, like the other two.

582 passed, coverage 100.00%, every hook passes.

fametrano (commented):
## ACK on the content, one thing to do before it lands

**ACK `87b72c1`** for everything the branch says and does: both required
items are fixed, correctly, and the note with them. `COMMENTED` rather
than `APPROVED` for the reason the earlier ones carried — GitHub takes
neither state from the pull request's own author.

- the arithmetic agrees with `wipe` fifteen lines below it now: *a
  quarter of the private key inside a keypair*, which is what eight of
  ninety-six octets are
- the frame is one figure, with the negative result recorded rather than
  left for the next reader to re-derive: *the same frame is added at
  each, and two harnesses put a different one of the two ahead*. Your
  numbers put `to_pubkey` at three times `_from_pubkey_` and all three of
  mine put it below — which is the case for the sentence you wrote
- `HISTORY.md` names the three rather than counting them, like the other
  two sites

### Required before merge: it is behind `main`, and the rebase is not free

`origin/main` has moved two commits since this branch was cut —
`b2b22be` and `e8902ee` — and GitHub says `CONFLICTING` / `DIRTY`. That
verdict is the union driver, not a real conflict: `git rebase
origin/main` here applies the commit with nothing to resolve, and the
rebased tree is sound — **627 passed, coverage 100.00%**, `mypy` clean on
36 source files, `ruff check` clean, `sphinx-build -W` succeeded. The
branch and `main`'s new `ssa`/`xonly` work do not collide.

What the rebase *does* do is glue this branch's prose onto `main`'s, in
both files, because `merge=union` keeps both sides' added lines and
neither side's separating blank:

| file | after the rebase | caught by |
| --- | --- | --- |
| `CHANGELOG.md:312` | `### What a wrapper works out per call` lands directly under `main`'s last bullet | MD022, so the markdownlint hook fixes it |
| `HISTORY.md:311-312` | `main`'s `__version__` paragraph and this one become **one paragraph** | **nothing** |

The second is the one to do by hand. markdownlint has no rule for two
paragraphs run together — it is valid Markdown that renders as one blob —
and no test reads that file's shape, so it would reach the published
release notes in silence. This is `.gitattributes`' own stated price
being paid for the first time: "two branches editing the same line merge
in silence".

So: `git fetch && git rebase origin/main`, put the two blank lines back,
`uv run pre-commit run --all-files`, force-push. Then it is a
fast-forward of one commit, which is what its sha wants.

### One for a follow-up, not for this branch

`e8902ee` landed `xonly._from_keypair_` while this was in review, and
[`ssa.py:598`](https://github.com/btclib-org/btclib-secp256k1/blob/e8902ee/btclib_secp256k1/ssa.py#L598)
reads it as `xonly_pubkey, _ = xonly._from_keypair_(keypair_obj)` — the
parity allocated and thrown away, on the path that verifies every
signature it makes. `secp256k1_extrakeys.h:210` documents
`secp256k1_keypair_xonly_pub`'s `pk_parity` as "Ignored if NULL", the
same contract `secp256k1_xonly_pubkey_from_pubkey` gives and the one this
branch's whole first bullet rests on. It is the shape you have just
removed from `_drop_y`, arriving in `main` from the other direction.

My recommendation is an issue rather than a widening of this pull
request: every figure here is one named session, and a sixth site would
need it re-run for a saving the same size as the one already measured.

Issue: [#212](https://github.com/btclib-org/btclib-secp256k1/issues/212).
CI is green on all 26 checks at `87b72c1`.

fametrano:
The follow-up is filed: https://github.com/btclib-org/btclib-secp256k1/issues/219 — `xonly._from_keypair_` allocating a parity that `ssa._abort_unless_verified` discards, with the header's "Ignored if NULL" quoted and the honest note that the figure does not carry over from `_drop_y`: it sits on a signing path that costs tens of microseconds, so the case is one convention rather than two, not speed. Nothing for this branch.

fametrano:
Pushed at 2e175b0 — the rebase, with both junctions repaired by hand.

**Reproduced your table before fixing it.** After `git rebase
origin/main` the two boundaries are exactly where you said:

```
CHANGELOG.md
  311| '  and reading the metadata directly all answer what they answered.'
  312| '### What a wrapper works out per call, and now works out once'
  blank line before? False

HISTORY.md
  311| 'cost, the first read and every one after it.'
  312| '**Serializing and parsing cost a little less, and nothing changed about'
  blank line before? False
```

Both blank lines are back, and the diff against `main` now shows each of
my additions starting after an unchanged blank line rather than against
the previous author's last line.

`main` had moved a third time since you wrote — `71aa9ac`, #214 — so the
rebase is over three commits rather than two. The verdict is unchanged:
nothing to resolve, and the tree is sound. **627 passed, coverage
100.00%**, every hook passes, which is your number.

**On the follow-up.** Agreed, and an issue rather than this branch, for
the reason you give: every figure here is one named session and a sixth
site would need it re-run. Checked before filing —
`secp256k1_extrakeys.h:210` reads `pk_parity: Ignored if NULL` for
`secp256k1_keypair_xonly_pub` too, and `xonly._from_keypair_` allocates
one that `ssa.py:598` discards with `xonly_pubkey, _ = ...`, on the
verify-after-sign path. Same shape, arriving from the other direction. I
will open it once this lands, so it can name the merged commit.

PR #216, opened 2026-08-17
Check a signature before answering with it, in both signing modules

fametrano:
BIP340 ends *Default Signing* with "If Verify(bytes(P), m, sig) returns
failure, abort", and Bitcoin Core ends `CKey::Sign` with the same step for
ECDSA, under a comment naming what it is for. Neither was done here.

What the step catches is not a bad argument — those have all raised by the
time it runs — but the computation itself going wrong, by bad memory or by
an induced fault, whose cost is a published signature that is invalid and
may say something about the key. In Schnorr what leaks is not vague:
two signatures over one nonce and two challenges give up `d` by subtraction.

## What changes

`dsa.sign` and `ssa.sign` verify what they made, under the public key of the
key that made it, and take a **keyword-only `verify` defaulting to `True`**.
`ssa.sign_custom`, both `Signer` methods and the three private halves
(`dsa._sign_`, `ssa._sign32`, `ssa._sign_custom`) take it with the same
default, so no path reaches libsecp256k1 with a weaker policy than the one
above it. A signature that fails is a `RuntimeError` and is never returned.

`verify` is keyword-only because `dsa.sign` reached six arguments, which
PLR0913 refuses; the `noqa` it carries has the reason above it. A caller
that wants the old cost passes `verify=False`, which is the whole of the
remedy — the signature is the same bytes either way.

`xonly._from_keypair_` is new, and is what keeps the BIP340 half cheap:
`xonly.from_keypair` is now that call with a serialize behind it, where
reaching the key through the public half would write a point out only to
lift it back, and a lift is a field square root.

## What it costs

Alternated in one process over 7 rounds of 20 000 calls, minimum kept for
each — an Apple M5, macOS 26.6, arm64, CPython 3.13.14, noise ±0.04:

| the call | `verify=False` | `verify=True` |
| --- | --- | --- |
| `dsa.sign` | 12.15 | 31.67 |
| `ssa.sign` | 15.87 | 28.57 |
| `ssa.Signer.sign` | 8.18 | 20.82 |

The finding is the gap between the two increments: 19.5 for ECDSA against
12.7 for BIP340, and the 6.8 between them is the point multiplication
`secp256k1_ec_pubkey_create` does and `secp256k1_keypair_xonly_pub` does
not. The step the specification prescribes is also the cheaper of the two,
which is why neither is opt-in.

`grind` is deliberately absent from that table: measured against a single
key whose first attempt already had the low r, those rows would have
measured no grinding at all.

## What is tested

The fault is not, no input producing one — the `raise RuntimeError` is
excluded from coverage for the reason every other one is. The wiring around
it is: the verification is substituted for one that refuses and each entry
point held to raising rather than returning; the signature is asserted equal
with the check and without it at each of eight entry points; the default is
read off every signature that has one; and keys of both y parities are
signed with, both parities asserted to occur, because a check read off the
wrong half of the negation BIP340 prescribes would pass for the wrong reason.

`tests/test_verified_signing.py` is new. `test_every_private_half_is_paired`
and `test_the_sweep_is_whole` both caught `xonly._from_keypair_` before this
description did, which is what those two are for.

Gate green, 609 tests, coverage 100.00%.

## Not in this change

`recovery.sign` still does not verify. Core closes that hole too, and in a
different shape — `CKey::SignCompact` recovers the key from the signature
and compares it with `secp256k1_ec_pubkey_cmp`, rather than calling
`ecdsa_verify` — so it is its own design, its own tests and its own
measurement. Worth a change of its own.

## Summary by Sourcery

Add default-on self-verification to ECDSA and Schnorr signing APIs, ensuring signatures are checked against the signing key before being returned, while allowing callers to opt out via a keyword-only flag.

New Features:
- Introduce a keyword-only `verify` parameter (defaulting to `True`) across public and internal ECDSA and Schnorr signing functions and Signer methods to control post-signature verification behavior.

Enhancements:
- Implement internal helpers to verify freshly produced signatures against their corresponding public keys for both ECDSA and BIP340 Schnorr flows, preventing invalid signatures from being emitted due to computation faults.
- Optimize Schnorr verification by deriving the x-only public key directly from keypair objects via a new `xonly._from_keypair_` helper, avoiding unnecessary serialization and decompression.
- Update changelog, README, history, and key parsing tests to document and validate the new verified-signing behavior and the x-only keypair helper, including coverage of edge cases around key parity and defaults.

Tests:
- Add `tests/test_verified_signing.py` to ensure verified signing does not alter signature bytes, that verification is enabled by default on all entry points, and that failures are correctly wired to runtime errors.
- Extend existing tests to cover the new `xonly._from_keypair_` helper and to keep private/public helper pairings in sync with the new functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,6 +577,79 @@ microseconds, and a caller who wanted the other form never has to pay it:
 `compact=True` answers the 64 octets of `r ‖ s` directly, where reaching
 them through `to_compact` writes the DER and parses it straight back.
 
+## Checking the signature before answering with it
+
+`dsa.sign` and `ssa.sign` verify what they just made, under the public
+key of the very key that made it, and a signature that fails is raised
+on rather than returned. It is the one argument here that defaults to
+on, and `verify=False` is how a caller declines it.
+
+The two have the same reasoning behind them and did not arrive by the
+same route. For BIP340 it is a step of the algorithm: *Default Signing*
+ends with "If Verify(bytes(P), m, sig) returns failure, abort", and the
+note beside it says why — "Verifying the signature before leaving the
+signer prevents random or attacker provoked computation errors. This
+prevents publishing invalid signatures which may leak information about
+the secret key. It is recommended, but can be omitted if the computation
+cost is prohibitive." Schnorr is linear, so what leaks is not vague: two
+signatures over one nonce and two challenges give up `d` by subtraction.
+For ECDSA no standard asks for it and Bitcoin Core does it anyway, at the
+end of `CKey::Sign`, under a comment reading "Additional verification
+step to prevent using a potentially corrupted signature" — and Core
+offers no way to turn it off, where this does.
+
+What it catches is not a bad argument: those have all raised by the time
+it runs. It catches the computation itself going wrong, whether by bad
+memory or by a fault induced on purpose, and the whole of the protection
+is not publishing the result.
+
+It costs what a verification costs, and in ECDSA a point multiplication
+besides — the public key, which BIP340 needs to sign at all and ECDSA
+needs for neither of the two things `sign` does. Measured with the
+variants alternated in one process, an Apple M5, macOS 26.6, arm64,
+CPython 3.13.14, seven rounds of 20 000 calls with the minimum of each
+kept, and a last row running one of them a second time so that the noise
+has a figure of its own:
+
+| the call | `verify=False` | `verify=True` |
+| --- | --- | --- |
+| `dsa.sign` | 12.15 | 31.67 |
+| `ssa.sign` | 15.87 | 28.57 |
+| `ssa.Signer.sign` | 8.18 | 20.82 |
+| `ssa.sign` again (the noise) | | ±0.04 |
+
+Microseconds per signature. The finding is the gap between the two
+increments: 19.5 for ECDSA against 12.7 for BIP340, and the 6.8 between
+them is the multiplication `secp256k1_ec_pubkey_create` does and
+`secp256k1_keypair_xonly_pub` does not — the keypair holds the point
+already, which is the same 7.55 the [signer](#two-outposts-past-the-boundary)
+hoists. So the step the specification prescribes is also the cheaper of
+the two, and a `Signer` pays it at the same price as a bare `ssa.sign`.
+
+**`recovery.sign` has no `verify`, and answers a signature nothing here
+has checked.** It is the one signing entry point this does not cover, and
+saying so is the point: the two sections above read as a policy of the
+package, and a caller reaching for a recoverable signature would
+otherwise conclude it was checked. The reason is that the check is a
+different one. Core closes the same hole in `CKey::SignCompact`, and not
+with a verification: it recovers the public key from the signature and
+compares it with `secp256k1_ec_pubkey_cmp`, which is the natural question
+to ask of a signature carrying a recovery id, and it is its own design,
+its own tests and its own measurement rather than this argument applied
+twice.
+
+What each half checks is not quite the same region either. `ssa` verifies
+the very 64 octets it answers with; `dsa` verifies the signature *object*
+and serializes it afterwards, so the DER or compact encoding is outside
+what was checked — as it is in Core, whose `CKey::Sign` verifies the
+`secp256k1_ecdsa_signature` and serializes after. Both are defensible,
+the serializers being memcpy-shaped where the signing is arithmetic, and
+the difference is worth a sentence rather than a change.
+
+Every other microsecond figure in this file was measured before this
+argument existed and is therefore the signature without the check:
+`verify=False` is the column they are in.
+
 ## Grinding for a low r
 
 `dsa.sign(grind=True)` answers the signature of the same key and message
--- a/btclib_secp256k1/dsa.py
+++ b/btclib_secp256k1/dsa.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 
 from typing import overload
 
-from . import BytesLike, CData, MutableBytesLike, ffi, lib
+from . import BytesLike, CData, MutableBytesLike, ffi, keys, lib
 from ._scalar import in_range, octets, optional_entropy, scalar
 from ._secret import take
 from .context import ctx
@@ -226,6 +226,8 @@ def _sign_(
     prvkey: BytesLike | int,
     aux_rand32: BytesLike | None = None,
     grind: bool = False,
+    *,
+    verify: bool = True,
 ) -> CData:
     """Create an ECDSA signature, as the parsed signature.
 
@@ -244,6 +246,12 @@ def _sign_(
             documents. Not with aux_rand32: the two write the same 32
             octets, so asking for both is refused rather than resolved
             here.
+        verify: whether to check the signature under the public key of
+            the key that made it before answering with it, as `sign`
+            documents. Where `grind` asks too, what is checked is the
+            signature the grinding settled on: the attempts it discarded
+            are never answered with, so a fault in one of them is a
+            wasted attempt rather than a published signature.
 
     Returns:
         The libsecp256k1 signature object, in the lower-s form
@@ -255,7 +263,8 @@ def _sign_(
             together, or if the private key is not 32 bytes, does not fit
             in them, or is not in [1, n-1].
         RuntimeError: if libsecp256k1 fails to serialize an attempt while
-            grinding, which a signature it has just made cannot do.
+            grinding, which a signature it has just made cannot do, or if
+            `verify` asks and the signature does not verify.
     """
     prvkey_bytes = scalar(prvkey, "private key")
     msg_bytes = octets(msg_bytes, "message hash", 32)
@@ -266,15 +275,33 @@ def _sign_(
     _signed(signature, msg_bytes, prvkey_bytes, optional_entropy(aux_rand32))
     if grind:
         _grind(signature, msg_bytes, prvkey_bytes)
+    if verify:
+        # through `_verify_` and `keys._pubkey_from_prvkey_` rather than
+        # through the two public halves, which would serialize this
+        # signature into DER and the point into octets only to parse both
+        # straight back -- and for a compressed key that parse is a field
+        # square root. `normalize` is not passed: libsecp256k1 has just
+        # answered the lower-s form, so there is nothing to normalize
+        pubkey = keys._pubkey_from_prvkey_(prvkey_bytes)
+        if not _verify_(msg_bytes, pubkey, signature):
+            raise RuntimeError("signing produced a signature that does not verify")
     return signature
 
 
-def sign(
+# six arguments, where PLR0913 allows five. The alternative is an options
+# object, and it would be one for this function alone: `aux_rand32`,
+# `compact` and `grind` are ECDSA's own three questions, `verify` is the
+# fourth these bindings add, and none of the four is a group with either
+# of the others. `verify` is keyword-only, as the two before it should
+# have been, so what a call site actually carries is named
+def sign(  # noqa: PLR0913
     msg_bytes: BytesLike,
     prvkey: BytesLike | int,
     aux_rand32: BytesLike | None = None,
     compact: bool = False,
     grind: bool = False,
+    *,
+    verify: bool = True,
 ) -> bytes:
     """Create an ECDSA signature.
 
@@ -297,6 +324,23 @@ def sign(
     and never done by default. `s` is not ground for and could not be:
     libsecp256k1 has already returned the lower of the two.
 
+    `verify` is the second thing beyond a libsecp256k1 call, and it is
+    the other half of that same `CKey::Sign`: the signature is checked
+    under the public key of the very key that made it, and one that
+    fails is never answered with. What it catches is not a bad argument
+    -- those have all raised by then -- but a computation that went
+    wrong, a bit flipped in memory or a fault induced on purpose, whose
+    cost is a published signature that is invalid and may say something
+    about the key. It is the one argument here that is on by default,
+    and the asymmetry with `grind` is the point: grinding buys an octet
+    of DER, this buys the guarantee that what was answered is a
+    signature. Core pays it on every signature and offers no way not to;
+    here there is one, because the check is a point multiplication and a
+    verification and the caller who has measured that against its own
+    threat model is the one entitled to refuse it -- `ssa.sign` refuses
+    it more cheaply, BIP340 needing the public key to sign at all where
+    ECDSA needs it for neither of the two things `sign` does.
+
     Args:
         msg_bytes: the 32-byte hash of the message.
         prvkey: the private key, 32 bytes or an int below 2**256.
@@ -310,6 +354,11 @@ def sign(
             32 octets, so a caller asking for both is asking for two
             different values of one argument, and Core's counter is what
             makes the result reproducible by anyone else.
+        verify: whether to check the signature under the public key of
+            the key that made it before returning it. It costs a point
+            multiplication and a verification, and False is what a
+            caller that has measured those against its own threat model
+            passes.
 
     Returns:
         The signature, in the lower-s form libsecp256k1 always produces:
@@ -341,7 +390,7 @@ def sign(
     # three land within 0.03 microseconds of this one and on the wrong
     # side of it. A signature is 11.9 microseconds of libsecp256k1, and
     # two python frames do not show against it
-    signature = _sign_(msg_bytes, prvkey, aux_rand32, grind)
+    signature = _sign_(msg_bytes, prvkey, aux_rand32, grind, verify=verify)
     return serialize_compact(signature) if compact else serialize_der(signature)
 
 
--- a/btclib_secp256k1/ssa.py
+++ b/btclib_secp256k1/ssa.py
@@ -136,7 +136,11 @@ def nonce_bip340(
 
 
 def sign(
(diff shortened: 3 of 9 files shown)


fametrano (commented):
## Review — changes requested

*(Posted as a comment because GitHub refuses `REQUEST_CHANGES` on one's own pull request; the verdict is the heading, not the event.)*

The design is right, the security argument holds under checking, and the figures agree across every file that states them. What blocks it is one thing, and I proved it rather than suspected it: **`verify=False` is untested for the only thing it does.** Replace all three `if verify:` with `if True:` — the flag entirely ignored, the opt-out silently dead — and the suite is 609 passed. Two smaller items follow it.

### What I verified, and how

Worktree off `pr-216`, submodule initialised, `uv sync --locked`.

- `uv run --locked --no-default-groups --group test pytest --cov` — 609 passed, 782 stmts / 168 branches, 0 missed, 100.00%. As the description states.
- `uv run --locked pre-commit run --all-files` — every hook Passed, `mypy`, `pydoclint` and `ruff` included, the `noqa: PLR0913` doing its job.
- **The BIP340 half checks against the right point, and I went to the vendored source rather than taking the docstring's word.** `secp256k1_keypair_xonly_pub` does not re-derive: `secp256k1_keypair_load` casts `&keypair->data[32]` and `pubkey_load`s the *stored* public half (`secp256k1/src/modules/extrakeys/main_impl.h:183`). So the check reads the same bytes the signature was computed against — `secp256k1_schnorrsig_sign_internal` loads `pk` from that very struct (`schnorrsig/main_impl.h:139`) and negates `sk` on its parity.

  That is sound, and worth saying why, because "the check uses what signing used" is the shape a verify-after-sign step usually gets wrong. The verification takes only `x(P)` and lifts it to the even-y point, so it tests the consistency of `s` with the point being claimed: a fault in the stored secret half gives `s = k + e·d_faulty` against `P = d·G` and fails; a fault in the stored point changes `x(P)` and fails; a fault that only flips the stored `y`'s parity flips the negation and fails, `x` being unchanged and `e` with it. The one case that slips through is a wholly self-consistent wrong keypair — and that case is equally invisible to Core's ECDSA shape, which derives from the same possibly-corrupted secret bytes. So reading the point off the keypair buys the 6.8 microseconds at no loss of what the step catches.
- **`_verify_` is the right verification for `_sign_custom`.** It passes `len(msg_bytes)` to `secp256k1_schnorrsig_verify`, so the variable-length half is not being checked by a 32-byte-only verifier — which is the mistake that shape invites.
- **`recovery.sign` is genuinely untouched**, not accidentally swept in: `recovery._sign_` is its own function calling `secp256k1_ecdsa_sign_recoverable`, and nothing routes it through `dsa._sign_`. The scope statement is accurate.
- **The figures agree everywhere they appear** — 12.15/31.67, 15.87/28.57, 8.18/20.82 in the description, CHANGELOG.md, HISTORY.md and README.md alike, one session, one machine. The README's closing line — that every other microsecond figure in the file predates this argument and is therefore the `verify=False` column — is the sentence that keeps the rest of that file true, and I could not find a figure it fails to cover.
- **The `grind` ordering is right**: `_grind` settles first and the check is on what it settled on, so a fault in a discarded attempt costs an attempt rather than a publication, exactly as `_sign_`'s docstring claims.

### Necessary — the opt-out has no test

Run on this branch, `PYTHONDONTWRITEBYTECODE=1`, all three `if verify:` replaced by `if True:`:

```
609 passed in 2.64s
```

The mutant that ignores the flag survives. `test_the_check_changes_no_signature` cannot catch it — it asserts the two calls answer the same bytes, which they do whether or not `verify=False` was honoured — and `test_a_signature_that_does_not_verify_is_not_answered_with` only ever exercises the default. So the suite holds that the check happens and says nothing about it being declinable, which is half of what this pull request adds and the half a caller reaches for under load.

The opposite mutant does die (`if False:` → that test fails, 1 failed / 608 passed), which is what says the wiring test works in one direction only.

The fix is three lines in the test that already holds the wiring, and I ran it: with them, the surviving mutant fails and the branch stays green. Inline below.

This is the class of gap `.github/mutation/bindings.toml` exists for — "the line still runs and coverage still reads 100%" — and a weekly `mutation` session would find it; better it does not have to.

### Necessary — `recovery.sign` is exempt and the shipped docs do not say so

CHANGELOG.md announces "A signer checks its own signature", README.md gains "Checking the signature before answering with it", and HISTORY.md calls it "the one change to act on that is not a rename". None of the three mentions that `recovery.sign` still answers an unchecked signature. The scope note lives only in this description, which is the one place a reader of the release will not look.

README.md:701 already carries the sibling asymmetry — "`recovery.sign` has no `grind`, and the reason is that…" — so the file has a place for this and an established shape for saying it. One sentence there and one in the CHANGELOG entry, plus an issue for the `SignCompact` design, and the omission is stated rather than discovered.

### Suggested

Two, inline: the doctest in `ssa.sign` demonstrates nothing about the argument it was added for, and `_verified` is a name that reads like a predicate in a module that already has one.

With the opt-out test and the `recovery.sign` sentence this is an ACK from me. The step itself is the right step, defaulting to on is the right default, and reading the point off the keypair is a real saving that costs nothing in what the check catches.

tests/test_verified_signing.py
```diff
@@ -0,0 +1,183 @@
...
+        patch.setattr(ssa, "_verify_", refusing)
+        with pytest.raises(RuntimeError, match="does not verify"):
+            ssa.sign(MSG, PRVKEY, AUX)
+        with pytest.raises(RuntimeError, match="does not verify"):
+            ssa.sign_custom(MSG, PRVKEY, AUX)
+        with (
+            pytest.raises(RuntimeError, match="does not verify"),
+            ssa.Signer(PRVKEY) as signer,
+        ):
+            signer.sign(MSG, AUX)
```
fametrano: **This is the blocker.** This test holds one direction only. With all three `if verify:` replaced by `if True:` on this branch — the flag ignored, `verify=False` silently dead — the suite is `609 passed`. `test_the_check_changes_no_signature` cannot see it either: it asserts the two calls answer the same bytes, which they do whether or not the False was honoured.

The three lines that close it, which I ran — the mutant above fails with them and the branch stays green:

```suggestion
            signer.sign(MSG, AUX)

    # the other half: verify=False does not reach the check at all, which
    # no assertion above can see -- the signature is the same bytes either
    # way, so a flag that was ignored would answer exactly what it should
    with pytest.MonkeyPatch.context() as patch:
        patch.setattr(dsa, "_verify_", refusing)
        patch.setattr(ssa, "_verify_", refusing)
        assert dsa.sign(MSG, PRVKEY, verify=False)
        assert ssa.sign(MSG, PRVKEY, AUX, verify=False)
        assert ssa.sign_custom(MSG, PRVKEY, AUX, verify=False)
```

Worth extending to the `Signer` methods and `dsa._sign_` on the same principle, the eight entry points being the point of the table above.

And this is the gap `.github/mutation/bindings.toml` was written for — "change `!=` to `<=`, or 32 to 31, and the line still runs and coverage still reads 100%". The weekly session would find it; it should not have to.

README.md
```diff
@@ -577,6 +577,59 @@ microseconds, and a caller who wanted the other form never has to pay it:
 `compact=True` answers the 64 octets of `r ‖ s` directly, where reaching
 them through `to_compact` writes the DER and parses it straight back.
 
+## Checking the signature before answering with it
+
+`dsa.sign` and `ssa.sign` verify what they just made, under the public
```
fametrano: **Necessary:** this section, the CHANGELOG entry and HISTORY.md's "the one change to act on that is not a rename" all read as a package-wide policy, and `recovery.sign` is not in it. A reader who signs with `recovery.sign` will conclude from these three files that the signature was checked, and it was not.

The scope note is in the pull request description, which is the one place a reader of the release does not look. README.md:701 already says "`recovery.sign` has no `grind`, and the reason is that…", so this file has both a place for the second asymmetry and a shape for stating it — and the reason is a good one rather than an omission: Core closes that hole through `CKey::SignCompact`, recovering the key and comparing with `secp256k1_ec_pubkey_cmp`, which is its own design and its own measurement.

One sentence here, one in the CHANGELOG entry, and an issue for the `SignCompact` shape. There is no open issue for it — I looked.

btclib_secp256k1/ssa.py
```diff
@@ -163,18 +183,22 @@ def sign(
...
 
     Example:
         >>> from btclib_secp256k1 import keys, ssa, xonly
         >>> msg, prvkey = bytes(32), 1
         >>> pubkey, _ = xonly.from_pubkey(keys.pubkey_from_prvkey(prvkey))
-        >>> ssa.verify(msg, pubkey, ssa.sign(msg, prvkey, bytes(32)))
+        >>> sig = ssa.sign(msg, prvkey, bytes(32))
+        >>> ssa.verify(msg, pubkey, sig)
+        True
+        >>> ssa.sign(msg, prvkey, bytes(32), verify=True) == sig
```
fametrano: Suggested: this example demonstrates nothing about the argument it was added for. `sig` two lines up was made with the default, and the default *is* `verify=True`, so this compares the checked path against itself. I substituted `verify=False` and ran `tests/test_examples.py`: 16 passed either way, which is what says the line is vacuous rather than merely redundant.

`verify=False` is also the claim the prose around it makes — "the signature is the same bytes either way" — and the one a reader of this example wants to see, the whole question about an argument like this being whether declining it changes the answer:

```suggestion
        >>> ssa.sign(msg, prvkey, bytes(32), verify=False) == sig
```

What this line currently asserts — that the default is the checked path — is already held by `test_every_signer_defaults_to_checking` and `test_the_check_is_on_where_nothing_asks`, and held there against all eight entry points rather than one.

btclib_secp256k1/ssa.py
```diff
@@ -500,8 +552,51 @@ def verify(
     )
 
 
+def _verified(keypair_obj: CData, msg_bytes: BytesLike, signature_bytes: bytes) -> None:
```
fametrano: Suggested, a name: `_verified` reads as a predicate, and this module already has the predicate — `_verify_`, thirty lines up, returning `bool`. So the two neighbours are `_verify_` which answers and `_verified` which raises, distinguished by two characters of tense.

The failure that invites is quiet: `if _verified(...)` and `if not _verified(...)` are both syntactically fine, the first never raising on its own and the second raising always, and neither reads as wrong at the call site. `_verify_or_raise`, or `_abort_unless_verified` after the BIP's own "abort", says which of the two it is in the name.

The body I have no quarrel with — I checked the vendored source, and reading the point off the keypair is sound as well as cheap, for the reason in the review.

btclib_secp256k1/dsa.py
```diff
@@ -266,15 +275,33 @@ def _sign_(
...
     if grind:
         _grind(signature, msg_bytes, prvkey_bytes)
+    if verify:
+        # through `_verify_` and `keys._pubkey_from_prvkey_` rather than
+        # through the two public halves, which would serialize this
+        # signature into DER and the point into octets only to parse both
+        # straight back -- and for a compressed key that parse is a field
+        # square root. `normalize` is not passed: libsecp256k1 has just
+        # answered the lower-s form, so there is nothing to normalize
+        pubkey = keys._pubkey_from_prvkey_(prvkey_bytes)
```
fametrano: An observation rather than a change, since Core does exactly this: what is checked here is the signature *object*, and `sign` serializes it afterwards, so the DER or compact encoding is outside the checked region. `ssa` verifies the very 64 bytes it answers with. Both are defensible — the serializers are memcpy-shaped where the signing is arithmetic, which is where a fault matters — but README.md's "the whole of the protection is not publishing the result" is very slightly stronger for the BIP340 half than for this one, and the asymmetry is invisible from the prose.

The derivation itself is right and is the reason this half costs more: `keys._pubkey_from_prvkey_` re-derives from the private key bytes rather than reading a cached point, which is what makes it Core's step and not a weaker one.

fametrano:
Both **Necessary** items are in, both **Suggested** ones taken, and the observation turned into the sentence it asked for. Amended into the one commit (`468ebd7` → `cd58ff3`) so the sha the matrix runs is the sha that lands.

## The blocker — reproduced first, then closed

I did not take the mutant on trust. On `468ebd7`, all three `if verify:` replaced by `if True:`, `PYTHONDONTWRITEBYTECODE=1`:

```
609 passed in 2.80s
```

Exactly as you found. The gap is real and the diagnosis is exact: `test_the_check_changes_no_signature` compares two calls that agree whether or not the flag was honoured, and the wiring test only ever ran the default.

`test_the_refused_check_is_not_made_at_all` is new, and rather than the three lines it takes the whole `SIGNERS` table — the refusing verification is patched into both modules and `signer(verify=False)` is asserted to answer at all eight entry points, on your own principle that the table is the point. I also parametrized the existing wiring test over the same eight, which it was not: it hand-listed five and missed `dsa.sign` compact, `dsa.sign` grind and `Signer.sign_custom`.

Both mutants now die, eight failures each, and the control is clean:

| the tree | result |
| --- | --- |
| `if True:` — the flag ignored | **8 failed**, 616 passed |
| `if False:` — the check never made | **8 failed**, 616 passed |
| restored | 624 passed, coverage 100.00% |

## `recovery.sign` — stated in all three files, and an issue

README.md's new section carries it in the shape you pointed at, beside the `grind` asymmetry it already had; the CHANGELOG entry and HISTORY.md each carry a sentence. **#217** has the `SignCompact` design — the `ecdsa_recover` + `ec_pubkey_cmp` shape, why it is the stronger question here (it checks the recovery id, which a plain verification does not look at, and a wrong id verifies fine and recovers the wrong key), what it would take, and the three files to correct when it closes.

## Suggested, both taken

- **The doctest was vacuous** and you are right about why: `sig` was made with the default, so it compared the checked path with itself. It is now `verify=False`, which is the claim the prose beside it makes.
- **`_verified` → `_abort_unless_verified`**, with your reasoning in its docstring. I took the second of your two names because the verb is BIP340's own. Worth noting `dsa._signed` is the same past-participle-that-raises shape, so the precedent existed — but its neighbour is not a `bool`-returning `_sign_`, and yours is the case where the confusion is actually available.

## The observation

Taken as prose, not as a change, since Core does the same: the README now says `ssa` verifies the 64 octets it answers with while `dsa` verifies the object and serializes after, so the encoding is outside the checked region — the serializers being memcpy-shaped where the signing is arithmetic. That removes the asymmetry the file was hiding.

Thank you for going to the vendored source on `keypair_xonly_pub` rather than taking the docstring's word. The enumeration of which faults the stored-point read still catches, and the one it does not, is a better argument for that 6.8 microseconds than the one in my description — I would not have written it that precisely.

fametrano (commented):
## ACK — cd58ff3

Every item from the previous review is addressed, and I re-ran the checks rather than reading the diff for them.

### The blocker is closed, and I measured that it is

The mutant that ignores the flag — every `if verify:` replaced by `if True:`, `PYTHONDONTWRITEBYTECODE=1` — now dies, and dies at all eight entry points:

```
test_the_refused_check_is_not_made_at_all[dsa.sign]
test_the_refused_check_is_not_made_at_all[dsa.sign compact]
test_the_refused_check_is_not_made_at_all[dsa.sign grind]
test_the_refused_check_is_not_made_at_all[dsa._sign_]
test_the_refused_check_is_not_made_at_all[ssa.sign]
test_the_refused_check_is_not_made_at_all[ssa.sign_custom]
test_the_refused_check_is_not_made_at_all[Signer.sign]
test_the_refused_check_is_not_made_at_all[Signer.sign_custom]
8 failed, 616 passed
```

That is the whole of the kill set, which makes the claim in the module docstring — "Replacing every `if verify:` in the package with `if True:` leaves every other test here passing" — exactly true rather than approximately. The opposite mutant (`if False:`) dies the same way, 8 failures under `test_a_signature_that_does_not_verify_is_not_answered_with`, so both directions are now held at every entry point where before only one was held at some.

Parametrizing both over `SIGNERS` is more than I asked for and is the better shape: I suggested three call sites and this covers eight, and it turned the older ad-hoc test into the same table.

### The rest, re-checked

- `uv run --locked --no-default-groups --group test pytest --cov` — 624 passed (609 before), 782 stmts / 168 branches, 0 missed, 100.00%.
- `uv run --locked pre-commit run --all-files` — every hook Passed.
- **`recovery.sign`** is now stated in all three shipped files — README.md beside the `grind` asymmetry it already carried, CHANGELOG.md, HISTORY.md — and #217 carries the `SignCompact` design with Core's own code quoted. The omission is a documented decision now rather than something a reader discovers.
- **The doctest** is `verify=False`, so the line demonstrates the claim the prose makes instead of comparing the default with itself.
- **`_abort_unless_verified`** — and the paragraph explaining the choice against `_verify_` thirty lines up is the part that will keep it named that way.
- **The checked-region asymmetry** I raised as an observation and not a change is documented in both README.md and CHANGELOG.md, with Core's shape named. That is more than the finding asked for.
- Nothing new introduced: the `noqa: E731` lambda became a documented function, no line in the changed files exceeds 80 columns except pre-existing unbreakable links, and the figures still agree across the description, CHANGELOG.md, HISTORY.md and README.md.

### One note, for after this lands, not for this branch

`btclib-benchmarks` resolves this package from `main`, and its wrappers page times `dsa.sign` and `ssa.sign` against three other wrappers of the same C library. The moment this is on `main` those rows would start verifying by default and stop being comparable with coincurve's and secp256k1-py's, which do not. That is the benchmark's change to make and not this one's — the rows there now pass `verify=False` explicitly and carry a checked row beside each, which is the same pair shape `grind` already had. Worth knowing that the lock upgrade and the re-measurement go together.

ACK cd58ff3e80b8f4bbca4a2b875dd2f227a461578f.

PR #218, opened 2026-08-17
Recover the key a recoverable signature claims, before answering with it

### LINKED ISSUE #217: recovery.sign answers a signature nothing has checked
[#216](https://github.com/btclib-org/btclib-secp256k1/pull/216) gave `dsa.sign` and `ssa.sign` a `verify` defaulting to `True`: the signature is checked under the public key of the key that made it, and one that fails is raised on rather than returned. BIP340 puts that step inside *Default Signing*; Bitcoin Core puts it at the end of `CKey::Sign`.

`recovery.sign` was left out, deliberately and for a reason: the check a recoverable signature wants is not the one `dsa.sign` makes.

## The shape Core uses

`CKey::SignCompact` does not call `secp256k1_ecdsa_verify`. It recovers the public key from the signature it just made and compares it with the one derived from the private key:

```cpp
// Additional verification step to prevent using a potentially corrupted signature
secp256k1_pubkey epk, rpk;
ret = secp256k1_ec_pubkey_create(secp256k1_context_sign, &epk, UCharCast(begin()));
assert(ret);
ret = secp256k1_ecdsa_recover(secp256k1_context_static, &rpk, &rsig, hash.begin());
assert(ret);
ret = secp256k1_ec_pubkey_cmp(secp256k1_context_static, &epk, &rpk);
assert(ret == 0);
```

That is the stronger question for this signature, and the natural one: it checks the recovery id as well, which a plain verification does not look at. A recoverable signature whose id is wrong verifies fine as a signature and recovers the wrong key — which is precisely what a caller of `recovery.sign` is going to use it for.

## What it would take

- `recovery._sign_` grows the same keyword-only `verify`, defaulting to `True`, and `recovery.sign` passes it through. Every wrapper the check needs is already here: `keys._pubkey_from_prvkey_`, `recovery._recover_` and `keys._pubkey_cmp_`, so it composes out of private halves without a new C call.
- A measurement of its own. It is a derivation plus a recovery plus a comparison, where `dsa.sign`'s is a derivation plus a verification — the two are not the same price and the README table should say which is which.
- Its own tests, on the pattern `tests/test_verified_signing.py` sets: the flag honoured in both directions, the substituted refusal wired to the raise, and the default read off the signature.
- One sentence each in README.md, CHANGELOG.md and HISTORY.md, replacing the three that currently say this hole is open.

## Where it is currently documented as open

README.md's *Checking the signature before answering with it*, the CHANGELOG entry for v0.8.0.3, and HISTORY.md's note beside the `verify=False` diff all state the exemption, so nothing claims a coverage the package does not have. Those are the three places to update when this closes.

fametrano:
Closes #217.

[#216](https://github.com/btclib-org/btclib-secp256k1/pull/216) gave `dsa.sign` and `ssa.sign` a `verify` defaulting to `True` and left `recovery.sign` out, deliberately: the reason it was exempt is the reason it needs a different check.

## Why not a verification

**A verification does not look at the recovery id**, which is the whole of what this signature carries beyond a plain one. A recoverable signature with the wrong id verifies perfectly and then recovers a key that is not the signer's — and recovering a key is the one thing a caller of this module is going to do with the answer. Verifying it would have been a check that passes while the thing the caller wants is broken.

So the check is the one the id deserves: recover from the signature, compare with the key that signed, refuse anything else.

That is Bitcoin Core's own distinction, made in one file:

| Core | ends in |
| --- | --- |
| `CKey::Sign` | `secp256k1_ecdsa_verify` |
| `CKey::SignCompact` | `secp256k1_ecdsa_recover` + `secp256k1_ec_pubkey_cmp` |

both under the same "Additional verification step to prevent using a potentially corrupted signature". And it subsumes the verification rather than skipping it: a key comes back from `r` and `s` only where those two are consistent, so a fault in either gives a different key or no key at all.

## What changes

`recovery.sign` and `recovery._sign_` take the same keyword-only `verify: bool = True`. `_abort_unless_recovered` is named as `ssa._abort_unless_verified` is, and does the one thing this shape needs beyond the others: `_recover_` reports "no key can be recovered" as a `ValueError` — an argument error for a signature a caller handed in, and nothing of the sort for one made a line earlier — so it becomes the `RuntimeError` this package means by a computation having gone wrong.

## What it costs

Measured in a session of its own, which **re-measured `dsa.sign` beside it** rather than printing the two together. Apple M5, macOS 26.6, arm64, CPython 3.13.14, 7 rounds of 20 000 calls, minimum kept, noise ±0.02:

| the call | `verify=False` | `verify=True` |
| --- | --- | --- |
| `recovery.sign` | 12.02 | 34.41 |
| `dsa.sign`, in this same session | 12.06 | 31.54 |

The second row agrees with the 12.15 / 31.67 already in the README, which is what makes the first comparable. So this shape costs **22.4** against ECDSA's 19.5 — a recovery being about a verification's work, with the comparison and the derivation making up the rest.

## What is tested

Both entry points join `tests/test_verified_signing.py`'s table, so the four questions it already asks are asked of them too — the check changes no signature, it is on where nothing asks, a refusal is wired to the raise, and `verify=False` does not reach the check at all. The substituted refusal is now per module, since these two reach a recovery where the others reach a verification.

**The failure a verification would have missed has a test of its own**: a recovery substituted to answer a valid key that is not the signer's, held to being refused — and then held to answering normally under `verify=False`, which is what says the refusal is the comparison and not the recovery.

Both mutants die:

| the tree | result |
| --- | --- |
| recovery's `if verify:` → `if True:` | **3 failed**, 635 passed |
| recovery's `if verify:` → `if False:` | **3 failed**, 635 passed |
| restored | 638 passed, coverage 100.00% |

Before the change neither direction was covered at all.

## Documentation

README.md, CHANGELOG.md and HISTORY.md each carried a sentence saying this hole was open; all three now carry the check instead. HISTORY.md closes with the thing a user wants to know — every signing call of the package now checks itself, and the same `verify=False` declines any of them.

## Summary by Sourcery

Add verification-style safety checks to recoverable ECDSA signing so that returned signatures are guaranteed to recover the original key by default, and document and test this behavior across the library.

New Features:
- Introduce a `verify` keyword argument (defaulting to True) to `recovery.sign` and `recovery._sign_` to validate that freshly produced recoverable signatures recover the signing key.

Bug Fixes:
- Prevent recoverable signatures with incorrect recovery IDs from being returned by refusing signatures that either fail key recovery or recover a different public key.

Enhancements:
- Add `_abort_unless_recovered` to encapsulate post-signing recovery-and-compare checks and convert internal recovery `ValueError`s into `RuntimeError`s when signing goes wrong.
- Extend the verified-signing test matrix and helpers to cover recovery-based signing, including new tests ensuring mismatched recovered keys are rejected and that `verify=False` bypasses the check.
- Update README, CHANGELOG, and HISTORY to describe the new recovery-based check, its cost, and the unified `verify` semantics across all signing entry points.

Tests:
- Broaden `tests/test_verified_signing.py` to cover recovery-based signing, with shared helpers to force verification/recovery failures and a dedicated test for signatures that would recover another key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,23 +54,53 @@ release-notes length in the first place, and are still in
 - **A caller that signs and does not publish pays for a guarantee it may
   not want**, and `verify=False` is the whole of the remedy: the
   signature is the same bytes either way, which
-  `tests/test_verified_signing.py` asserts for each of the eight entry
-  points rather than for one of them.
-- **`recovery.sign` is not in it, and the shipped documentation says so
-  rather than leaving it to a pull request description.** It answers a
-  signature nothing here has checked, and the reason is that the check
-  is a different one: Core closes the same hole in `CKey::SignCompact`
-  by recovering the public key and comparing it with
-  `secp256k1_ec_pubkey_cmp`, which is the question a recovery id invites
-  and is its own design, tests and measurement -- #217 carries it, and
-  names the three files to correct when it closes. README.md carries
-  the asymmetry beside the one it already carries for `grind`.
-- **The two halves do not check quite the same region**, which is worth
-  a sentence and not a change. `ssa` verifies the 64 octets it answers
+  `tests/test_verified_signing.py` asserts at each entry point rather
+  than at one of them.
+- **`recovery.sign` is in it too, and asks a different question** (#217).
+  It takes the same keyword-only `verify`, defaulting to `True`, and so
+  does `recovery._sign_`; what it does with it is recover the public key
+  from the signature and refuse one that is not the signer's, rather
+  than verify. The difference is the recovery id, which a verification
+  does not look at: a signature carrying the wrong one verifies
+  perfectly and then recovers somebody else's key, and recovering a key
+  is what a caller of that module does with the answer. Core makes the
+  same distinction in the same file -- `CKey::Sign` ends in
+  `secp256k1_ecdsa_verify`, `CKey::SignCompact` in
+  `secp256k1_ecdsa_recover` and `secp256k1_ec_pubkey_cmp` -- and it
+  subsumes the verification exactly rather than probably. Recovery is
+  not selective: for a given id it answers *the* key under which that
+  `r` and `s` verify, so an inconsistent pair does not fail, it comes
+  back as a different key -- and fails only where `r` is not the x of a
+  point at all, which a faulted `r` is about half the time. Which is the
+  stronger argument: the recovered
+  key is by construction the key that verifies the signature, so
+  `recovered == signer` **is** a verification, with the id checked
+  besides. Nothing is given up by not verifying, and that is provable
+  rather than argued.
+- **What that check costs was measured in a session of its own**, which
+  re-measured `dsa.sign` beside it rather than printing the two
+  together: `recovery.sign` 12.02 against 34.41, `dsa.sign` 12.06
+  against 31.54 there, agreeing with the 12.15 and 31.67 above. So the
+  recovery shape costs 22.4 against ECDSA's 19.5, a recovery being about
+  a verification's work with the comparison and the derivation making up
+  the rest. Same method as the rest, noise +/-0.02.
+- **`_abort_unless_recovered` turns `_recover_`'s `ValueError` into a
+  `RuntimeError`**, which is the one place the two exception kinds of
+  this package meet. "No key can be recovered" is an argument error of a
+  signature a caller handed in, and nothing of the sort for one made a
+  line earlier: what it reports there is the computation having gone
+  wrong, which is what a `RuntimeError` means throughout.
+- **The three do not check quite the same region**, which is worth a
+  sentence and not a change. `ssa` verifies the 64 octets it answers
   with; `dsa` verifies the signature object and serializes it after, so
-  the encoding is outside what was checked -- as it is in Core, for the
-  same reason, the serializers being memcpy-shaped where the signing is
-  arithmetic.
+  the encoding is outside what was checked. `recovery` is the case worth
+  the sentence most: what its check reads is the id inside the
+  `secp256k1_ecdsa_recoverable_signature`, and the id the caller
+  receives is the one `serialize_compact` writes afterwards -- so the id
+  is outside the checked region exactly as the DER is. Core does the
+  same, `CKey::SignCompact` recovering from `rsig` and not from the
+  octets it filled, and for the same reason: the serializers are
+  memcpy-shaped where the signing is arithmetic.
 - **`xonly._from_keypair_` is new**, and is what makes the BIP340 half
   cheap: `xonly.from_keypair` is now that call with a serialize behind
   it, where reaching the key through the public half would have written
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -96,11 +96,18 @@ back where it was:
 +sig = dsa.sign(msg, prvkey, verify=False)
 ```
 
-`recovery.sign` is the one signing call this does not cover: it answers a
-signature nothing here has checked, and it takes no `verify`. The check a
-recoverable signature wants is a different one — recover the key from the
-signature and compare it, which is what Bitcoin Core's `CKey::SignCompact`
-does — so it is a change of its own rather than this one applied twice.
+`recovery.sign` takes the same argument and asks a different question:
+it recovers the public key from the signature and refuses one that is not
+the signer's, rather than verifying. The difference is the recovery id,
+which a verification does not look at — a signature carrying the wrong
+one verifies perfectly and then recovers somebody else's key, which is
+exactly what a caller of that module is going to use it for. It costs
+12.02 microseconds against 34.41 — measured in a session of its own,
+where `dsa.sign` was 12.06 against 31.54, so that the two are compared
+against each other and not across two runs of the machine.
+
+So every signing call of the package now checks itself, and the same
+`verify=False` declines any of them.
 
 What the rest of the release adds: one for a caller that signs more than
 once under one key, one for a caller whose ECDSA signatures go into a
--- a/README.md
+++ b/README.md
@@ -579,12 +579,12 @@ them through `to_compact` writes the DER and parses it straight back.
 
 ## Checking the signature before answering with it
 
-`dsa.sign` and `ssa.sign` verify what they just made, under the public
-key of the very key that made it, and a signature that fails is raised
+`dsa.sign`, `ssa.sign` and `recovery.sign` check what they just made
+against the very key that made it, and a signature that fails is raised
 on rather than returned. It is the one argument here that defaults to
 on, and `verify=False` is how a caller declines it.
 
-The two have the same reasoning behind them and did not arrive by the
+The three have the same reasoning behind them and did not arrive by the
 same route. For BIP340 it is a step of the algorithm: *Default Signing*
 ends with "If Verify(bytes(P), m, sig) returns failure, abort", and the
 note beside it says why — "Verifying the signature before leaving the
@@ -603,6 +603,9 @@ it runs. It catches the computation itself going wrong, whether by bad
 memory or by a fault induced on purpose, and the whole of the protection
 is not publishing the result.
 
+`recovery.sign` asks a different question, and the section below says
+why. What it shares with the other two is the reason and the argument.
+
 It costs what a verification costs, and in ECDSA a point multiplication
 besides — the public key, which BIP340 needs to sign at all and ECDSA
 needs for neither of the two things `sign` does. Measured with the
@@ -618,6 +621,15 @@ has a figure of its own:
 | `ssa.Signer.sign` | 8.18 | 20.82 |
 | `ssa.sign` again (the noise) | | ±0.04 |
 
+and `recovery.sign` in a session of its own, which re-measured `dsa.sign`
+beside it so that the two are comparable rather than merely printed
+together — 12.06 against 31.54 for `dsa.sign` there, agreeing with the
+row above, and a noise row of ±0.02:
+
+| the call | `verify=False` | `verify=True` |
+| --- | --- | --- |
+| `recovery.sign` | 12.02 | 34.41 |
+
 Microseconds per signature. The finding is the gap between the two
 increments: 19.5 for ECDSA against 12.7 for BIP340, and the 6.8 between
 them is the multiplication `secp256k1_ec_pubkey_create` does and
@@ -626,25 +638,46 @@ already, which is the same 7.55 the [signer](#two-outposts-past-the-boundary)
 hoists. So the step the specification prescribes is also the cheaper of
 the two, and a `Signer` pays it at the same price as a bare `ssa.sign`.
 
-**`recovery.sign` has no `verify`, and answers a signature nothing here
-has checked.** It is the one signing entry point this does not cover, and
-saying so is the point: the two sections above read as a policy of the
-package, and a caller reaching for a recoverable signature would
-otherwise conclude it was checked. The reason is that the check is a
-different one. Core closes the same hole in `CKey::SignCompact`, and not
-with a verification: it recovers the public key from the signature and
-compares it with `secp256k1_ec_pubkey_cmp`, which is the natural question
-to ask of a signature carrying a recovery id, and it is its own design,
-its own tests and its own measurement rather than this argument applied
-twice.
+**`recovery.sign` recovers instead of verifying**, and the difference is
+the recovery id. A verification does not look at it: a recoverable
+signature carrying the wrong one verifies perfectly and then recovers a
+key that is not the signer's — and recovering a key is the one thing a
+caller of that module is going to do with the answer. So the check is
+the one the id deserves: recover from the signature, and refuse a key
+that is not the one that signed.
 
-What each half checks is not quite the same region either. `ssa` verifies
-the very 64 octets it answers with; `dsa` verifies the signature *object*
-and serializes it afterwards, so the DER or compact encoding is outside
-what was checked — as it is in Core, whose `CKey::Sign` verifies the
-`secp256k1_ecdsa_signature` and serializes after. Both are defensible,
-the serializers being memcpy-shaped where the signing is arithmetic, and
-the difference is worth a sentence rather than a change.
+That is Bitcoin Core's own distinction, and it makes it in the same
+file. `CKey::Sign` ends in `secp256k1_ecdsa_verify`; `CKey::SignCompact`
+ends in `secp256k1_ecdsa_recover` followed by `secp256k1_ec_pubkey_cmp`
+against the key derived from the private one. Both under the same
+comment.
+
+It subsumes the verification exactly, rather than probably. Recovery is
+not selective: for a given id it answers *the* key under which that `r`
+and `s` verify, so an inconsistent pair does not fail — it comes back as
+a different key, and fails only where `r` is not the x of a point at
+all. That is the stronger argument rather than a weaker one. Because the
+recovered key is by construction the key that verifies the signature,
+`recovered == signer` **is** a verification, with the id checked besides;
+nothing is given up by no longer verifying, and it is provable rather
+than argued. What it costs is 22.4 microseconds against ECDSA's 19.5 — a
+recovery being about a verification's work, and the comparison and the
+derivation making up the rest.
+
+What the three check is not quite the same region either, and
+`recovery`'s is the one worth knowing: what its check reads is the id
+held inside the `secp256k1_ecdsa_recoverable_signature`, while the id the
+caller receives is the one `serialize_compact` writes afterwards — so the
+recovery id is outside the checked region exactly as a DER encoding is.
+Core does the same, `CKey::SignCompact` recovering from its `rsig` and
+not from the octets it filled. `ssa` verifies the very 64 octets it
+answers with; `dsa` verifies the signature *object* and serializes it
+afterwards, so the DER or compact encoding is outside what was checked —
+as it is in Core, whose `CKey::Sign` verifies the
+`secp256k1_ecdsa_signature` and serializes after. All three are
+defensible, the serializers being memcpy-shaped where the signing is
+arithmetic, and the difference is worth a sentence rather than a
+change.
 
 Every other microsecond figure in this file was measured before this
 argument existed and is therefore the signature without the check:
--- a/btclib_secp256k1/recovery.py
+++ b/btclib_secp256k1/recovery.py
@@ -13,15 +13,92 @@ id, which is what `to_der` drops.
 
 from __future__ import annotations
 
-from . import BytesLike, CData, ffi, lib
+from . import BytesLike, CData, ffi, keys, lib
 from ._scalar import in_range, octets, optional_entropy, scalar
 from .context import ctx
 from .dsa import serialize_der
 from .keys import serialize
 
 
+def _abort_unless_recovered(
+    signature: CData, msg_bytes: bytes, prvkey_bytes: bytes
+) -> None:
+    """Recover from a signature just made, and refuse another key.
+
+    What `dsa` and `ssa` do after signing, in the shape a recoverable
+    signature asks for. Bitcoin Core makes the same distinction:
+    `CKey::Sign` ends in `secp256k1_ecdsa_verify`, and `CKey::SignCompact`
+    ends in `secp256k1_ecdsa_recover` followed by
+    `secp256k1_ec_pubkey_cmp` against the key that signed.
+
+    The reason is the recovery id, which is what this signature has
+    beyond a plain one and what a verification does not look at. A
+    signature carrying the wrong id verifies perfectly and recovers
+    somebody else's key -- and recovering a key is the one thing a caller
+    of this module is going to do with it, so the check has to be the one
+    the id answers.
+
+    It subsumes the verification exactly rather than probably, which is
+    what makes not verifying safe here. Recovery is not selective: for a
+    given id it answers *the* key under which that `r` and `s` verify, so
+    an inconsistent pair does not fail, it comes back as a different key
+    -- and fails only where `r` is not the x of a point at all, which is
+    the branch below that reports no key recovering.
+    So the recovered key is by construction the key that verifies the
+    signature, and `recovered == signer` is a verification with the id
+    checked besides.
+
+    Named as `ssa._abort_unless_verified` is, and for the same reason:
+    it raises where a `_foo_` half would answer, and the verb is the one
+    BIP340 uses of the step.
+
+    **Neither call below needs a `context.check` behind it, and both are
+    the kind that usually would.** `keys._pubkey_cmp_` answers an
+    ordering that means nothing where libsecp256k1 could not read an
+    object, and that answer is a security decision here -- but both
+    objects come from calls that returned success a line earlier, so the
+    only way its `0` could lie is both keys being unreadable at once,
+    which is not a state either call can leave behind. And
+    `keys._pubkey_from_prvkey_` raises a `ValueError` for a key outside
+    [1, n-1], which this one is not: libsecp256k1 accepted it for signing
+    immediately above. That is why `_recover_`'s `ValueError` is
+    converted below and this one is not -- the first reports a property
+    of the signature, which a fault can change, and the second a property
+    of an argument that has already been proved. `dsa._sign_` leaves the
+    same call unconverted for the same reason.
+
+    Args:
+        signature: the recoverable signature just made.
+        msg_bytes: the 32-byte hash it was made over, already checked.
+        prvkey_bytes: the 32-byte private key that made it, already
+            checked.
+
+    Raises:
+        RuntimeError: if no key recovers from the signature, or if the
+            one that does is not the signer's. Neither is reachable by
+            any input: what they report is the computation itself having
+            gone wrong.
+    """
+    try:
+        recovered = _recover_(msg_bytes, signature)
+    except ValueError as failure:
+        # `_recover_` reports "no key can be recovered" as a ValueError,
+        # that being an argument error for a signature a caller handed
+        # in. For one made a line ago it is not: nothing was passed that
+        # could have caused it
+        raise RuntimeError("signing produced a signature no key recovers from") from (
+            failure
+        )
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -56,15 +56,32 @@ release-notes length in the first place, and are still in
   signature is the same bytes either way, which
   `tests/test_verified_signing.py` asserts for each of the eight entry
```
fametrano: **Requested: eight is now ten.** `SIGNERS` in `tests/test_verified_signing.py` had eight rows before this branch and has ten after it, `recovery.sign` and `recovery._sign_` being the two added; counted the other way -- the entry points the bullet above names -- the package went from eight to ten as well. So this sentence ships a count of eight for a thing there are ten of, which is the one kind of number CLAUDE.md asks never to write where nothing checks it. Either the number moves, or it goes: "asserts for each of them rather than for one of them" says the same thing and cannot go stale.

CHANGELOG.md
```diff
@@ -56,15 +56,32 @@ release-notes length in the first place, and are still in
...
+  from the signature and refuse one that is not the signer's, rather
+  than verify. The difference is the recovery id, which a verification
+  does not look at: a signature carrying the wrong one verifies
+  perfectly and then recovers somebody else's key, and recovering a key
+  is what a caller of that module does with the answer. Core makes the
+  same distinction in the same file -- `CKey::Sign` ends in
+  `secp256k1_ecdsa_verify`, `CKey::SignCompact` in
+  `secp256k1_ecdsa_recover` and `secp256k1_ec_pubkey_cmp` -- and it
+  subsumes the verification rather than skipping it, a key coming back
+  from `r` and `s` only where those two are consistent.
```
fametrano: **Requested: the reason contradicts the conclusion.** "A key coming back from `r` and `s` only where those two are consistent" reads as recovery being selective, and it is not: `secp256k1_ecdsa_recover` answers, for a given id, the key under which that `(r, s)` verifies, so an inconsistent pair does not fail -- it comes back as a *different* key. It fails only where `r` is not the x of a point at all, which a faulted `r` is about half the time.

The conclusion is right and the same sentence already states the true reason two words later ("a different key or no key at all"). The stronger form of the argument is that the subsumption is exact rather than probable: the recovered key is by construction the key that verifies the signature, so `recovered == signer` *is* a verification, with the id checked besides. Nothing is lost by verifying no longer, and that is provable rather than argued.

Same clause to fix in `README.md:655-656`, `btclib_secp256k1/recovery.py:39-41` and the paragraph in `HISTORY.md`.

README.md
```diff
@@ -626,17 +638,26 @@ already, which is the same 7.55 the [signer](#two-outposts-past-the-boundary)
...
+comment.
+
+It subsumes the verification rather than skipping it: a key comes back
+from `r` and `s` only where those two are consistent, so a fault in
+either gives a different key or no key at all. What it costs is 22.4
+microseconds against ECDSA's 19.5 — a recovery being about a
+verification's work, and the comparison and the derivation making up the
+rest.
 
 What each half checks is not quite the same region either. `ssa` verifies
```
fametrano: **Suggested: three halves.** This paragraph and the matching `CHANGELOG.md:85` bullet still say "each half" and "the two halves" -- and recovery is the case worth the sentence most, because the region left outside its check is the recovery id itself. `_sign_` checks the id held in the `secp256k1_ecdsa_recoverable_signature`, and the id the caller receives is the one `serialize_compact` writes afterwards, in `sign`. Core does exactly the same -- `CKey::SignCompact` recovers from `sig`, not from the octets it filled -- so this is a sentence and not a change, as the paragraph already argues for `dsa`. But said of `dsa`'s DER and not of the id, it reads as if the id were inside the checked region, which is the one thing a reader of this section will want to know.

HISTORY.md
```diff
@@ -96,11 +96,16 @@ back where it was:
...
-recoverable signature wants is a different one — recover the key from the
-signature and compare it, which is what Bitcoin Core's `CKey::SignCompact`
-does — so it is a change of its own rather than this one applied twice.
+`recovery.sign` takes the same argument and asks a different question:
+it recovers the public key from the signature and refuses one that is not
+the signer's, rather than verifying. The difference is the recovery id,
+which a verification does not look at — a signature carrying the wrong
+one verifies perfectly and then recovers somebody else's key, which is
+exactly what a caller of that module is going to use it for. It costs
+12.02 microseconds against 34.41.
```
fametrano: **Suggested: this figure is from another session than the one three paragraphs up.** A reader here has just been given `dsa.sign` 12.15 against 31.67 and now gets 12.02 against 34.41, so the comparison they will make -- 34.41 against 31.67 -- is across sessions, which is the comparison README.md and CHANGELOG.md both went to the trouble of avoiding by re-measuring `dsa.sign` beside it. The clause they carry carries here too: measured in a session of its own, where `dsa.sign` was 12.06 against 31.54. HISTORY.md is the file the downstream reader reads, so it is the one where the unqualified pair does the most harm.

btclib_secp256k1/recovery.py
```diff
@@ -13,15 +13,69 @@
...
+        recovered = _recover_(msg_bytes, signature)
+    except ValueError as failure:
+        # `_recover_` reports "no key can be recovered" as a ValueError,
+        # that being an argument error for a signature a caller handed
+        # in. For one made a line ago it is not: nothing was passed that
+        # could have caused it
+        raise RuntimeError("signing produced a signature no key recovers from") from (
+            failure
+        )
+    if keys._pubkey_cmp_(recovered, keys._pubkey_from_prvkey_(prvkey_bytes)):
```
fametrano: **Suggested: two sentences this line's own neighbours ask for.**

`keys._pubkey_cmp_`'s docstring says that where libsecp256k1 cannot read an object it compares a key of zeros in its place, "so the answer is an ordering like any other and means nothing: this raises nothing, and `context.check` immediately after the call is what says the answer is not to be believed." Here that answer is a security decision, and the fault model of this whole function is exactly the state in which the warning applies. It is fine as written -- both objects come from calls that returned success, and the only way a `0` lies is both keys being unreadable at once -- but this is the first call site in the package to use that comparison as proof, and the docstring above it is the place to say why no `check()` is needed rather than to leave the reader to derive it.

The second is the asymmetry with the paragraph you added: `_recover_`'s `ValueError` becomes a `RuntimeError` because for a signature made a line earlier it reports no argument at all, and by the same argument `keys._pubkey_from_prvkey_` on this very line can raise a `ValueError` for a key libsecp256k1 accepted for signing one line before. `dsa._sign_` has the same shape, so leaving it is defensible -- but the docstring that explains the one conversion should say why the other is not one.

tests/test_verified_signing.py
```diff
@@ -223,3 +288,23 @@ def test_the_keypair_is_checked_against_the_key_that_signed() -> None:
         # not raise, and the signature is held to the key besides
         assert ssa.verify(MSG, pubkey, ssa.sign(MSG, prvkey, AUX))
     assert parities == {0, 1}
+
+
+def test_a_signature_that_recovers_another_key_is_not_answered_with() -> None:
```
fametrano: **Suggested: the headline failure is reachable for real, and this test asserts it instead.** Everything here goes through a stand-in `_recover_`, so what is proved is that the signing path refuses what a recovery reports -- not that a wrong recovery id is what produces that report. The premise of the change ("a signature carrying the wrong id verifies perfectly and then recovers a key that is not the signer's") is tested nowhere, and it is testable with no monkeypatch at all. Run against this branch:

```python
signature, recid = recovery.sign(MSG, PRVKEY)          # recid 1 here
for other in (0, 1, 2, 3):
    parsed = recovery.parse_compact(signature, other)
    recovery._abort_unless_recovered(parsed, MSG, PRVKEY_BYTES)
```

| id | what `_abort_unless_recovered` does |
| --- | --- |
| 0 | `RuntimeError: signing produced a signature that recovers another key` |
| 1 | accepted |
| 2 | `RuntimeError: signing produced a signature no key recovers from` |
| 3 | `RuntimeError: signing produced a signature no key recovers from` |

So one test of a real signature and a real libsecp256k1 recovery reaches both branches, including the `ValueError` -> `RuntimeError` conversion, and it says the thing the stand-ins cannot: that the id is what the check catches. Adding `assert dsa._verify_(MSG, keys._pubkey_from_prvkey_(PRVKEY_BYTES), dsa.parse_compact(signature))` beside it makes the argument of the whole pull request an assertion -- the same octets a verification accepts, refused here.

It also stops depending on `_recover_`'s exception type being restated in `_not_recovering`: today the substituted `ValueError` is a copy of `_recover_`'s contract, and a change there leaves this suite green while the conversion stops firing.

The stand-ins stay either way -- the `SIGNERS` parametrization needs them. This is an addition, not a replacement.

One nit while here: `match="signing produced"` at line 245 now matches all three modules, so that test no longer pins which failure was reported. `_refuse_every_check` already knows the module; the expected message could travel with it.

fametrano (commented):
**Verdict: changes requested** -- stated in words because GitHub declines a formal *Request changes* on a pull request under the same account. Two documentation items below are blocking, four are suggestions, and no code has to move.

Reviewed at 6d9fbdc, in a worktree of its own, and every claim in the description was re-run rather than read:

| what was checked | result |
| --- | --- |
| `uv run --locked --no-default-groups --group test pytest --cov` | 638 passed, 100.00%, CPython 3.14.6 |
| `uvx pre-commit run --all-files` | exit 0, no hook rewrote a line |
| recovery's `if verify:` → `if True:` | 3 failed, 635 passed |
| recovery's `if verify:` → `if False:` | 3 failed, 635 passed |

and the three that die are the right three in each direction: `if True:` loses `test_the_refused_check_is_not_made_at_all[recovery.sign]`, `[recovery._sign_]` and `test_a_signature_that_recovers_another_key_is_not_answered_with`; `if False:` loses `test_a_signature_that_does_not_verify_is_not_answered_with[recovery.sign]`, `[recovery._sign_]` and the same third one. `except ValueError` appears once in the package, so the "one place the two exception kinds meet" is a claim that holds too.

The design is right, and it is right for the reason given: the recovery id is outside what `secp256k1_ecdsa_verify` looks at, so for this module a verification would have been a check that passes while the answer the caller wants is broken. Recovering and comparing is strictly stronger here, it is Core's own distinction, and putting it behind the same keyword-only `verify` is what keeps one policy across the package rather than two.

I am asking for changes on two documentation points and offering four suggestions. Nothing in the code needs to move, and nothing needs re-measuring.

## Changes requested

1. **`CHANGELOG.md:57` states a count this pull request makes wrong.** "asserts for each of the eight entry points" — `SIGNERS` had eight rows before this and has ten now, and read the other way (the entry points named in the bullet above it) the package went from eight to ten as well. Either number or none: the shipped file says eight of a thing there are ten of.
2. **The clause that says no verification is lost gives the wrong reason for a right conclusion**, and it is in four places: `CHANGELOG.md:70-71`, `README.md:655-656`, `btclib_secp256k1/recovery.py:39-41`, and the same argument in `HISTORY.md`. "A key comes back from `r` and `s` only where those two are consistent" is not how recovery behaves: recovery answers *the* key under which that `(r, s)` verifies, so an inconsistent pair does not fail — it comes back as a different key. That is what the next half of the sentence already says, and what the flipped-id probe below shows. The subsumption is exact, and the reason is the stronger one: because the recovered key is by construction the key that verifies the signature, `recovered == signer` **is** a verification, plus the id. As shipped, the sentence contradicts itself in the middle of the load-bearing claim.

## Suggested

3. **The failure this whole change exists for is reachable without a stand-in, and testing it that way would prove what the substitution only asserts** — see the comment on `tests/test_verified_signing.py`. A real signature parsed with the other parity is refused with "recovers another key", and with id 2 or 3 it takes the other branch, `ValueError` → `RuntimeError` and all, through real libsecp256k1.
4. **`README.md:662` and `CHANGELOG.md:85` still say "each half" and "the two halves" of the region question**, now that there are three — and recovery's is the interesting one, the region left outside being the id itself.
5. **`HISTORY.md:105` compares across sessions**, which is the comparison the README went to trouble to avoid.
6. **`recovery.py:69` leans on `keys._pubkey_cmp_` for a security decision**, whose own docstring says what its answer is worth when libsecp256k1 could not read an object.

## Two notes, neither a change to this branch

- **The branch is one commit behind `main`** (2e175b0, "Work out at import what a wrapper was working out per call"), so it wants a rebase before the fast-forward. That commit hoists cdecls in `keys.serialize`, `xonly.serialize` and `silentpayments.serialize_label` and touches nothing the new check calls, so the table survives the rebase.
- **btclib's fast path pays this without asking.** `btclib/ecc/dsa.py:877` calls `libsecp256k1_recovery.sign(msg_hash, q)`, and the comment two lines below it says "nothing here proves the library its own output" — which stops being true the day this ships, at some 22 microseconds a signature. Whether btclib wants `verify=False` there or the guarantee is btclib's call, but that comment becomes wrong and `HISTORY.md` is what its reader reads; a follow-up issue on btclib is the cheap way to not lose it.

fametrano:
Both blocking items are in, all four suggestions taken, and the branch is rebased onto `main`. `6d9fbdc` → `fed52c0`, amended into the one commit.

## Changes requested

**1. The count.** Gone rather than corrected, since ten is as perishable as eight: it now reads "asserts at each entry point rather than at one of them", which is what CLAUDE.md asks for and cannot go stale on the next entry point.

**2. The reason that contradicted the conclusion.** You are right, and it is worse than imprecise — it argues for the conclusion from a property recovery does not have. I checked the behaviour rather than the wording: `secp256k1_ecdsa_recover` answers, for a given id, the key under which that `(r, s)` verifies, so an inconsistent pair does not fail, it comes back as a different key.

Replaced with the exact form in all three places it appeared — `CHANGELOG.md`, `README.md`, `recovery.py`. `HISTORY.md` turned out not to carry the clause at all, so there was nothing to fix there.

> the recovered key is by construction the key that verifies the signature, so `recovered == signer` **is** a verification, with the id checked besides

That is now the sentence, close to your words because they are better than mine: the subsumption is *provable*, not probable.

## Suggested, all four taken

**3. The real failure, without a stand-in.** Your probe reproduces exactly, and I ran it before writing the test:

| id | `_abort_unless_recovered` |
| --- | --- |
| 0 | `RuntimeError: ... recovers another key` |
| 1 | accepted |
| 2, 3 | `RuntimeError: ... no key recovers from` |

`test_the_recovery_id_is_what_the_check_catches` is new and does all of it through real libsecp256k1 — both branches, the `ValueError` → `RuntimeError` conversion included — and ends with the assertion you proposed:

```python
assert dsa._verify_(
    MSG, keys._pubkey_from_prvkey_(PRVKEY_BYTES), dsa.parse_compact(signature)
)
```

which makes the premise of the pull request an assertion instead of a claim in its description: the octets a verification accepts are the octets refused three lines up. Its docstring says why the id table is structural and not lucky — the other parity is a point like any other, `r + n` exceeds the field.

**4. Three, not two.** Both `README.md` and the `CHANGELOG.md` bullet now say three and lead with `recovery`'s, which is the one worth the sentence: what the check reads is the id inside the `secp256k1_ecdsa_recoverable_signature`, and the id the caller receives is the one `serialize_compact` writes afterwards — so the id is outside the checked region exactly as the DER is, and Core recovers from `rsig` rather than from the octets it filled for the same reason.

**5. `HISTORY.md` compared across sessions.** Carries the qualifying clause now, with `dsa.sign` at 12.06 against 31.54 from the same run.

**6. The two sentences at `recovery.py:69`.** Both in the docstring. `keys._pubkey_cmp_`'s warning does not apply because both objects come from calls that returned success a line earlier, so the only way its `0` lies is both keys being unreadable at once — and the asymmetry is now stated rather than left to be derived: `_recover_`'s `ValueError` reports a property of the signature, which a fault can change, while `keys._pubkey_from_prvkey_`'s reports a property of an argument libsecp256k1 accepted for signing immediately above, which is why one is converted and the other is not, `dsa._sign_` leaving the same call alone for the same reason.

**The nit.** The expected message travels with the entry point now: `SIGNERS` rows are `(name, call, refusal)`, so `test_a_signature_that_does_not_verify_is_not_answered_with` pins `does not verify` for the `dsa`/`ssa` rows and `no key recovers from` for the two `recovery` ones, instead of the prefix all three share.

## The two notes

- **Rebased** onto `2e175b0`, one commit still, gate and suite re-run after it: 639 passed, coverage 100.00%, and both mutants re-killed on the rebased tree — 3 failed / 636 passed in each direction.
- **btclib.** Rather than a new issue I put it on [btclib#982](https://github.com/btclib-org/btclib/issues/982), which already opens the same decision for `sign_recoverable_`; a second issue would have split one choice across two threads. The comment names `btclib/ecc/dsa.py:877`, quotes the clause that stops being true — "nothing here proves the library its own output" — and lays out the three ways out, noting that the one thing not to do is leave the comment as it is, since it is the stated reason for skipping a parse.

Thank you for the id probe in particular. The test it produced is the one that says what the change is for, and I had written a suite that could not have failed if the premise were wrong.

btclib_secp256k1/recovery.py
```diff
@@ -13,15 +13,90 @@
...
+    beyond a plain one and what a verification does not look at. A
+    signature carrying the wrong id verifies perfectly and recovers
+    somebody else's key -- and recovering a key is the one thing a caller
+    of this module is going to do with it, so the check has to be the one
+    the id answers.
+
+    It subsumes the verification exactly rather than probably, which is
+    what makes not verifying safe here. Recovery is not selective: for a
+    given id it answers *the* key under which that `r` and `s` verify, so
+    an inconsistent pair does not fail, it comes back as a different key.
```
fametrano: **Nit, not a condition: the qualifier the README has and this does not.** `README.md:658` ends the same sentence with "and fails only where `r` is not the x of a point at all", and that clause is what makes it true: a faulted `r` is not an x-coordinate about half the time, and then nothing is recovered -- which is the branch the `Raises` section below documents and the branch `test_the_recovery_id_is_what_the_check_catches` reaches with ids 2 and 3. As written here and at `CHANGELOG.md:72`, "an inconsistent pair does not fail" is an absolute the same file disproves twelve lines down.

The subsumption does not need it either way -- it rests on a returned key being a key the signature verifies under -- so this is one clause copied from the README, not a rethink.

tests/test_verified_signing.py
```diff
@@ -12,6 +12,13 @@
...
 
+`recovery` asks a different question for the same reason, as Core's
+`CKey::SignCompact` does: it recovers the key from the signature and
+refuses one that is not the signer's, the recovery id being what this
+signature has beyond a plain one and what a verification does not look
+at. So the refusal substituted below is per module rather than one for
+all three.
+
 What can be tested about it is not the fault, which no input produces.
 It is the four things around it: that the check changes no signature,
```
fametrano: **Nit, not a condition: this header is now contradicted by the file's last test.** "What can be tested about it is not the fault, which no input produces. It is the four things around it" was exactly right when every fault needed a substitution. It is not right for `recovery`: the wrong recovery id is a fault an input produces -- one `parse_compact` away -- and `test_the_recovery_id_is_what_the_check_catches` is a fifth thing, the one that says what the check is for rather than that the raise is wired to it.

Worth a sentence here because this header is what a reader of the module reads first, and it currently sends them past the strongest test in the file.

README.md
```diff
@@ -626,23 +638,44 @@ already, which is the same 7.55 the [signer](#two-outposts-past-the-boundary)
...
+held inside the `secp256k1_ecdsa_recoverable_signature`, while the id the
+caller receives is the one `serialize_compact` writes afterwards — so the
+recovery id is outside the checked region exactly as a DER encoding is.
+Core does the same, `CKey::SignCompact` recovering from its `rsig` and
+not from the octets it filled. `ssa` verifies the very 64 octets it
+answers with; `dsa` verifies the signature *object* and serializes it
+afterwards, so the DER or compact encoding is outside what was checked —
+as it is in Core, whose `CKey::Sign` verifies the
+`secp256k1_ecdsa_signature` and serializes after. All three are
+defensible,
```
fametrano: **Nit, not a condition:** `Both are defensible,` became `All three are` + `defensible,` and left an eleven-character line in the middle of a flowed paragraph. Reflowing 676-680 is the whole of it -- no hook reflows prose, which is why it survived the gate.

fametrano (commented):
**ACK** — stated in words because GitHub declines a formal *Approve* on a pull request under the same account. Both blocking items are closed, all four suggestions are in, and nothing below is a condition.

Re-verified at fed52c0 in a fresh worktree, rebuilt from the submodule up, rather than read:

| what was checked | result |
| --- | --- |
| `uv run --locked --no-default-groups --group test pytest --cov` | 639 passed, 100.00% |
| `uvx pre-commit run --all-files` | exit 0 |
| recovery's `if verify:` → `if True:` / `if False:` | 3 failed, 636 passed, each direction |
| base and shape | one commit on 2e175b0, `G Ferdinando M. Ametrano` |

**The count is gone rather than corrected**, which is the better of the two fixes: ten was as perishable as eight.

**The reasoning now says the thing that is true, and the README says it airtight.** "Fails only where `r` is not the x of a point at all" is the clause that makes the sentence hold for a faulted `r` as well as a faulted `s`, and with it the subsumption is a proof rather than an argument.

**The new test earns its place beyond the flag, which I checked by mutating what the flag does not reach.** `if verify:` was never the interesting mutation for a test that calls `_abort_unless_recovered` itself, so I ran the two that are:

| the tree | result |
| --- | --- |
| `if keys._pubkey_cmp_(...)` → `if False:` | 2 failed — `test_a_signature_that_recovers_another_key_...` and `test_the_recovery_id_is_what_the_check_catches` |
| the `ValueError` → `RuntimeError` conversion removed | 3 failed, and `test_the_recovery_id_is_what_the_check_catches` is the only one of them that catches it without a stand-in |

So the conversion, which the description argues for at length, is now held by a test that reaches it through real libsecp256k1 — and the id table in that test's docstring is structural, `r + n` exceeding the field for any `r` a signature carries.

## Three lines of prose, none of them a condition

1. **`recovery.py:44` and `CHANGELOG.md:72` are missing the qualifier `README.md:658` carries.** All three say an inconsistent pair "does not fail, it comes back as a different key"; only the README adds "and fails only where `r` is not the x of a point at all". Without it the sentence is false for about half of faulted `r` values — and false about the very branch the docstring's own `Raises` documents and the new test exercises with ids 2 and 3. Two files now say what the third corrects.
2. **`tests/test_verified_signing.py:22-23` is contradicted by its own last test.** "What can be tested about it is not the fault, which no input produces. It is the four things around it" was true when every fault needed a stand-in. For `recovery` an input does produce it — the wrong id is one `parse_compact` away — and that is the fifth thing the file now tests, and the only one that says what the change is for.
3. **`README.md:678`** leaves `defensible,` alone on a line; the paragraph wants reflowing across 676-680.

Land it. The three above fold into the same commit if you want them, and none is worth a second round.

tests/test_verified_signing.py
```diff
@@ -171,17 +281,18 @@ def test_a_signature_that_does_not_verify_is_not_answered_with(
     Args:
         name: the entry point, for the test id.
         signer: it, as a call taking the keyword under test.
+        refusal: what a refused check reports there, unused where the
```
fametrano: **This is the one test where `refusal` is used, and its docstring says it may be unused.** Line 289 is `match=refusal`; the line here reads "unused where the test does not refuse one", which is true of the three other consumers of `SIGNERS` (197, 219, 294) and false of exactly this one. The sentence was copied to all four. Whatever happens to the shape below, this line wants to say what `refusal` is *for* here -- the message this entry point refuses with, `does not verify` for `dsa` and `ssa`, `no key recovers from` for the two `recovery` rows.

**And the shape, which is taste and which I left out of the ACK.** Four tests are parametrized over `SIGNERS`, one uses the third element, and the three that do not each carry two lines of docstring explaining that they do not. The alternative is not to carry it: the refusal is a property of the module, and the module is already in the name --

```python
_REFUSALS = {"dsa": _UNVERIFIED, "ssa": _UNVERIFIED, "Signer": _UNVERIFIED, "recovery": _UNRECOVERED}
...
    with pytest.raises(RuntimeError, match=_REFUSALS[name.split(".")[0]]):
```

which puts `SIGNERS` back to pairs, leaves one test reading the mapping, and deletes eight docstring lines about an argument nothing uses. Against it: the row is then not the whole truth about the entry point, and one table saying everything is the argument the current shape makes. Your call -- I would not hold the branch for either.

CHANGELOG.md
```diff
@@ -54,23 +54,53 @@ release-notes length in the first place, and are still in
...
+  is what a caller of that module does with the answer. Core makes the
+  same distinction in the same file -- `CKey::Sign` ends in
+  `secp256k1_ecdsa_verify`, `CKey::SignCompact` in
+  `secp256k1_ecdsa_recover` and `secp256k1_ec_pubkey_cmp` -- and it
+  subsumes the verification exactly rather than probably. Recovery is
+  not selective: for a given id it answers *the* key under which that
+  `r` and `s` verify, so an inconsistent pair does not fail, it comes
+  back as a different key -- and fails only where `r` is not the x of a
+  point at all, which a faulted `r` is about half the time. Which is the
+  stronger argument: the recovered
```
fametrano: **The same orphan the README fix removed, introduced here by the same insertion.** "Which is the" now ends line 74 and "stronger argument: the recovered" is a 34-character line in the middle of a flowed paragraph, with 69-to-72-character lines either side. Reflowing 70-79 is the whole of it; no hook reflows prose, which is why this one and the README's both passed the gate.

fametrano (commented):
The three prose nits are in at bea6a04, and the interdiff from fed52c0 touches docstrings and markdown only -- no statement changed -- so the run I reported on fed52c0 carries: 639 passed, 100.00%, gate clean, both mutants dead. `ruff format --diff` on the two Python files at this head says both are already formatted. **The ACK stands.**

Two more of the same kind, raised because the first was asked for and the second is its twin. Neither is a condition, and one of them is a wrong sentence rather than a preference.

fametrano:
Tracked post-merge, the two review notes above plus one they turned up: #225 carries the prose the tail of this change left wrong — `CHANGELOG.md` still opens a bullet with "The fault itself is not tested, because no input produces one", which `test_the_recovery_id_is_what_the_check_catches" makes false, and the same file's header was corrected here for exactly that. #226 carries the `refusal` shape as a question, since it is a decision rather than a defect.

PR #220, opened 2026-08-17
Let the caller of a keypair ask for the parity it reads

### LINKED ISSUE #219: The parity `ssa` throws away is still allocated
`xonly._from_keypair_` allocates a parity for a caller that discards it,
which is the shape
[PR 215](https://github.com/btclib-org/btclib-secp256k1/pull/215) has
just removed from `xonly._drop_y` and which arrived in `main` from the
other direction while that branch was in review.

## What is there

[`btclib_secp256k1/xonly.py:158`](https://github.com/btclib-org/btclib-secp256k1/blob/main/btclib_secp256k1/xonly.py#L158)
allocates the parity unconditionally:

```python
def _from_keypair_(keypair_obj: CData) -> tuple[CData, int]:
    xonly_pubkey = ffi.new("secp256k1_xonly_pubkey *")
    parity = ffi.new("int *")
    converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, parity, keypair_obj)
```

It has two callers, and one of them throws that parity away —
[`btclib_secp256k1/ssa.py:598`](https://github.com/btclib-org/btclib-secp256k1/blob/main/btclib_secp256k1/ssa.py#L598),
inside `_abort_unless_verified`:

```python
xonly_pubkey, _ = xonly._from_keypair_(keypair_obj)
```

The other, `xonly.from_keypair`, reads it and returns it.

libsecp256k1 says the pointer is optional, in the same words it uses for
`secp256k1_xonly_pubkey_from_pubkey` — `secp256k1_extrakeys.h`, on
`secp256k1_keypair_xonly_pub`:

```
 *    pk_parity: Ignored if NULL. Otherwise, pointer to an integer that will be set to the
 *               pk_parity argument of secp256k1_xonly_pubkey_from_pubkey.
```

So the same treatment applies: the parameter defaults to `ffi.NULL`, the
caller that reads a parity allocates one, and the caller that does not
passes nothing.

## What it is worth, which is not much

Saying so up front, because the `_drop_y` figure does not carry over.
There the allocation was a sixth of a `xonly.parse` — 0.4295 microseconds
against 0.5021 in PR 215's session. Here it sits on the BIP340 signing
path: `_abort_unless_verified` runs by default from both `ssa.sign` and
`ssa.sign_custom`, and what it does is a whole verification. An
`ffi.new` of a literal cdecl is about 0.07 microseconds by that same
session's primitives table; a sign with the check is some tens of
microseconds of libsecp256k1. The ratio is a fraction of a percent, and
nobody will see it.

The case is therefore consistency and not speed: the package would have
one convention for an optional out-parameter rather than two of them, one
per site, written the same week. If that is not
worth the change, the honest outcome is a comment on `_from_keypair_`
saying the parity is allocated because both spellings cost the same here
— a measured negative result, as
[CHANGELOG.md](https://github.com/btclib-org/btclib-secp256k1/blob/main/CHANGELOG.md)
records the others.

## If it is taken

- `_from_keypair_(keypair_obj, parity=ffi.NULL)`, matching `_drop_y`
- `from_keypair` allocates its own, the way `_drop_y_with_parity` does
- `ssa._abort_unless_verified` drops the `, _`
- one measurement on `ssa.sign` with the check on, against a noise row,
  and the figure goes wherever it lands — including in a declining
  comment if it lands in the noise

The `_from_keypair_` half arrived in
[`e8902ee`](https://github.com/btclib-org/btclib-secp256k1/commit/e8902ee),
`_drop_y`'s in
[PR 215](https://github.com/btclib-org/btclib-secp256k1/pull/215) for
[#212](https://github.com/btclib-org/btclib-secp256k1/issues/212).

fametrano:
The other half of #212's first bullet, arriving from the direction
`e8902ee` came from.

**Stacked on #215**, whose `xonly.py` this touches: the base is
`cheaper-buffers` at `2e175b0`, so the diff here is this change alone.
Retarget to `main` once #215 fast-forwards — its sha survives that merge,
which is what leaves this branch applying with nothing to redo.

## What changed

- `xonly._from_keypair_(keypair_obj, parity=ffi.NULL)`, matching
  `_drop_y`. It answers the key alone now.
- `xonly.from_keypair` allocates the `int *` it reads. At the call site
  rather than behind a helper: `_drop_y` needed `_drop_y_with_parity`
  because two of its callers read a parity, and this has one.
- `ssa._abort_unless_verified` drops the `, _`.

libsecp256k1 documents the pointer as optional in the same words for
both calls — `secp256k1_extrakeys.h`, on `secp256k1_keypair_xonly_pub`:

```
 *    pk_parity: Ignored if NULL. Otherwise, pointer to an integer that will be set to the
 *               pk_parity argument of secp256k1_xonly_pubkey_from_pubkey.
```

## The measurement, and the half of it that says nothing

The issue asked for one on `ssa.sign` with the check on, against a noise
row, and said the figure goes wherever it lands. It landed in two
places. Two spellings alternated in one process, 15 rounds, an Apple M5,
macOS 26.6, arm64, CPython 3.14.6:

| | with the parity | with NULL |
| --- | --- | --- |
| the conversion alone | 0.2384 | 0.1718 |
| `ssa.sign` with the check on, which is where it runs | 19.9400 | 20.0303 |
| *noise*, the conversion twice | 0.2353 | 0.2404 |

**−28.0% on the call, +0.45% on the path** — the second against a noise
row that moved 2.18%, which is to say nothing, and the wrong way at
that. An `ffi.new` is some hundredths of a microsecond; a signature with
its verification is twenty.

So this is taken **for the convention and not for the speed**, which is
the outcome the issue put first, and the negative half is in CHANGELOG.md
rather than dropped — it is the same sentence that would have gone into
a declining comment had the first row landed in the noise too.

## What judges it

`tests/test_parsed_keys.py` gains
`test_the_keypair_parity_is_the_callers_to_ask_for`: the key is the same
whether a parity is asked for or not, and the parity is the one
`from_keypair` answers. That is "Ignored if NULL" held to rather than
quoted. The pair test beside it moves to the new signature and says why
in its docstring.

628 passed, coverage gate at 100.00%, every hook passes.

Closes #219

## Summary by Sourcery

Make xonly keypair conversion treat the parity out-parameter as an optional caller-owned pointer, aligning internal helpers and SSA verification with libsecp256k1’s convention.

Enhancements:
- Change `xonly._from_keypair_` to return only the x-only public key and accept an optional parity pointer, with `from_keypair` allocating and reading parity at its call site.
- Update SSA verification to call `xonly._from_keypair_` without a parity pointer, reflecting that it does not use the parity value.
- Document the new out-parameter convention and its performance implications in CHANGELOG and HISTORY, emphasizing consistency with libsecp256k1.

Tests:
- Add tests ensuring the keypair parity is requested explicitly by callers and that key output is identical whether or not parity is read, with `from_keypair` remaining the source of parity information.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -317,6 +317,16 @@ it, and does afterwards. Reading it — as the attribute, as
 metadata — answers what it always answered, and costs what it always
 cost, the first read and every one after it.
 
+**On both conversions, the parity is the caller's to ask for.**
+`xonly._drop_y` and `xonly._from_keypair_` take an `int *` where one is
+wanted, and the NULL libsecp256k1 documents as "Ignored" where it is
+not: `from_pubkey`, `to_pubkey` and `from_keypair` allocate one, and
+`parse`, `_parsed`, `_tweak_add_` and the check `ssa` makes of its own
+signature pass nothing. Both conversions are new in this release, so
+there is nothing here to act on — what a caller of the octets API sees
+is unchanged, and so is what it costs, a signature with its verification
+being 20 microseconds either way.
+
 **Serializing and parsing cost a little less, and nothing changed about
 what they answer.** Four things a wrapper used to work out at every call
 are worked out once: the cffi type of a buffer whose declaration was
--- a/btclib_secp256k1/xonly.py
+++ b/btclib_secp256k1/xonly.py
@@ -72,7 +72,7 @@ def _drop_y(pubkey: CData, parity: CData = ffi.NULL) -> CData:
             reading it, which is the default because the callers that
             want the object alone are why it exists: `parse`, `_parsed`
             and `_tweak_add_`. libsecp256k1 documents `pk_parity` as
-            "can be NULL", so the allocation is the caller's to make
+            "Ignored if NULL", so the allocation is the caller's to make
             rather than this function's, and skipping it is 0.4295
             microseconds against 0.5021 on `xonly.parse` of a 65-byte
             key -- CHANGELOG.md names the session every figure in this
@@ -207,7 +207,7 @@ def from_prvkey(prvkey: BytesLike | int) -> tuple[bytes, int]:
     return _from_pubkey_(keys._pubkey_from_prvkey_(prvkey))
 
 
-def _from_keypair_(keypair_obj: CData) -> tuple[CData, int]:
+def _from_keypair_(keypair_obj: CData, parity: CData = ffi.NULL) -> CData:
     """Return the x-only public key of a keypair, as the parsed point.
 
     The private half of `from_keypair`, for a caller about to hand the
@@ -221,11 +221,19 @@ def _from_keypair_(keypair_obj: CData) -> tuple[CData, int]:
     Args:
         keypair_obj: the libsecp256k1 keypair object, as `ssa.Signer`
             holds and as `secp256k1_keypair_create` writes.
+        parity: an `int *` to receive the parity of y, 0 for even and 1
+            for odd -- or the NULL that says nobody is reading it, which
+            is the default because `ssa._abort_unless_verified` is in
+            that case and `from_keypair` is the one that is not.
+            libsecp256k1 documents `pk_parity` as "Ignored if NULL", the
+            same words it gives `secp256k1_xonly_pubkey_from_pubkey`,
+            which is why `_drop_y` above takes its pointer the same way.
 
     Returns:
-        The libsecp256k1 x-only public key object, and the parity of y:
-        0 for even, 1 for odd. The parity is of the point the private
-        key gives, the keypair being the negated key where that y is odd.
+        The libsecp256k1 x-only public key object. The parity written
+        through the pointer, where one is given, is of the point the
+        private key gives, the keypair being the negated key where that
+        y is odd.
 
     Raises:
         RuntimeError: if libsecp256k1 will not read the object -- a NULL
@@ -235,11 +243,10 @@ def _from_keypair_(keypair_obj: CData) -> tuple[CData, int]:
             deserves, and `context.check` is what tells the two apart.
     """
     xonly_pubkey = ffi.new("secp256k1_xonly_pubkey *")
-    parity = ffi.new("int *")
     converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, parity, keypair_obj)
     if not converted:
         raise RuntimeError("x-only public key conversion failed")
-    return xonly_pubkey, parity[0]
+    return xonly_pubkey
 
 
 def from_keypair(keypair_obj: CData) -> tuple[bytes, int]:
@@ -274,8 +281,11 @@ def from_keypair(keypair_obj: CData) -> tuple[bytes, int]:
             fails to serialize the result, which a keypair it built
             cannot make it do.
     """
-    xonly_pubkey, parity = _from_keypair_(keypair_obj)
-    return serialize(xonly_pubkey), parity
+    # the one caller of the two that reads a parity, so the allocation
+    # is here rather than behind a helper: `_drop_y` needed
+    # `_drop_y_with_parity` because it has two
+    parity = ffi.new("int *")
+    return serialize(_from_keypair_(keypair_obj, parity)), parity[0]
 
 
 def _tweak_add_(pubkey: CData, tweak: BytesLike | int) -> tuple[bytes, int]:
--- a/tests/test_parsed_keys.py
+++ b/tests/test_parsed_keys.py
@@ -43,6 +43,9 @@ from btclib_secp256k1._secret import keypair, wipe
 N = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
 
 PRVKEY = 7
+# the small scalar whose point has odd y, as tests/test_nonces.py keeps
+# it: a parity test over an even-y key alone compares zeros with zeros
+ODD_Y_PRVKEY = 6
 TWEAK = 11
 SCAN_PRVKEY = 13
 MSG = hashlib.sha256(b"btclib_secp256k1").digest()
@@ -354,15 +357,51 @@ def test_the_keypair_pair_reads_the_point_it_already_holds() -> None:
     caller could have handed in. What the private half saves is that
     serialize and the lift back that reading it again would cost, which
     is what `ssa.sign(verify=True)` takes it for.
+
+    The parity is asked for through the pointer, which is the half of
+    the pair a NULL default leaves to the caller: passing none is what
+    `ssa._abort_unless_verified` does, and the test below is where that
+    spelling answers alike.
     """
     keypair_obj = keypair(PRVKEY)
     try:
-        xonly_pubkey, parity = xonly._from_keypair_(keypair_obj)
-        assert (xonly.serialize(xonly_pubkey), parity) == xonly.from_keypair(
+        parity = ffi.new("int *")
+        xonly_pubkey = xonly._from_keypair_(keypair_obj, parity)
+        assert (xonly.serialize(xonly_pubkey), parity[0]) == xonly.from_keypair(
             keypair_obj
         )
     finally:
         wipe(keypair_obj)
+
+
+def test_the_keypair_parity_is_the_callers_to_ask_for() -> None:
+    """`xonly._from_keypair_` answers the same key with a parity and without.
+
+    The NULL default is what `ssa._abort_unless_verified` takes, so the
+    key it verifies against has to be the key `from_keypair` names --
+    libsecp256k1 documenting `pk_parity` as "Ignored if NULL" is the
+    claim, and this is it held to.
+
+    Both parities, and the 0 and the 1 asserted rather than only the
+    agreement between the two calls: `ffi.new("int *")` zeroes the
+    buffer it allocates, so over an even-y key alone every side of every
+    comparison here is 0 whether the pointer was written through or not.
+    Discarding the caller's pointer inside the helper is then a mutant
+    the whole suite passes, while `from_keypair` answers 0 for every
+    odd-y key. 6 is the odd-y scalar `tests/test_nonces.py` already
+    keeps for this, under the name it gives it there.
+    """
+    for prvkey, expected in ((PRVKEY, 0), (ODD_Y_PRVKEY, 1)):
+        keypair_obj = keypair(prvkey)
+        try:
+            parity = ffi.new("int *")
+            with_parity = xonly._from_keypair_(keypair_obj, parity)
+            without = xonly._from_keypair_(keypair_obj)
+            assert xonly.serialize(with_parity) == xonly.serialize(without)
+            assert parity[0] == expected
+            assert parity[0] == xonly.from_keypair(keypair_obj)[1]
+        finally:
+            wipe(keypair_obj)
 
 
 def test_the_taproot_pair_starts_from_the_point_the_x_belongs_to() -> None:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -340,6 +340,46 @@ release-notes length in the first place, and are still in
   Nothing else changes: the attribute, the `from … import __version__`,
   and reading the metadata directly all answer what they answered.
 
+### The out-parameter that is the caller's to ask for
+
+- **`xonly._from_keypair_` takes its parity pointer, defaulting to NULL**
+  (#219), which is the shape the entry below gives `xonly._drop_y` and
+  which libsecp256k1 documents in the same words for both:
+  `secp256k1_extrakeys.h` says `pk_parity: Ignored if NULL` of
+  `secp256k1_keypair_xonly_pub` as it does of
+  `secp256k1_xonly_pubkey_from_pubkey`. `from_keypair` allocates one and
+  reads it; `ssa._abort_unless_verified`, which threw it away, passes
+  nothing. The allocation sits at the one call site that wants it rather
+  than behind a helper — `_drop_y` needed `_drop_y_with_parity` because
+  it has two such callers, and this has one.
+- **It is worth 28% of the conversion and nothing where the conversion
+  is**, and both halves of that are the point of taking it. Same session
+  shape as the entry below — the two spellings alternated in one process
+  over 15 rounds, a noise row beside each, an Apple M5, macOS 26.6,
+  arm64, CPython 3.14.6:
+
+  | | with the parity | with NULL |
+  | --- | --- | --- |
+  | the conversion alone | 0.2384 | 0.1718 |
+  | `ssa.sign` with the check on, which is where it runs | 19.9400 | 20.0303 |
+  | *noise*, the conversion twice | 0.2353 | 0.2404 |
+
+  −28.0% on the call, +0.45% on the path against a noise row that moved
+  2.18% — which is to say nothing, an `ffi.new` being some hundredths of
+  a microsecond against a signature and a verification. **So this is
+  taken for the convention and not for the speed**: the package had two
+  answers to one question, written a week apart, and now has one.
+- **One test holds it, and it takes two keys to.**
+  `test_the_keypair_parity_is_the_callers_to_ask_for` runs over an even-y
+  key and an odd-y one and asserts the 0 and the 1, because
+  `ffi.new("int *")` zeroes what it allocates: over an even-y key alone
+  every side of every comparison is 0 whether the pointer was written
+  through or not, and discarding it inside the helper is then a mutant
+  the whole suite passes while `xonly.from_keypair` answers 0 for every
+  odd-y key. With both keys that mutant fails at the assertion meant to
+  catch it. The same mutation on `_drop_y` fails eight tests, so the
+  claim these two conversions now share is one the suite holds on both.
+
 ### What a wrapper works out per call, and now works out once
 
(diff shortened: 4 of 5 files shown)


tests/test_parsed_keys.py
```diff
@@ -354,17 +354,42 @@ def test_the_keypair_pair_reads_the_point_it_already_holds() -> None:
...
+    libsecp256k1 documenting `pk_parity` as "Ignored if NULL" is the
+    claim, and this is it held to.
+    """
+    keypair_obj = keypair(PRVKEY)
+    try:
+        parity = ffi.new("int *")
+        with_parity = xonly._from_keypair_(keypair_obj, parity)
+        without = xonly._from_keypair_(keypair_obj)
+        assert xonly.serialize(with_parity) == xonly.serialize(without)
+        assert parity[0] == xonly.from_keypair(keypair_obj)[1]
```
fametrano: **Required.** `0 == 0` on both sides of this, for a key of even y and a
zero-initialised buffer, so it holds whether or not the pointer was
written through. See the review body for the mutant that survives all
628 tests because of it.

What closes it, and what I ran to check that it does:

```python
    for prvkey, expected in ((PRVKEY, 0), (ODD_Y_PRVKEY, 1)):
        keypair_obj = keypair(prvkey)
        try:
            parity = ffi.new("int *")
            with_parity = xonly._from_keypair_(keypair_obj, parity)
            without = xonly._from_keypair_(keypair_obj)
            assert xonly.serialize(with_parity) == xonly.serialize(without)
            assert parity[0] == expected
            assert parity[0] == xonly.from_keypair(keypair_obj)[1]
        finally:
            wipe(keypair_obj)
```

with `ODD_Y_PRVKEY = 6` beside `PRVKEY`, which is the scalar
`tests/test_nonces.py` already calls `ODD_Y`, and its comment there says
why 6 and not another. `assert parity[0] == expected` is the line the
mutant fails on; the agreement between the two calls alone does not
catch it, both being 0.

The test above, `test_the_keypair_pair_reads_the_point_it_already_holds`,
is the same shape and equally blind on the parity half. It reads as the
pair equality it is about, so it need not grow the loop — but with this
one exercising both parities, the docstring cross-reference between them
becomes true of the parity as well as of the key.

btclib_secp256k1/xonly.py
```diff
@@ -221,11 +221,19 @@ def _from_keypair_(keypair_obj: CData) -> tuple[CData, int]:
     Args:
         keypair_obj: the libsecp256k1 keypair object, as `ssa.Signer`
             holds and as `secp256k1_keypair_create` writes.
+        parity: an `int *` to receive the parity of y, 0 for even and 1
+            for odd -- or the NULL that says nobody is reading it, which
+            is the default because `ssa._abort_unless_verified` is in
+            that case and `from_keypair` is the one that is not.
+            libsecp256k1 documents `pk_parity` as "Ignored if NULL", the
```
fametrano: **Suggested.** This quotes the header correctly, and `_drop_y` — 150
lines above, and the function this argues symmetry with — quotes the same
sentence as `"can be NULL"`. The vendored header, at the `6e2c8bc` this
tree pins, says `Ignored if NULL` for `pk_parity` on both extrakeys
functions; `can be NULL` is libsecp256k1's phrasing for other arguments
(`secp256k1_schnorrsig.h` on `extraparams`, `secp256k1_ecdh.h` on its
hash function, `secp256k1_musig.h` in several places) and not for this
one.

So the module states two different quotations of one sentence, and the
claim made here — *the same words* — is what makes them worth
reconciling: one word in `_drop_y`'s `Args`, and the same word in the
`_drop_y` entry of CHANGELOG.md, which is in the unreleased v0.8.0.3
section and still editable. Then the symmetry this docstring asserts can
be read off the two docstrings rather than off the header.

HISTORY.md
```diff
@@ -310,6 +310,16 @@ it, and does afterwards. Reading it — as the attribute, as
 metadata — answers what it always answered, and costs what it always
 cost, the first read and every one after it.
 
+**One convention for an optional out-parameter, where there were two.**
```
fametrano: **Suggested.** "where there were two", and "It is the private half's
signature that changed", are true of the tree and not of anything a
reader of these notes can have installed: `xonly._from_keypair_` is new
in v0.8.0.3 — CHANGELOG.md says so, under *A signer checks its own
signature* — so no released version ever had the two conventions, and no
caller ever had the signature that changed.

CHANGELOG.md is where the change is the story, and it tells it there.
Here the release reads as what it ships: on both conversions the parity
is the caller's to ask for, `ssa` does not, and nothing at the octets API
moves. Non-blocking, and a judgement call about which of the two files
carries the churn.

fametrano (commented):
## Changes requested — one of them, and it is a test

*Posted as a comment: GitHub will not take a formal request-changes from
the author of the pull request, so the verdict is in words. **Not an ACK
until the required item below is addressed.***

The source change is right, and I checked its premise rather than took
it: at the pin this tree vendors, `6e2c8bc`,
`secp256k1/include/secp256k1_extrakeys.h` gives `pk_parity` of
`secp256k1_keypair_xonly_pub` and of `secp256k1_xonly_pubkey_from_pubkey`
the same sentence, "Ignored if NULL", so the pointer is an argument here
exactly as it is on `_drop_y`. 628 pass on `d68381b`, and the `ffi.NULL`
default crosses cffi as the `int *` that call wants.

Taking it for the convention and publishing the half that says nothing is
the right call, and the two sessions agree where they can be compared:
the conversion alone is 0.2384 against 0.1718 here and 0.2425 against
0.1747 on `_drop_y`, a week apart, which is worth more than either figure
alone.

### Required — the new test cannot see the parity being written

`PRVKEY = 7` has even y, so `from_keypair` answers 0, and
`ffi.new("int *")` zero-initialises. Both sides of
`assert parity[0] == xonly.from_keypair(keypair_obj)[1]` are therefore 0
whether or not anything was written through the pointer, and the claim
the PR puts first — "the parity is the one `from_keypair` answers" — is
held by nothing.

That is not an argument, it is a mutant that lives. Discarding the
caller's pointer inside the helper:

```diff
-    converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, parity, keypair_obj)
+    converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, ffi.NULL, keypair_obj)
```

leaves all 628 tests passing, with `xonly.from_keypair` then answering 0
for every odd-y key — a public entry point wrong about half its inputs,
and the suite silent. The coverage gate cannot say so and is not meant
to: the line runs, and what it wrote is never read.

The release notes of this very version already name the failure mode, in
the `_from_keypair_` neighbourhood: *"the keys of both y parities are
signed with as well, both parities asserted to occur, because a check
read off the wrong half of the negation BIP340 prescribes would pass for
the wrong reason"*. This is that sentence applied to the parity itself.

One key closes it — 6 is the small scalar whose y is odd, and
`tests/test_nonces.py` already names it `ODD_Y` for this reason. With the
new test run over an even-y key and an odd-y one, and the parity asserted
against the 0 and the 1 rather than only against the other call's answer,
the mutant above fails at that assertion. Verified both ways, not
reasoned about. The neighbour in the same file already works like this:
`test_every_serialization_of_a_key_is_the_same_x_only_key` pins
`from_pubkey(negated)[1] == 1` precisely so that a parity nobody wrote
cannot pass for one that was.

### Suggested — two, both documentation, neither blocking

Detail in the inline comments: the same header sentence is quoted two
ways in one module, and HISTORY.md narrates a change to something that
has not shipped.

### Housekeeping, and nothing to do about it

GitHub retargeted this to `main` when `cheaper-buffers` was deleted —
base oid `2e175b0` is `main`'s tip — so the branch is one signed commit
over `main` and the fast-forward is available. Every check is green bar
the two static-wheel matrix jobs, still running when I read them.

fametrano:
Pushed at 27021e7. The required item was a real defect and I reproduced
it before fixing it; both suggestions taken.

### The test held nothing, and the mutant proved it

Ran your diff on `d68381b`:

```
628 passed
from_keypair(6) parity = 0   (should be 1)
```

All 628 green with `xonly.from_keypair` wrong about every odd-y key. You
are right about why: `ffi.new("int *")` zeroes what it allocates and
`PRVKEY = 7` has even y, so both sides of every comparison were 0
whether the pointer was written through or not. I ran the suite when I
added that test but never handed it something bad — which is the one
thing this repository says to do when adding a check, and the reason
`test_every_serialization_of_a_key_is_the_same_x_only_key` pins the 1
next door.

Fixed as you spelled it: `ODD_Y_PRVKEY = 6` beside `PRVKEY`, the test
looping over `((PRVKEY, 0), (ODD_Y_PRVKEY, 1))`, and `parity[0] ==
expected` as the assertion that carries it. Re-ran the mutant:

```
tests/test_parsed_keys.py:401: AssertionError
FAILED tests/test_parsed_keys.py::test_the_keypair_parity_is_the_callers_to_ask_for
1 failed, 627 passed
```

and 628 pass with the file restored. The docstring says why 6 and why
one key would not do, so the next reader does not re-derive it.

I left `test_the_keypair_pair_reads_the_point_it_already_holds` as the
pair equality it is about, as you suggested.

### Suggested — both in

**The quotation.** You are right that the header says `Ignored if NULL`
for `pk_parity` on both extrakeys functions, and that `can be NULL` is
libsecp256k1's phrasing elsewhere — `secp256k1_ecdh.h`,
`secp256k1_schnorrsig.h`, `secp256k1_recovery.h`, all for other
arguments. `_drop_y`'s `Args` and its CHANGELOG entry now quote what the
header says, so the symmetry this docstring asserts reads off the two
docstrings.

**HISTORY.md.** Also right, and it is the kind of thing only a reader of
the notes would catch: `_from_keypair_` is new in v0.8.0.3, so no
released version had two conventions and no caller had a signature that
changed. It now says what the release ships — on both conversions the
parity is the caller's to ask for, and there is nothing to act on — and
CHANGELOG.md keeps the story of the change.

628 passed, coverage 100.00%, every hook passes.

fametrano (commented):
## ACK `27021e7`

*Explicitly: this is an ACK, posted as a comment because GitHub will not
take a formal approval from the author of the pull request. Nothing
remains that I would hold the merge for.*

All three items of the previous round are addressed, and I re-ran what I
ran to raise them rather than reading the diff and believing it.

### The required one, held now

The test runs over an even-y key and an odd-y one and asserts the 0 and
the 1, so the mutant that survived every one of the 628 tests on
`d68381b` now dies at exactly the assertion meant to catch it:

```diff
-    converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, parity, keypair_obj)
+    converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, ffi.NULL, keypair_obj)
```

`1 failed, 627 passed`, the failure being
`test_the_keypair_parity_is_the_callers_to_ask_for`. Unmutated: 628
passed, coverage 100.00%.

And I checked the other half of the symmetry the release notes now claim,
which the previous round did not: the same mutation on `_drop_y` — the
caller's pointer discarded there — fails 8 tests, `test_nonces.py`'s
`[odd y]` case and
`test_every_serialization_of_a_key_is_the_same_x_only_key` among them. So
"on both conversions the parity is the caller's to ask for" is now a
sentence the suite holds on both conversions, one of them by exactly one
test.

### The two suggested ones

The header sentence is quoted one way in the module now, `_drop_y` and
`_from_keypair_` alike, and the `_drop_y` CHANGELOG entry with them, so
the symmetry the docstring asserts can be read off the two docstrings.
And HISTORY.md states the convention the release ships rather than a
change to a helper that never shipped — "Both conversions are new in this
release, so there is nothing here to act on" is the sentence that was
missing, and it belongs in the file whose reader has the release and not
the branch.

I considered and dropped one objection to the fix: `ODD_Y_PRVKEY = 6`
beside `tests/test_nonces.py`'s `ODD_Y = 6` looks like one thing said
twice, and it is not — 7 is already spelled as `PRVKEY` in four test
modules and as `EVEN_Y` in a fifth, so a scalar named for what it
exhibits, in the module that exhibits it, is this suite's convention.
Sharing it would couple two test modules to buy nothing.

### Optional, and the ACK does not wait on it

The CHANGELOG entry does not name what judges the change, where the
commit message now explains it at length. Elsewhere in the same release
the notes do name it — *"Two tests hold the saving"*, *"a test of its own
holds the equality"* — and the entry's own claim, that one question has
one answer now, is what the both-parities assertion is holding. A clause
would carry it. Equally, the commit message is where the mutant belongs
and it is there, so leaving CHANGELOG.md as it stands is a defensible
reading of the same convention.

### Merge posture

One commit, signed by Ferdinando M. Ametrano and verifying `G`; base is
`main` at `2e175b0`, which is `main`'s tip, so the fast-forward is
available unsquashed. Lint and type-check, coverage at 100%, the
documentation, pre-commit.ci and every sdist install are green on
`27021e7`; four wheel and dynamic-wheel matrix jobs were still running
when I read them, and they are the only thing left to wait for.

fametrano:
Pushed at c18fa65 — the optional clause taken, which is the only change
since the ACK'd `27021e7`. The diff is CHANGELOG.md and the commit
message; no source, no test.

You made the case for it and against it fairly, and what decided me is
your own reading: the entry claims one question has one answer now, and
the both-parities assertion is what holds that claim — so the entry
should say so, the way *"Two tests hold the saving"* and *"a test of its
own holds the equality"* do in the same release.

The clause names the test, says why it takes two keys, and carries the
mutant in one sentence. It also states the half you checked and I had
not:

```
FAILED tests/test_examples.py::test_the_examples_of_a_module_run[btclib_secp256k1.xonly]
FAILED tests/test_verified_signing.py::test_the_keypair_is_checked_against_the_key_that_signed
FAILED tests/test_nonces.py::test_bip340_nonce_is_the_one_the_signature_was_made_with[odd y]
FAILED tests/test_parsed_keys.py::test_every_serialization_of_a_key_is_the_same_x_only_key
FAILED tests/test_modules.py::test_a_taproot_input_pays_as_its_even_y_key
FAILED tests/test_properties.py::test_schnorr_and_taproot_tweaking
FAILED tests/test_keys.py::test_xonly_pubkey_verify_and_to_pubkey_are_the_two_x_only_twins
FAILED tests/test_keys.py::test_xonly_from_prvkey
8 failed, 620 passed
```

Eight, as you had it, and the two you named are among them — so
*on both conversions the parity is the caller's to ask for* is a sentence
the suite holds on both, one of them by exactly one test. I re-derived
that rather than quoting your count.

On `ODD_Y_PRVKEY` beside `test_nonces.py`'s `ODD_Y`: agreed, and thank
you for raising and dropping it in the open — a scalar named for what it
exhibits, in the module that exhibits it, is what this suite does, and
sharing it would couple two test modules for nothing.

628 passed, coverage 100.00%, every hook passes. Waiting on the matrix,
then the fast-forward.

fametrano (commented):
## The ACK carries to `c18fa65`, and the convention has no third site

`c18fa65` differs from the `27021e7` I ACKed by the CHANGELOG paragraph
alone, and its claims are the ones I ran rather than new ones: the mutant
that discards the pointer inside the helper, and the same mutation on
`_drop_y` failing eight tests. Both figures are as stated.

### Whether anything is left of #212's bullet

I went looking for a third place the convention would have to reach, and
there is none. `Ignored if NULL` occurs exactly twice in the vendored
headers — both in `secp256k1_extrakeys.h`, and both are these two
conversions. The one other optional *out*-parameter in the wrapped
surface is `sigout` of `secp256k1_ecdsa_signature_normalize`, "can be
NULL if you're only interested in whether the signature was already
normalized", and `dsa` passes `ffi.NULL` for it already. Everything else
that takes a NULL there is an input or a function pointer — `noncefp`,
`ndata`, `extraparams`, the ECDH and ellswift hash functions, the
silent-payments label lookup — and `recovery`'s `recid` is
`SECP256K1_ARG_NONNULL(3)`, so the `ffi.new("int *")` beside it is
required and not an oversight of the same kind.

So this closes the family and not just #219: no follow-up issue to open,
and nothing to grep for again next release.

### One thing to do before the merge

The branch is `BEHIND`: `bea6a04` landed on `main` while this was in
review. It rebases onto `origin/main` with no conflict — CHANGELOG.md and
HISTORY.md included, though both commits write inside the v0.8.0.3
section — and the rebased commit still verifies, which I checked in a
throwaway worktree rather than assumed. So `git rebase origin/main`, a
`--force-with-lease` push, and then the fast-forward of the sha; still
one commit, so nothing to squash.

fametrano:
Rebased onto `main` at `bea6a04` (#218) and pushed: **68657e1**. Content
unchanged from the ACK'd `c18fa65` — a rebase and nothing else.

Pushed before waiting out `c18fa65`'s matrix on purpose: `main` moved
while it was running, so that run was on an object that would never
land. This way the matrix runs once, on the commit the fast-forward will
take.

`bea6a04` touches both `merge=union` files, so I checked the junctions
before anything else — the lesson of #215:

```
CHANGELOG.md:343  blank line before? True
HISTORY.md:320    blank line before? True
```

Both clean this time: each addition lands after an unchanged blank line,
so there was nothing to repair by hand.

**640 passed**, coverage 100.00%, every hook passes — twelve more tests
than before, which is `test_verified_signing.py` growing in `bea6a04`.
And the mutant still dies on the rebased tree:

```
FAILED tests/test_parsed_keys.py::test_the_keypair_parity_is_the_callers_to_ask_for
1 failed, 639 passed
```

One signed commit over `main`'s tip, so the fast-forward is available
unsquashed. Waiting on the matrix.

PR #221, opened 2026-08-17
Declare a buffer's width once, in every module that unpacks one

fametrano:
Eight sites still asked `ffi.sizeof` of a cdata at every call to say how
many bytes to unpack, where `keys`, `xonly` and `silentpayments` had
stopped doing so in #212: `dsa.serialize_der`, `dsa.serialize_compact`,
`recovery.serialize_compact`, `hashes.tagged_sha256`, `ellswift.create`,
`ellswift._encode_`, `ssa._sign32` and `ssa._sign_custom`. Each states its
width once at module level now, builds the buffer's type from it with
`ffi.typeof`, and uses it for the length — and where a module also
validates an argument against the width it writes, `dsa`, `recovery`,
`ssa` and `ellswift` check it against that one statement rather than a
second copy of the number. Nothing shared across modules: four of them
declare 64 independently, those being four different encodings that agree
on a number.

### What it saves

`main`'s spelling of each call written out and alternated against the
shipped one in a single process over one build **of the rebased tree**,
minimum of 15 rounds, a noise row for each site rather than for one of
them, every pair asserted to answer alike before it is timed. An Apple M5,
macOS 26.6, arm64, CPython 3.14.6, microseconds per call:

| | main | now | noise |
| --- | --- | --- | --- |
| `ffi.unpack` of a 64-byte buffer | 0.0690 | 0.0520 | 0.0005 |
| `dsa.serialize_compact` | 0.1823 | 0.1765 | 0.0007 |
| `dsa.serialize_der` | 0.2784 | 0.2665 | 0.0008 |
| `recovery.serialize_compact` | 0.2720 | 0.2569 | −0.0009 |
| `hashes.tagged_sha256` of an empty message | 0.5957 | 0.5800 | 0.0007 |

The per-site saving is 0.006 to 0.016 and is neither one number nor the
primitive's 0.017. Each noise row moves under 0.001, so no row above is
noise — but across four sessions the same site moved by more than its own
row, `dsa.serialize_compact` between 0.006 and 0.013, lowest of the four
every time, with `recovery` and `hashes` highest. What the measurement
supports is a hundredth of a microsecond at a serialization rather than a
figure per site; no row is a regression.

### Four of the eight save nothing measurable, and are changed anyway

`ellswift.create` and `ellswift._encode_` cost 19.90 and 14.85
microseconds, the two `ssa` signings some 30.4 each, and timing each
shipped call against *itself* moves 0.003 to 0.10 — as much as this is
worth or more. The figure is not what decides them. The cost of a host is
invisible where its buffer is declared, so two spellings of one shape
would be a difference no reader could account for at either site, and a
comment saying why a site was *left* alone documents a non-change. One
spelling, and the saving pays for the four where it shows.

### The mutation surface, measured rather than asserted

`core/NumberReplacer` emits two mutants per numeric literal — `dsa.py`
holds 20 and `cosmic-ray init` enumerates 40 of its 376 — and an AST
census of the package counts **67 literals before this change and 67
after**: `ellswift` −2 and `ssa` −1, `hashes` +1 and `keys` +2. The
surface is the size it was. Only three of the seven widths had no int
literal anywhere before (`hashes._HASH_SIZE`, `keys._COMPRESSED_SIZE`,
`keys._UNCOMPRESSED_SIZE`), so six mutants are new; the other four replace
the seven literals that already sat at the validation sites — `dsa`'s
`!= 64`, `recovery.parse_compact`, `ssa._verify_` and `ssa.verify`,
`ellswift._decode_` and both of `xdh`'s — fourteen mutants between them.

What changed is what each covers: a width inside `"char[64]"` was
reachable by no operator, so a wrong *buffer* was a mutation nobody could
ask about, and the mutable copy answered only for the argument check. One
number answers for both now — same count, more code under each. All
fourteen die, each applied to the source one at a time with
`PYTHONDONTWRITEBYTECODE=1` and a passing control run either side:

| constant | +1 | −1 |
| --- | --- | --- |
| `dsa._COMPACT_SIZE` | killed | killed |
| `recovery._COMPACT_SIZE` | killed | killed |
| `hashes._HASH_SIZE` | killed | killed |
| `ssa._SIGNATURE_SIZE` | killed | killed |
| `ellswift._ENCODING_SIZE` | killed | killed |
| `keys._COMPRESSED_SIZE` | killed | killed |
| `keys._UNCOMPRESSED_SIZE` | killed | killed |

`dsa`'s DER capacity is the one width that keeps its cdecl, with
`ffi.sizeof` deriving the capacity at import, and it is the site where
consolidating would have traded a kill for the symmetry: what
`serialize_der` unpacks is `length[0]`, so `_DER_SIZE = 73` leaves the
whole suite passing where 71 fails `test_der_reaches_all_72_octets` — the
shape `.github/mutation/bindings.toml` records as closed. Of the 40
`NumberReplacer` mutants `cosmic-ray init` enumerates in the module, the two
that fall in the declaration block are both on `_COMPACT_SIZE`, and none on
those two lines.

And what consolidating costs is recorded rather than left to be
rediscovered: `ellswift` had three independently mutable copies of 64 and
`ssa` two, so one mutant now changes every site at once and a check that
later loses its own test can no longer surface as a survivor.

### `keys` derives its two the other way round now

It had `ffi.typeof("char[33]")` with the size read back by `ffi.sizeof` of
that type: one statement of the width as well, but with the number inside
a C declaration where the other modules put it in an int. It is the int
here too, so the six modules read alike.

Rebased onto `68657e1`. **640 tests pass**, coverage 100%,
`uvx pre-commit run --all-files` exits 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -400,12 +400,112 @@ release-notes length in the first place, and are still in
   beside the hoisted types.
 - **`ffi.sizeof` of a buffer whose size is a constant is not asked per
   call**, and that is what asked `keys.serialize` for a type at module
-  level: each size is `ffi.sizeof` of the very type beside it, so the
+  level: the width is stated once and the type built from it, so the
   buffer and the length cannot say different numbers, and the 0.0175
   microseconds is not paid at every call. `keys.serialize` also decides
   its flag in the branch that picks its buffer, so the one condition is
   asked once. The length *object* is still built per call, for the
   thread-safety reason its comment gives.
+- **Every buffer in the package whose bytes are unpacked is spelled that
+  way now**, the eight sites that still asked `ffi.sizeof` of a cdata
+  per call included: `dsa.serialize_der`, `dsa.serialize_compact`,
+  `recovery.serialize_compact`, `hashes.tagged_sha256`,
+  `ellswift.create`, `ellswift._encode_`, `ssa._sign32` and
+  `ssa._sign_custom`. An int constant states the width, `ffi.typeof` of
+  it is the buffer's type, and that same constant is the length — and
+  where a module validates an argument against the width it writes,
+  `dsa`, `recovery`, `ssa` and `ellswift` check it against the constant
+  rather than against a second copy of the number.
+- **What the mutants say, which is not that there are more of them.**
+  `core/NumberReplacer` emits two mutants per numeric literal — `dsa.py`
+  holds 20 and `cosmic-ray init` enumerates 40 of them among its 376 — and
+  an AST census of the package counts 67 literals before this change and
+  67 after: `ellswift` sheds two and `ssa` one, `hashes` gains one and
+  `keys` two, and the surface is the size it was. Only three of the
+  seven widths had no int literal anywhere before (`hashes._HASH_SIZE`,
+  `keys._COMPRESSED_SIZE`, `keys._UNCOMPRESSED_SIZE`), so six mutants are
+  new; the other four replace the seven literals that sat at the
+  validation sites — `dsa`'s `!= 64`, `recovery.parse_compact`,
+  `ssa._verify_` and `ssa.verify`, `ellswift._decode_` and both of
+  `xdh`'s — which were fourteen of their own.
+  What changed is what each covers: a width inside `"char[64]"` was
+  reachable by no operator, so a wrong *buffer* was a mutation nobody
+  could ask about, and the mutable copy answered only for the argument
+  check. One number answers for both now, so each of the fourteen tests
+  both — same count, more code under each. All fourteen die, each applied
+  to the source one at a time with bytecode writing off and a passing
+  control run either side.
+- **The DER capacity is the one width not written as an int**, and it is
+  the site where taking that trade would have cost a kill rather than
+  bought one: what `serialize_der` unpacks is `length[0]`, the length
+  libsecp256k1 reports back, so 73 leaves the whole suite passing where 71
+  fails `test_der_reaches_all_72_octets` — the shape
+  `.github/mutation/bindings.toml` records as closed, six of thirteen
+  survivors, answered by deriving the capacity from the buffer rather than
+  writing it twice. So it keeps `ffi.typeof("char[72]")` with `ffi.sizeof`
+  of it, asked at import and not per call. Of the 40 `NumberReplacer`
+  mutants `cosmic-ray init` enumerates in the module, the two that fall in
+  the declaration block are both on `_COMPACT_SIZE` and none on those two
+  lines: the width is stated once and there is no int for the operator to
+  move.
+- **What consolidating costs, since nothing else records it.** `ellswift`
+  had three independently mutable copies of 64 — `_decode_` and both of
+  `xdh`'s — and `ssa` had two; each is one constant now, so a single
+  mutant changes every site at once and dies on whichever test reaches it
+  first. Nothing is hidden today: every one of those seven literals was
+  mutated on `main`, one at a time, and each dies on its own. But a check
+  that later loses the test that covers it can no longer surface as a
+  survivor, and `bindings.toml` asks for a survivor list to be read
+  expecting nothing — so the trade is named here rather than left for a
+  session to rediscover.
+- **Four of the eight save nothing the measurement can resolve, and are
+  spelled that way regardless.** The saving is 2.65% to 5.53% of the four
+  serializations, and some hundredths of a microsecond of the two
+  `ellswift` encodings and the two `ssa` signings, which cost 14.85 to
+  30.54 — where timing each of those four against *itself* moves 0.003 to
+  0.10, so the effect is under the resolution of its own measurement.
+  What decided them is not the figure: a reader cannot see the cost of a
+  host, so two spellings of one shape would leave the difference
+  unexplainable at both, and a comment saying why a site was *left* alone
+  is a comment about a non-change. One spelling is the whole of the
+  reason, and the saving is what pays for the four where it shows.
+- **`keys.py` derives its two the other way round**, so the six modules
+  read alike: it stated `ffi.typeof("char[33]")` and took the size as
+  `ffi.sizeof` of that type, where `xonly` and `silentpayments` state an
+  int and build the type from it. The int is the spelling that says what
+  the number is, and it is the one the whole package uses now.
+- **What the eight cost, and what they cost now.** `main`'s spelling of
+  each call written out and alternated against the shipped one in a single
+  process over one build on the tree this branch lands as, minimum of 15
+  rounds — 500 000 calls for the
+  primitive, 300 000 for the serializations, 20 000 for the four hosts —
+  with every pair asserted to answer alike before it is timed, and a noise
+  row for each site rather than for one of them. An Apple M5, macOS 26.6,
+  arm64, CPython 3.14.6, microseconds per call:
+
+  | | main | now | noise |
+  | --- | --- | --- | --- |
+  | `ffi.unpack` of a 64-byte buffer | 0.0690 | 0.0520 | 0.0005 |
+  | `dsa.serialize_compact` | 0.1823 | 0.1765 | 0.0007 |
+  | `dsa.serialize_der` | 0.2784 | 0.2665 | 0.0008 |
+  | `recovery.serialize_compact` | 0.2720 | 0.2569 | −0.0009 |
+  | `hashes.tagged_sha256` of an empty message | 0.5957 | 0.5800 | 0.0007 |
+
+  The per-site saving is 0.006 to 0.016, and it is neither one number nor
+  the primitive's 0.017. Within a session the noise rows move under 0.001,
+  so no row here is noise; across four sessions the same site moved by
+  more than that — `dsa.serialize_compact` between 0.006 and 0.013, the
+  lowest of the four every time, `recovery` and `hashes` the highest. So
+  what the measurement supports is a hundredth of a microsecond at a
+  serialization rather than a figure per site, and no row is a regression.
+
+  And the four hosts the saving disappears into, each timed against itself
+  so that the row *is* the resolution: `ellswift.encode` 14.9537 and
+  14.8501, `ellswift.create` 19.9003 and 19.9046, `ssa.sign` 30.2860 and
+  30.2031, `ssa.sign_custom` 30.5408 and 30.5379 — moving 0.003 to 0.10
+  either way around an effect of some hundredths. Every figure in the
+  comments of `dsa.py`, `recovery.py`, `hashes.py`, `ellswift.py`,
+  `ssa.py` and the last paragraph of `xonly.py`'s is from this session.
 - **`_secret.take` zeroes through the view it already holds.** It built
   an `ffi.buffer` to read the secret and then called `wipe`, which built
   a second one over the same cdata. The statement that writes the zeros
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -341,6 +341,17 @@ row beside each: `xonly.serialize` 0.197 microseconds against 0.257,
 Nothing on a signature or a verification, which are twelve microseconds
 of libsecp256k1; something on a caller doing point arithmetic in a loop,
 which is who these calls are short for.
+The second of those four now reaches every buffer in the package whose
+bytes are unpacked, and the serializations that were still working their
+own length out at each call are the ones it shows on:
+`dsa.serialize_compact` 0.177 microseconds against 0.182,
+`dsa.serialize_der` 0.267 against 0.278, `recovery.serialize_compact`
+0.257 against 0.272, `hashes.tagged_sha256` of an empty message 0.580
+against 0.596. The signing and encoding calls in `ssa` and `ellswift`
+carry the same spelling and gain nothing measurable from it, which is
+deliberate: one spelling for one shape is worth more to whoever reads
+this package than some hundredths of a microsecond are to whoever calls
+it.
 
 ## v0.8.0.2
 
--- a/btclib_secp256k1/dsa.py
+++ b/btclib_secp256k1/dsa.py
@@ -25,6 +25,38 @@ from ._scalar import in_range, octets, optional_entropy, scalar
 from ._secret import take
 from .context import ctx
 
+# the two buffers a signature is serialized into, and the lengths that go
+# with them in both directions: `_parsed` accepts a compact signature of
+# the width `serialize_compact` writes, so one statement of it answers
+# for the argument check as well. `ffi.sizeof` of a cdata is asked at
+# neither call, which is worth a hundredth of a microsecond of the 0.278
+# the DER serialization costs and of the 0.182 the compact one does --
+# 0.012 and 0.006 in the session `xonly.py` names, and not a figure this
+# site can be held to between sessions: that comment says why.
+#
+# The compact width is an int and the DER capacity is not, which is the
+# one place this module departs from the other five, and `length[0]` is
+# the reason: what `serialize_der` unpacks is the length libsecp256k1
+# reports back, so a capacity above 72 is absorbed on the way out and no
+# test can tell it from 72 -- verified, 73 leaves the suite passing where
+# 71 fails `test_der_reaches_all_72_octets`. Written as an int it would
+# be a number the mutation operator reaches and nothing checks, which is
+# the shape `.github/mutation/bindings.toml` records as closed, six of
+# thirteen survivors having been it. So it stays inside the cdecl, where
+# the width is still stated once and `ffi.sizeof` derives the capacity
+# from the buffer's own type -- at import, not per call.
+#
+# 72 is the maximum a signature of this curve can encode to, and it is
+# structural rather than generous: secp256k1_ecdsa_sig_serialize writes
+# 6 + lenR + lenS, and each of those is at most 33 -- 32 octets of
+# scalar and the leading zero DER wants when the top bit is set.
+# `to_der` reaches it, a high-s signature being one it serializes rather
+# than refuses
+_DER_BUFFER_TYPE = ffi.typeof("char[72]")
+_DER_SIZE = ffi.sizeof(_DER_BUFFER_TYPE)
+_COMPACT_SIZE = 64
+_COMPACT_BUFFER_TYPE = ffi.typeof(f"char[{_COMPACT_SIZE}]")
+
 
 @overload
 def nonce_rfc6979(
@@ -792,7 +824,7 @@ def _parsed(signature_bytes: BytesLike, compact: bool) -> CData | None:
     signature_bytes = octets(signature_bytes, name)
     signature = ffi.new("secp256k1_ecdsa_signature *")
     if compact:
-        if len(signature_bytes) != 64:
+        if len(signature_bytes) != _COMPACT_SIZE:
             return None
         parsed = lib.secp256k1_ecdsa_signature_parse_compact(
             ctx, signature, signature_bytes
@@ -857,21 +889,20 @@ def serialize_der(signature: CData) -> bytes:
             72-byte buffer makes unreachable. `context.check` is what
             tells the two apart.
     """
-    # 72 is the maximum a signature of this curve can encode to, and it
-    # is structural rather than generous: secp256k1_ecdsa_sig_serialize
-    # writes 6 + lenR + lenS, and each of those is at most 33 -- 32
-    # octets of scalar and the leading zero DER wants when the top bit
-    # is set. `to_der` reaches it, a high-s signature being one it
-    # serializes rather than refuses.
+    # the length passed in is `ffi.sizeof` of the buffer's own type, so
+    # the two cannot drift apart -- the top of this module says where 72
+    # comes from and why it is the one width not spelled as an int -- and
+    # what comes out is what libsecp256k1 says it wrote, this being the
+    # one serialization here whose length varies, 70 to 72 depending on
+    # how many octets r and s need. Where the length is fixed the width
+    # itself is unpacked instead, and keys.serialize says why.
     #
-    # The number is written once and the rest is derived from it: the
-    # length passed in is the buffer's own size, so the two cannot drift
-    # apart, and what comes out is what libsecp256k1 says it wrote --
-    # this being the one serialization here whose length varies, 70 to 72
-    # depending on how many octets r and s need. Where the length is
-    # fixed the buffer is unpacked instead, and keys.serialize says why
-    sig_bytes = ffi.new("char[72]")
-    length = ffi.new("size_t *", ffi.sizeof(sig_bytes))
+    # The length *object* is built per call and not hoisted, for the
+    # thread-safety reason keys.serialize gives: libsecp256k1 writes 0
+    # into it before it does anything and a shared one left at zero would
+    # refuse every later serialization
+    sig_bytes = ffi.new(_DER_BUFFER_TYPE)
+    length = ffi.new("size_t *", _DER_SIZE)
     serialized = lib.secp256k1_ecdsa_signature_serialize_der(
         ctx, sig_bytes, length, signature
     )
@@ -896,10 +927,10 @@ def serialize_compact(signature: CData) -> bytes:
             signature it parsed cannot make it do. `context.check` is
             what tells the two apart.
     """
-    sig_bytes = ffi.new("char[64]")
+    sig_bytes = ffi.new(_COMPACT_BUFFER_TYPE)
     serialized = lib.secp256k1_ecdsa_signature_serialize_compact(
         ctx, sig_bytes, signature
     )
     if not serialized:
         raise RuntimeError("signature serialization failed")
-    return ffi.unpack(sig_bytes, ffi.sizeof(sig_bytes))
+    return ffi.unpack(sig_bytes, _COMPACT_SIZE)
--- a/btclib_secp256k1/ssa.py
+++ b/btclib_secp256k1/ssa.py
@@ -23,6 +23,16 @@ from .context import ctx
 # survive the preprocessing of the headers into cffi definitions
 EXTRAPARAMS_MAGIC = b"\xda\x6f\xb3\x8c"
 
+# the width of a BIP340 signature: 64 bytes, which is what `_sign32` and
+# `_sign_custom` write and what `_verify_` and `verify` read back, so the
+# one statement of it answers for the argument checks as well as for the
+# buffer, whose type is built from it. It saves some hundredths of a
+# microsecond on calls of 30.20 and 30.54, where timing each of those
+# against itself moves 0.08 and 0.003: nothing that measurement resolves,
+# and not the reason either -- `xonly.py` says the reason
+_SIGNATURE_SIZE = 64
+_SIGNATURE_BUFFER_TYPE = ffi.typeof(f"char[{_SIGNATURE_SIZE}]")
+
 # the tag secp256k1_schnorrsig_sign gives its nonce function, and what
 # makes the derivation BIP340's rather than another protocol's
 _NONCE_ALGO = b"BIP0340/nonce"
@@ -499,7 +509,7 @@ def _verify_(
         ValueError: if the signature is not 64 bytes.
     """
     msg_bytes = octets(msg_bytes, "message")
-    signature_bytes = octets(signature_bytes, "signature", 64)
+    signature_bytes = octets(signature_bytes, "signature", _SIGNATURE_SIZE)
 
     verified = lib.secp256k1_schnorrsig_verify(
         ctx, signature_bytes, msg_bytes, len(msg_bytes), xonly_pubkey
@@ -544,7 +554,7 @@ def verify(
     # key, and tests/test_parsed_keys.py asserts the two answer alike
     xonly_pubkey = xonly.parse(pubkey_bytes)
     msg_bytes = octets(msg_bytes, "message")
-    signature_bytes = octets(signature_bytes, "signature", 64)
+    signature_bytes = octets(signature_bytes, "signature", _SIGNATURE_SIZE)
     return bool(
         lib.secp256k1_schnorrsig_verify(
             ctx, signature_bytes, msg_bytes, len(msg_bytes), xonly_pubkey
@@ -636,12 +646,12 @@ def _sign32(
     """
     msg_bytes = octets(msg_bytes, "message hash", 32)
 
-    sig = ffi.new("char[64]")
+    sig = ffi.new(_SIGNATURE_BUFFER_TYPE)
     if not lib.secp256k1_schnorrsig_sign32(
         ctx, sig, msg_bytes, keypair_obj, entropy(aux_rand32)
     ):
         raise RuntimeError("schnorr signing failed")
-    signature_bytes = ffi.unpack(sig, ffi.sizeof(sig))
+    signature_bytes = ffi.unpack(sig, _SIGNATURE_SIZE)
     if verify:
         _abort_unless_verified(keypair_obj, msg_bytes, signature_bytes)
     return signature_bytes
@@ -678,7 +688,7 @@ def _sign_custom(
     """
     msg_bytes = octets(msg_bytes, "message")
 
-    sig = ffi.new("char[64]")
+    sig = ffi.new(_SIGNATURE_BUFFER_TYPE)
     ndata = ffi.new("char[32]", entropy(aux_rand32))
     extraparams = ffi.new("secp256k1_schnorrsig_extraparams *")
     extraparams.magic = EXTRAPARAMS_MAGIC
@@ -691,7 +701,7 @@ def _sign_custom(
         ctx, sig, msg_bytes, len(msg_bytes), keypair_obj, extraparams
     ):
         raise RuntimeError("schnorr signing failed")
-    signature_bytes = ffi.unpack(sig, ffi.sizeof(sig))
+    signature_bytes = ffi.unpack(sig, _SIGNATURE_SIZE)
     if verify:
         _abort_unless_verified(keypair_obj, msg_bytes, signature_bytes)
     return signature_bytes
--- a/btclib_secp256k1/xonly.py
+++ b/btclib_secp256k1/xonly.py
@@ -50,11 +50,34 @@ _XONLY_SIZE = 32
 # evaluated once, is the literal's price with the width still stated
 # once.
 #
-# A *literal* cdecl is already cached by cffi, and hoisting one saves
-# 0.003 -- 0.0656 against 0.0685, real at 4.3% and an order below the
-# 0.054 above. Too small to be worth a module-level name for every
-# cdecl in the package, so this is done where the string is built and
-# every other `ffi.new` here is spelled in full
+# What earns the name is the length rather than the cdecl. A name keeps
+# two statements of one width in step, and a buffer whose bytes nothing
+# unpacks states its width once already: naming that one would add a
+# statement rather than remove one, so `ecdh.py`'s `ffi.new("char[32]")`
+# and every other cdecl of that kind stay spelled in full. The price
+# agrees and does not decide it: cffi already caches the parse of a
+# literal cdecl, so hoisting one saves 0.003 -- 0.0656 against 0.0685,
+# real at 4.3% and an order below the 0.054 above.
+#
+# Every buffer in this package whose bytes *are* unpacked is declared the
+# way below -- an int constant, and `ffi.typeof` of it -- so that the
+# width is stated once and `ffi.sizeof` of a cdata is not asked per
+# call, `dsa`'s DER capacity excepted for the reason its own comment
+# gives: 0.0690 against 0.0520 microseconds on the primitive alone. That
+# is a session of its own, which CHANGELOG.md names, and every figure in
+# the comments of `dsa`, `recovery`, `hashes`, `ellswift` and `ssa` comes
+# from it rather than from the one above. At a call site it is 2.65% to
+# 5.53% of the serializations in the first three of those -- 0.006 to
+# 0.016 microseconds, which is neither one number nor the primitive's:
+# within a session each noise row moves under 0.001, and across four
+# sessions the same site moved by more than that, `dsa.serialize_compact`
+# between 0.006 and 0.013. What is solid is a hundredth of a microsecond
+# at a serialization, not a figure per site. In `ssa` and `ellswift` it is
+# that same hundredth of calls of 15 to 31 microseconds, where timing a
+# call against itself already moves as much or more: nothing those
+# measurements resolve. Those four sites are spelled this way regardless,
+# because a reader cannot see the cost of a host -- two spellings of one
+# shape would leave the difference unexplainable at both of them
 _XONLY_BUFFER_TYPE = ffi.typeof(f"char[{_XONLY_SIZE}]")
 
 
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -139,7 +139,7 @@
         "filename": "btclib_secp256k1/hashes.py",
         "hashed_secret": "a608d88cb7e04b46b5e44fa00bb7c01b7748e6f0",
         "is_verified": false,
-        "line_number": 40
+        "line_number": 49
       }
     ],
     "scripts/cffi_build.py": [
@@ -517,5 +517,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-17T08:06:26Z"
+  "generated_at": "2026-08-17T10:49:53Z"
 }
--- a/btclib_secp256k1/ellswift.py
+++ b/btclib_secp256k1/ellswift.py
@@ -18,6 +18,17 @@ from ._secret import take
 from .context import ctx
 from .keys import parse, serialize
 
+# the width of an ElligatorSwift encoding: 64 bytes, the two field
+# elements it is. `create` and `_encode_` write it, `_decode_` and `xdh`
+# accept it, and the one statement of it answers for all four -- the
+# buffer's type is built from it and so is every length check. It saves
+# some hundredths of a microsecond on calls of 19.90 and 14.85, where
+# timing each of those against itself moves 0.004 and 0.10: nothing that
+# measurement resolves, and not the reason either -- `xonly.py` says the
+# reason
+_ENCODING_SIZE = 64
+_ENCODING_BUFFER_TYPE = ffi.typeof(f"char[{_ENCODING_SIZE}]")
+
 
 def create(prvkey: BytesLike | int, aux_rand32: BytesLike | None = None) -> bytes:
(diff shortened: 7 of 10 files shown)


btclib_secp256k1/dsa.py
```diff
@@ -25,6 +25,24 @@
...
+# serialization costs, 0.0112 of the 0.1850 the compact one does.
+# `xonly.py` carries the session behind the spelling.
+#
+# 72 is the maximum a signature of this curve can encode to, and it is
+# structural rather than generous: secp256k1_ecdsa_sig_serialize writes
+# 6 + lenR + lenS, and each of those is at most 33 -- 32 octets of
+# scalar and the leading zero DER wants when the top bit is set.
+# `to_der` reaches it, a high-s signature being one it serializes rather
+# than refuses
+_DER_SIZE = 72
```
fametrano: `_DER_SIZE = 72` as a bare int is the one place this PR loses something, and
`.github/mutation/bindings.toml` is where it is written down: the DER capacity
"was slack over a serialization that cannot exceed 72", six of thirteen
survivors were that shape, and the answer taken then was to stop writing the
number twice and derive the capacity from the buffer with `ffi.sizeof`. On
`main` there is no mutant at all here — the 72 lives inside `"char[72]"`, a
string constant no operator touches, and the capacity is `ffi.sizeof` of the
cdata. Verified, bytecode writing off, one mutant at a time:

- `_DER_SIZE = 73` -> `627 passed in 2.74s` — **survives**
- `_DER_SIZE = 71` -> `tests/test_core.py::test_der_reaches_all_72_octets`
  fails, `RuntimeError: signature serialization failed` — dies

Widening survives because this is the one site whose capacity is not what is
unpacked: `serialize_der` returns `ffi.unpack(sig_bytes, length[0])`, so any
slack over 72 is absorbed on the way out. The seven fixed-width constants have
no such gap, which is why all fourteen of their mutants die and why the int
spelling is right for them.

Type-first for this one site keeps everything the PR is for and closes it again:

```python
_DER_BUFFER_TYPE = ffi.typeof("char[72]")
_DER_SIZE = ffi.sizeof(_DER_BUFFER_TYPE)
```

Verified: 627 pass, `_DER_SIZE == 72`, `_DER_BUFFER_TYPE` is `<ctype
'char[72]'>`, and 72 is gone from the module's int literals, so there is nothing
for the number operator to mutate. `ffi.sizeof` is asked once at import and not
per call, so the 0.0090 microseconds this row claims is untouched — the saving
was never the `ffi.sizeof`, it was asking it at module level instead of inside
the call.

This is the `keys.py` spelling the PR reverses, and DER is the site where the
reversal is not free: a fixed width unpacked as written keeps the two statements
in step in either direction, while a reported length only does so when the
capacity cannot be read back as a bare number. If the uniform int spelling is
worth more than the kill, then the survivor belongs in `bindings.toml`'s list of
shapes to read once and skip — but it should be a decision written down, not a
regression against a comment that says the shape is gone.

btclib_secp256k1/ssa.py
```diff
@@ -23,6 +23,16 @@
 # survive the preprocessing of the headers into cffi definitions
 EXTRAPARAMS_MAGIC = b"\xda\x6f\xb3\x8c"
 
+# the buffer a signature is written into, by `_sign32` and by
```
fametrano: These four constants also gate an input length now, and the comment at each
declaration says only that a buffer is written into. `ssa.py` gets it right in
passing — "what BIP340 signs and what `verify` reads back" — so this one is the
model; `dsa.py:28`, `recovery.py:22` and `ellswift.py:21` are the three that
read as output-only while `_parsed`, `parse_compact`, `_decode_` and `xdh`
depend on them. One clause each.

btclib_secp256k1/xonly.py
```diff
@@ -52,9 +52,24 @@
 #
 # A *literal* cdecl is already cached by cffi, and hoisting one saves
 # 0.003 -- 0.0656 against 0.0685, real at 4.3% and an order below the
-# 0.054 above. Too small to be worth a module-level name for every
-# cdecl in the package, so this is done where the string is built and
-# every other `ffi.new` here is spelled in full
+# 0.054 above. Too small to be worth a module-level name of its own, so
```
fametrano: The price argument lands here and is withdrawn two paragraphs down. As ordered,
`ecdh.py:74`'s `ffi.new("char[32]")` reads as one of the "two spellings of one
shape" the last paragraph says a reader cannot account for. "What earns the name
is the length rather than the cdecl" is the reason that survives both paragraphs
— a name keeps two statements of one width in step, and a `take`-bound buffer
has only one — so it would carry the boundary better as the first sentence than
as the third.

CHANGELOG.md
```diff
@@ -330,12 +330,62 @@ release-notes length in the first place, and are still in
...
+- **What the eight cost, and what they cost now.** `main`'s spelling of
+  each call written out and alternated against the shipped one in a single
+  process over one build, minimum of 9 rounds — 500 000 calls for the
+  primitive, 300 000 for the serializations, 20 000 for the four hosts —
+  with every pair asserted to answer alike before it is timed. An Apple
+  M5, macOS 26.6, arm64, CPython 3.14.6, microseconds per call:
+
+  | | main | now |
+  | --- | --- | --- |
+  | `ffi.unpack` of a 64-byte buffer | 0.0686 | 0.0520 |
```
fametrano: The primitive row is the removed work, 0.0166, and `xonly.py` adds 0.003 for the
hoisted cdecl, so the mechanism predicts about 0.020 at each of the four sites
where the saving is claimed to show. `recovery.serialize_compact` (0.0171) and
`hashes.tagged_sha256` (0.0179) show it; `dsa.serialize_der` (0.0090) and
`dsa.serialize_compact` (0.0112) show about half, against noise rows of 0.0001
and 0.0005. Either the two low rows have a cause worth a clause, or they want a
noise row of their own — `serialize_der` has none, and it is the row furthest
from the story.

fametrano (commented):
## Review: changes requested

Reviewed the tree rather than the diff. Worktree at `ecf7f97`, submodule at
`6e2c8bc`, `uv sync --locked --no-default-groups --group test`, `pytest -q -p
no:randomly`: **627 passed**, which is the count the description claims, and
CI's coverage gate and pre-commit.ci are both green.

The refactor is right in itself. One width per buffer, `ffi.typeof` of it, and
`ffi.sizeof` of a cdata gone from every per-call path — that is what the
description says it does, and it does it. What I checked rather than assumed is
the dimension the description leans on hardest, the mutant: on `main` most of
these widths lived inside a cdecl string, where no mutation operator reaches
them, so moving each into an int literal *creates* mutants that did not exist.
Every one of those but a single mutant dies, verified one at a time on a tree
with bytecode writing off:

| constant | +1 | -1 |
| --- | --- | --- |
| `dsa._COMPACT_SIZE` | killed | killed |
| `hashes._HASH_SIZE` | killed | killed |
| `ssa._SIGNATURE_SIZE` | killed | killed |
| `ellswift._ENCODING_SIZE` | killed | killed |
| `recovery._COMPACT_SIZE` | killed | killed |
| `keys._COMPRESSED_SIZE` | killed | killed |
| `keys._UNCOMPRESSED_SIZE` | killed | killed |
| `dsa._DER_SIZE` | **SURVIVED** | killed |

So fourteen new mutants and fourteen kills, which is the argument for the int
spelling made by measurement instead of by assertion — and one survivor, at the
one site where the reversal costs something. That is the change I am asking
for; the four notes after it are suggestions, none of them blocking.

### Change requested — `_DER_SIZE` re-opens a survivor `bindings.toml` records as closed

Inline on `dsa.py`. In short: `_DER_SIZE = 72` -> `73` leaves all 627 tests
passing, because the DER capacity is not what is unpacked — `length[0]` is, so
slack over 72 is absorbed exactly as it was before the session that removed
it. `.github/mutation/bindings.toml` names this shape as one of the two "no test
could kill" that were "answered in the code rather than excused in a comment",
DER buffer included, and the session behind that configuration reported no
survivor at all. Read against that comment, this PR puts one back.

The commit message's closing claim — "the widening mutant still dies: the length
and the buffer are one statement" — is true of the seven fixed-width constants
and false of `_DER_SIZE`, which is the only one whose length is reported back
rather than unpacked as written. Worth qualifying wherever it is restated.

### Suggested — the declaration comments describe half of what each constant does

Four of them now gate an *input* length as well as size an output buffer, which
the description presents as a feature of the change ("check it against the
constant rather than a second copy of the number"), and none of the four
comments says so:

- `dsa.py:28` — "the two buffers a signature is serialized into", also
  `_parsed`'s accepted compact length
- `recovery.py:22` — "the buffer `serialize_compact` writes the signature into",
  also `parse_compact`'s
- `ssa.py:26` — "the buffer a signature is written into", also `_verify_`'s and
  `verify`'s
- `ellswift.py:21` — "the buffer an ElligatorSwift encoding is written into, by
  `create` and by `_encode_` alike", also `_decode_`'s and `xdh`'s

A reader landing on `ssa._verify_` and following `_SIGNATURE_SIZE` up arrives at
a comment about a buffer that verification does not allocate. `ssa.py`'s already
gets the direction right in passing — "what BIP340 signs and what `verify` reads
back" — so this is one clause in each of the other three.

### Suggested — the two noise rows do not cover the two rows that need them

The table says the removed work is 0.0166 on the primitive, and `xonly.py` says
hoisting the cdecl is worth 0.003 more, so the story predicts about 0.020 at
each of the four sites where it is claimed to show. Two rows show it —
`recovery.serialize_compact` 0.0171, `hashes.tagged_sha256` 0.0179 — and two
show about half: `dsa.serialize_der` 0.0090, `dsa.serialize_compact` 0.0112.
The noise rows quoted are the primitive (0.0001) and `dsa.serialize_compact`
(0.0005), both an order below that spread, so as it stands the table cannot
distinguish "the change is worth less at those two sites" from "those two rows
are noisier than the two that were controlled". A noise row for
`serialize_der`, or a sentence saying the per-site figure is not expected to be
the same number, would settle it. Nothing here argues against the change: no
row is a regression.

### Suggested — `xonly.py` gives the perf argument as the reason, one paragraph before rejecting it

`xonly.py:55` reads "Too small to be worth a module-level name of its own, so a
cdecl whose buffer nothing unpacks is still spelled in full", and the paragraph
right after it argues that 0.017 is *not* what decides `ssa` and `ellswift` —
uniformity is. As ordered, the boundary looks decided by opposite criteria on
either side of it, and `ecdh.py:74`'s `ffi.new("char[32]")` reads as exactly the
"two spellings of one shape" the next paragraph says a reader cannot account
for. The real reason is the sentence after: "What earns the name is the length
rather than the cdecl" — a name keeps two statements of one width in step, and a
buffer handed to `take` states its width once, so naming it would add a
statement rather than remove one. That reason survives the paragraph that
follows; the price argument does not. Lead with it.

### Nit — the generated summary in the description contradicts the title

"introducing **shared** integer size constants", "the **shared** width
constants": nothing is shared across modules, and per-module is the design — the
title says so. Four modules declare 64 independently and should, those being
four different encodings that agree on a number. Worth trimming from the body,
which is what a reader of the merged history sees.

---

**Verdict.** Not an ACK yet, on the one line above; everything else is a
suggestion and none of it blocks. With `_DER_SIZE` derived rather than written,
this is a clear improvement — including in mutation coverage, where it would
then be fourteen new mutants and fourteen kills with nothing reopened.

*(GitHub will not take a formal `REQUEST_CHANGES` from the author's own token,
so this lands as a comment; read the verdict here as the review state.)*

fametrano:
Pushed as `859c21a`. The change requested is in, and so are the four
suggestions; every claim below was re-run rather than reasoned about.

### `_DER_SIZE` — done, and the survivor verified first

Reproduced before changing anything, bytecode writing off, one mutant at a
time: `_DER_SIZE = 73` → `627 passed`, `_DER_SIZE = 71` → `test_der_reaches_all_72_octets`
fails with `RuntimeError: signature serialization failed`. So the reading
is right: the DER capacity is the one width that is not what gets
unpacked, and as a bare int it was a number nothing checks.

Type-first for that site, as suggested:

```python
_DER_BUFFER_TYPE = ffi.typeof("char[72]")
_DER_SIZE = ffi.sizeof(_DER_BUFFER_TYPE)
```

`_DER_SIZE == 72`, `_DER_BUFFER_TYPE` is `<ctype 'char[72]'>`, and an AST
walk over the module confirms 72 is gone from its int constants — the
remaining ones among the widths are `32` (message hashes) and `_COMPACT_SIZE`'s
`64`. The comment at the declaration now carries the reason, `length[0]`
and the `bindings.toml` shape both, so the next reader does not have to
rediscover why one width in this module is spelled unlike the other five.

I also stopped citing your table and re-ran it: **14 mutants, 14 killed, 0
survived** across the seven fixed widths, each ±1, applied to the source
one at a time with `PYTHONDONTWRITEBYTECODE=1` and a passing control run
before and after. The table is in the description now.

### The declaration comments — done

Each of the four now says both directions in its first sentence: "the
width of a compact signature, in both directions", "the width of an
ElligatorSwift encoding … `create` and `_encode_` write it, `_decode_` and
`xdh` accept it", "the width of a BIP340 signature … what `_verify_` and
`verify` read back", and `dsa`'s names `_parsed`.

### The noise rows — measured, and the answer is not the one I expected

A noise row per site, 15 rounds instead of 9. The two low rows were partly
measurement, but not entirely, and the honest reading is in the
description: the per-site saving is **0.011 to 0.020**, the noise rows are
under **0.002** at every site, and the `dsa` pair came out lowest in each
of three sessions while `recovery` and `hashes` came out highest — a
repeatable spread the noise rows do not account for. So neither "the two
rows are noisier" nor a clean mechanism: the primitive's 0.018 is an upper
bound on what any site can save, and each row is what that site saves.
Said that way in `xonly.py` and CHANGELOG.md rather than left to be
inferred from a single noise row.

### `xonly.py`'s paragraph order — done

"What earns the name is the length rather than the cdecl" leads now, with
the reason you gave it: a name keeps two statements of one width in step,
and a buffer nothing unpacks has only one, so naming it would add a
statement. `ecdh.py`'s `ffi.new("char[32]")` is named there as the case
that reads correctly under that rule. The price argument follows as
agreement rather than as the criterion.

### The generated summary — trimmed

The whole "Summary by Sourcery" block is out of the description, "shared"
included. The one occurrence left is the sentence that denies it: four
modules declare 64 independently, those being four different encodings
that agree on a number.

Suite 627 passed, coverage 100%, `uvx pre-commit run --all-files` exits 0.
Not merging: the verdict was not an ACK, and this is a re-review request
rather than a claim that it is one now.

CHANGELOG.md
```diff
@@ -330,12 +330,82 @@ release-notes length in the first place, and are still in
...
+  `ellswift.create`, `ellswift._encode_`, `ssa._sign32` and
+  `ssa._sign_custom`. An int constant states the width, `ffi.typeof` of
+  it is the buffer's type, and that same constant is the length — and
+  where a module validates an argument against the width it writes,
+  `dsa`, `recovery`, `ssa` and `ellswift` check it against the constant
+  rather than against a second copy of the number.
+- **The DER capacity is the one width not written as an int, and the
+  mutants are why.** Moving a width out of a cdecl string and into an int
+  literal *creates* mutants where none existed: a number inside
+  `"char[64]"` is a string constant no operator reaches. Fourteen new ones
```
fametrano: "Fourteen new ones for the seven fixed widths" — six, and the package's
`NumberReplacer` surface does not grow at all.

Calibration first, so the count is the tool's and not mine: `core/NumberReplacer`
emits exactly two mutants per int literal — `dsa.py` holds 20 and `cosmic-ray
init` enumerates 40. Then an AST census of every int literal under
`btclib_secp256k1/`, current `main` against this branch rebased onto it:
`ellswift` 4 -> 2, `ssa` 3 -> 2, `hashes` 0 -> 1, `keys` 2 -> 4, **total 67 ->
67**.

Three of the seven constants are widths that had no int literal anywhere on
`main` — `hashes._HASH_SIZE`, `keys._COMPRESSED_SIZE`, `keys._UNCOMPRESSED_SIZE`
— so six mutants are genuinely new. The other four replace seven literals `main`
already had at the validation sites, `dsa.py:795`'s `!= 64`, `recovery.py:357`,
`ssa.py:502` and `:547`, `ellswift.py:128`, `:215` and `:216`, which were
fourteen mutants of their own. So the width mutants number fourteen on either
side of this change; what moved is where they sit.

Which is the better sentence anyway: on `main` the mutable copy of a width
reached only the argument check, and the buffer's width — inside a cdecl string —
was reachable by nothing, so a wrong buffer was a mutation nobody could ask
about. One number answers for both now, so each of the fourteen tests both. Same
count, more code under each, and the DER site is then the one place where taking
that trade would have cost a kill instead of buying one. The 14/14 table below
stands: I re-ran it constant by constant, ±1, bytecode writing off, control
passing either side.

CHANGELOG.md
```diff
@@ -330,12 +330,82 @@ release-notes length in the first place, and are still in
...
+  `length[0]`, the length libsecp256k1 reports back, so 73 leaves the
+  whole suite passing where 71 fails `test_der_reaches_all_72_octets`.
+  That is exactly the shape `.github/mutation/bindings.toml` records as
+  closed — six of thirteen survivors, answered by deriving the capacity
+  from the buffer rather than writing it twice — so this one keeps
+  `ffi.typeof("char[72]")` with `ffi.sizeof` of it, asked at import and
+  not per call. The width is still stated once, and there is no int for
+  the operator to move.
+- **Four of the eight save nothing the measurement can resolve, and are
+  spelled that way regardless.** The saving is 2.99% to 6.23% of the four
```
fametrano: This bullet is the first session's numbers and the table three bullets down is
the second's. 2.99% to 6.23% against 3.37% to 6.42%; 0.017 microseconds against
0.011 to 0.020; hosts of 14.74 to 30.71 against 14.84 to 30.72; self-timings of
0.01 to 0.08 against 0.01 to 0.09. `xonly.py` now says "3.38% to 6.42%", and
`ssa.py` and `ellswift.py` say "some hundredths of a microsecond" where this says
0.017 — so the bullet contradicts the comments that the bullet after it says are
all from this session. The description's counterpart section was rewritten; this
one was missed.

btclib_secp256k1/ssa.py
```diff
@@ -23,6 +23,17 @@
 # survive the preprocessing of the headers into cffi definitions
 EXTRAPARAMS_MAGIC = b"\xda\x6f\xb3\x8c"
 
+# the width of a BIP340 signature: 64 bytes, which is what `_sign32` and
+# `_sign_custom` write and what `_verify_` and `verify` read back, so the
+# one statement of it answers for the argument checks as well as for the
+# buffer, whose type is built from it. It
```
fametrano: Left mid-reflow: 41 columns, and the sentence resumes on the next line. Nothing
lints it — ruff does not reflow comments — and it is the only comment line under
50 columns in any of the six blocks this PR writes, so it reads as an edit
artefact rather than a paragraph break.

HISTORY.md
```diff
@@ -324,6 +324,16 @@ row beside each: `xonly.serialize` 0.197 microseconds against 0.257,
...
+The second of those four now reaches every buffer in the package whose
+bytes are unpacked, and the serializations that were still working their
+own length out at each call are the ones it shows on:
+`dsa.serialize_compact` 0.177 microseconds against 0.189,
+`dsa.serialize_der` 0.271 against 0.282, `recovery.serialize_compact`
+0.261 against 0.279, `hashes.tagged_sha256` of an empty message 0.576
+against 0.596. The signing and encoding calls in `ssa` and `ellswift`
+carry the same spelling and gain nothing measurable from it, which is
+deliberate: one spelling for one shape is worth more to whoever reads
+this package than 0.018 microseconds is to whoever calls it.
```
fametrano: 0.018 is the primitive's saving, which the new table and `xonly.py` both say is
not a per-site figure — and the sites this clause is about, the `ssa` signings
and the `ellswift` encodings, are the ones whose saving is under their own
resolution, so 0.018 is not what their callers gain either. "Some hundredths of a
microsecond", which is what `ssa.py` and `ellswift.py` now say, would agree with
the rest of the change.

fametrano (commented):
## Review of `859c21a`: code ACK, two corrections wanted in the record

Rebuilt and re-measured rather than read: worktree at `859c21a`, submodule
`6e2c8bc`, `uv sync --locked`, **627 passed**. The change requested last round is
in, and I verified it with the tool instead of by hand — a `cosmic-ray init`
scoped to `btclib_secp256k1/dsa.py` enumerates 376 mutants, of which exactly two
fall in the module-constant block, both `core/NumberReplacer` on line 56
(`_COMPACT_SIZE = 64`), and **none on lines 54–55**. The DER width is out of the
operator's reach, the comment carries `length[0]` and the `bindings.toml` shape
so the one asymmetric spelling is readable at the site, and an AST walk confirms
72 is gone from the module's int constants. The four declaration comments now say
both directions, `xonly.py` leads with the length rather than the price and names
`ecdh.py`'s cdecl as the case the rule covers, and the description no longer says
"shared". **The code is an ACK from me as it stands.**

I also re-ran your mutant table independently, each constant ±1, one at a time,
`PYTHONDONTWRITEBYTECODE=1`, with a passing control before and after: **14 of
14 killed**, matching yours exactly.

What is not ready is `CHANGELOG.md`, and one of the two is a measured claim that
does not hold.

### Necessary — "Fourteen new ones" is off by eight, and the surface is not larger at all

Inline at `CHANGELOG.md:352`. Calibrated first: `core/NumberReplacer` emits
exactly two mutants per int literal — `dsa.py` has 20 literals and cosmic-ray
enumerates 40. Then an AST census of every int literal in `btclib_secp256k1/`,
current `main` against this branch **rebased onto it**:

| module | main | this branch | Δ mutants |
| --- | --- | --- | --- |
| `ellswift.py` | 4 | 2 | −4 |
| `ssa.py` | 3 | 2 | −2 |
| `hashes.py` | 0 | 1 | +2 |
| `keys.py` | 2 | 4 | +4 |
| **package total** | **67** | **67** | **0** |

The `NumberReplacer` surface is unchanged, not fourteen larger. Only three of the
seven constants are widths that had no int literal anywhere on `main` —
`hashes._HASH_SIZE`, `keys._COMPRESSED_SIZE`, `keys._UNCOMPRESSED_SIZE` — so
**six** mutants are new. The other four replace seven literals `main` already
had at the validation sites: `dsa.py:795` `!= 64`, `recovery.py:357`,
`ssa.py:502` and `:547`, `ellswift.py:128`, `:215` and `:216`, fourteen mutants
between them.

The kills are real and the table stays; the sentence above it has to go, and what
replaces it is a stronger claim than the one it makes: each of the fourteen now
covers **both** the buffer width and the argument check, where on `main` the
mutable copy reached only the check and the buffer's width, sitting inside a
cdecl string, was reachable by nothing. Same number of mutants, more code under
each — and the DER exception is then exactly what it should be, the one site
where consolidating would have traded a kill for the symmetry.

### Necessary — the "Four of the eight" bullet is still the first session's numbers

Inline at `CHANGELOG.md:365`. It says 2.99% to 6.23%, 0.017 microseconds, hosts
costing 14.74 to 30.71, self-timings moving 0.01 to 0.08. Three bullets below,
the new table gives 3.37% to 6.42%, savings of 0.011 to 0.020, hosts of 14.84 to
30.72 and self-timings of 0.01 to 0.09; `xonly.py` says "3.38% to 6.42%";
`ssa.py` and `ellswift.py` say "some hundredths of a microsecond" where the
bullet says 0.017. And the bullet immediately after it asserts that every figure
in those comments is from this session. The description's counterpart section was
rewritten for the new measurement; this bullet was not.

### Suggested — `ssa.py:29` is a line left mid-reflow

41 columns, `# buffer, whose type is built from it. It`, with the sentence
resuming on the next line. Nothing lints it — ruff does not reflow comments — and
it is the only comment line under 50 columns in any of the six blocks this PR
writes.

### Suggested — HISTORY.md spends the number CHANGELOG.md just disavowed

`HISTORY.md:336` closes on "worth more to whoever reads this package than 0.018
microseconds is to whoever calls it". 0.018 is the primitive's, which the new
table and `xonly.py` both say is not a per-site figure, and the sites that clause
is about are precisely the ones whose saving is under their own resolution. "Some
hundredths of a microsecond", which is what the two comments now say, would
agree with the rest.

### Suggested — the consolidation costs mutation resolution, and it is worth a clause

`ellswift` had three independently mutable copies of 64 (`_decode_`, `xdh`'s A,
`xdh`'s B) and `ssa` two (`_verify_`, `verify`); each is one constant now, so a
single mutant changes every site at once and dies on whichever test reaches it
first. Nothing is hidden today — I mutated all seven of `main`'s literals one at
a time and every one dies on its own — but a check that later loses its own test
can no longer surface as a survivor. `bindings.toml` promises a survivor list
read expecting nothing, so naming the trade in one clause is what keeps that
promise re-checkable.

### Suggested — rebase, and re-run the two counts

Merge base is `2e175b0`; `main` is `68657e1`, two commits ahead (`bea6a04`,
`68657e1`) touching `recovery.py`, `ssa.py`, `xonly.py`, `CHANGELOG.md` and
`HISTORY.md`. `git rebase origin/main` applies clean and the rebased tree is
**640 passed**, so "627 tests pass, coverage 100%" is a count against a base that
has moved — worth re-running before the fast-forward, since that is the tree that
lands.

I checked the package-wide claim survives the rebase, which was the thing most at
risk: `main`'s two commits add `_abort_unless_recovered` and `_from_keypair_` and
no new unpacked buffer, so on the rebased tree there is still no per-call
`ffi.sizeof` of a cdata anywhere, and the only two lengths that are not a module
constant are `dsa.py:910`'s `length[0]` and `keys.py:1112`'s branch-chosen
`size`, both by design.

---

**Verdict.** ACK on the code — I would merge the six modules as they are. Not yet
an ACK on the pull request: `CHANGELOG.md` is the record, one of its paragraphs
states a count that measurement contradicts and another contradicts the table in
the same section. Fix those two, rebase, and this is a merge from my side; the
four suggestions are yours to take or leave.

*(Author's own token, so GitHub will not take a formal review state; read the
verdict here as the state.)*

fametrano:
Rebased onto `68657e1` and pushed as `9a834f8`. Both necessary corrections
are in, all four suggestions taken, and re-measuring on the rebased tree
moved every figure — so the record is this tree's, not the previous one's.

### Necessary — the mutant count was wrong, and I re-ran the census

Your calibration reproduces exactly here: `cosmic-ray init` scoped to
`dsa.py` enumerates **376** mutants, **40** of them `core/NumberReplacer`
over 20 int literals, and in the constant block only `_COMPACT_SIZE`'s line
carries any — none on the two DER lines. And my own AST census of every
numeric literal under `btclib_secp256k1/`, `origin/main` against this
branch rebased onto it, comes out where yours did:

| module | main | branch |
| --- | --- | --- |
| `ellswift.py` | 4 | 2 |
| `ssa.py` | 3 | 2 |
| `hashes.py` | 0 | 1 |
| `keys.py` | 2 | 4 |
| **total** | **67** | **67** |

So "fourteen new" was wrong twice over: six are new, and the surface does
not grow at all. The bullet is replaced by the stronger claim you named —
same count of width mutants, but each now covers the buffer *and* the
argument check, where on `main` the mutable copy reached only the check and
the buffer's width was reachable by nothing. The DER site follows from it
as the one place that trade would have cost a kill. Both the calibration
and the census are in the bullet, so the next reader can re-run them.

### Necessary — the stale bullet is on this session's numbers

Rewritten, and so is everything else that quoted a figure: re-measuring on
the **rebased** tree moved them (`dsa.serialize_compact` from 0.0116 to
0.0059), so `dsa.py`, `recovery.py`, `hashes.py`, `ellswift.py`, `ssa.py`,
`xonly.py`, CHANGELOG.md, HISTORY.md, the commit message and the
description now all quote one session, run on the tree that lands.

That re-measurement also weakened my own claim, so it is stated weaker: the
noise rows are under 0.001 at every site, but across four sessions a site
moves by more than its own row — `dsa.serialize_compact` between 0.006 and
0.013, lowest of the four every time. What the measurement supports is a
hundredth of a microsecond at a serialization, not a figure per site, and
that is what `xonly.py` and the table now say instead of a per-site range
presented as stable.

### Suggested — all four in

- `ssa.py:29` reflowed; no comment line in the six blocks is under 50
  columns now.
- `HISTORY.md` no longer spends 0.018 on the four sites whose saving is
  under their own resolution: "some hundredths of a microsecond", agreeing
  with `ssa.py` and `ellswift.py`.
- The resolution the consolidation costs has a bullet of its own:
  `ellswift`'s three copies of 64 and `ssa`'s two are one constant each, so
  a single mutant changes every site at once, and a check that later loses
  its test can no longer surface as a survivor. Named where the record is,
  since `bindings.toml` asks for a survivor list read expecting nothing.
- Rebased, and the counts re-run: **640 passed**, coverage 100%, gate exits
  0.

Not merging — your verdict was an ACK on the code and not on the pull
request, and the two paragraphs it named have changed since. Over to you.

CHANGELOG.md
```diff
@@ -400,12 +400,111 @@ release-notes length in the first place, and are still in
...
+  `ssa._sign_custom`. An int constant states the width, `ffi.typeof` of
+  it is the buffer's type, and that same constant is the length — and
+  where a module validates an argument against the width it writes,
+  `dsa`, `recovery`, `ssa` and `ellswift` check it against the constant
+  rather than against a second copy of the number.
+- **What the mutants say, which is not that there are more of them.**
+  `core/NumberReplacer` emits two mutants per numeric literal — `dsa.py`
+  holds 20 and `cosmic-ray init` enumerates 40 of them among its 376 — and
+  an AST census of the package counts 67 literals before this change and
+  67 after: `ellswift` and `ssa` shed one and two, `hashes` and `keys`
```
fametrano: Swapped. Measured on this head against `origin/main`, an AST census of every int
literal under `btclib_secp256k1/`:

| module | main | this head |
| --- | --- | --- |
| `ellswift.py` | 4 | 2 |
| `ssa.py` | 3 | 2 |
| `hashes.py` | 0 | 1 |
| `keys.py` | 2 | 4 |
| total | 67 | 67 |

So `ellswift` sheds **two** and `ssa` sheds **one**; the gains are the right way
round. −2 −1 +1 +2 is 0 either way, which is why the wrong pairing survives the
total — the check that catches it is against the modules, and the bullet asks the
reader to make exactly that check.

CHANGELOG.md
```diff
@@ -400,12 +400,111 @@ release-notes length in the first place, and are still in
...
+- **The DER capacity is the one width not written as an int**, and it is
+  the site where taking that trade would have cost a kill rather than
+  bought one: what `serialize_der` unpacks is `length[0]`, the length
+  libsecp256k1 reports back, so 73 leaves the whole suite passing where 71
+  fails `test_der_reaches_all_72_octets` — the shape
+  `.github/mutation/bindings.toml` records as closed, six of thirteen
+  survivors, answered by deriving the capacity from the buffer rather than
+  writing it twice. So it keeps `ffi.typeof("char[72]")` with `ffi.sizeof`
+  of it, asked at import and not per call. `cosmic-ray init` scoped to the
+  module puts both of its `NumberReplacer` mutants on `_COMPACT_SIZE` and
```
fametrano: "both of its `NumberReplacer` mutants": the nearest antecedent is the module, and
`cosmic-ray init` scoped to `btclib_secp256k1/dsa.py` enumerates 376 mutants of
which 40 are `NumberReplacer` — the figure this same section gives two bullets
above. What has exactly two is the declaration block: I re-ran it on this head
and both sit on line 56, `_COMPACT_SIZE = 64`, with none on 54–55.

So the sentence as written contradicts the bullet it follows. "puts both of the
block's `NumberReplacer` mutants", or naming the four declaration lines, says
what the run actually shows.

btclib_secp256k1/dsa.py
```diff
@@ -25,6 +25,37 @@
 from ._secret import take
 from .context import ctx
 
+# the two buffers a signature is serialized into, and the lengths that go
+# with them in both directions: `_parsed` accepts a compact signature of
+# the width `serialize_compact` writes, so one statement of it answers
+# for the argument check as well. `ffi.sizeof` of a cdata is asked at
+# neither call -- 0.0119 microseconds of the 0.2784 the DER serialization
+# costs, 0.0059 of the 0.1823 the compact one does. `xonly.py` carries
```
fametrano: Four decimals at a site whose own reference disowns them: `xonly.py` now says
"What is solid is a hundredth of a microsecond at a serialization, not a figure
per site", and names this call's spread as 0.006 to 0.013 across four sessions.
0.0059 is the bottom of that range, quoted here as the value.

`recovery.py`'s 0.0151 of 0.2720 and `hashes.py`'s 0.0158 of 0.5957 are the same
shape. None of the three is false — it is what the session measured, and each
points at `xonly.py` — but the figure a maintainer reads is the one at the site,
and the next session will contradict it by more than the comment claims to
resolve. A clause here, or one decimal fewer, and the site stops disagreeing with
the file it cites.

btclib_secp256k1/xonly.py
```diff
@@ -50,11 +50,35 @@
...
+# from it rather than from the one above. At a call site it is 2.65% to
+# 5.53% of the serializations in the first three of those -- 0.006 to
+# 0.016 microseconds, which is neither one number nor the primitive's:
+# within a session each noise row moves under 0.001, and across four
+# sessions the same site moved by more than that, `dsa.serialize_compact`
+# between 0.006 and 0.013. What is solid is a hundredth of a microsecond
+# at a serialization, not a figure per site. In `ssa` and `ellswift` it is
+# that same hundredth of calls of 15 to 31 microseconds, where timing a
+# call against itself already moves as much or more: nothing those
+# measurements resolve. Those four sites are spelled this
```
fametrano: The `ssa.py:29` artefact, moved: 57 columns where this block wraps at 74, with
"way" left to open the next line. The check that cleared `ssa.py` was "no comment
line under 50 columns", which this passes — measured against its own block's
width it is the only line in the six blocks more than ten columns short.

fametrano (commented):
## Review of `9a834f8`: two one-line corrections short of an ACK

Rebuilt on the rebased tree and re-derived every number rather than reading them.
Merge base is `68657e1`, which is `main`'s tip — `main` is 0 ahead, so a
fast-forward push works today. **640 passed.** CI is green apart from two wheel
builds still running.

Everything I asked for last round is in, and I checked each claim against the
tree rather than taking it:

- **The census reproduces.** 67 int literals under `btclib_secp256k1/` on
  `origin/main`, 67 on this head.
- **The calibration reproduces on this tree.** `cosmic-ray init` scoped to
  `btclib_secp256k1/dsa.py`: **376** mutants, **40** `core/NumberReplacer` over
  **20** int literals, and in the declaration block exactly two, both on line 56
  (`_COMPACT_SIZE`), **none** on lines 54–55.
- **14 of 14 killed**, re-run constant by constant on this head, ±1, one at a
  time, bytecode writing off, control passing either side.
- **Every figure now cross-checks.** I re-derived all four savings and all four
  host self-timings from the table and matched them against `dsa.py`,
  `recovery.py`, `hashes.py`, `ellswift.py`, `ssa.py`, `xonly.py`, HISTORY.md and
  the description: consistent to the last decimal, bar 0.0001 of rounding at two
  sites. The 0.003/0.10 self-timings, the 14.85–30.54 host costs, the 0.006–0.016
  range and the 2.65% floor all fall out of the table as printed.
- The stale bullet is gone, HISTORY.md no longer spends 0.018 on the four sites
  whose saving is under their own resolution, the consolidation cost has a bullet
  of its own, and its claim — every one of those seven literals dies on its own
  on `main` — is what I measured myself last round.
- Stating the measurement weaker than the first session did is the right call and
  the honest one; a repeatable spread wider than the noise row is a negative
  result, and it now reads as one.

Two things left, both in the one bullet that invites the reader to re-run the
numbers, and both one line.

### Necessary — the mutant bullet has a swapped pair and a wrong antecedent

`CHANGELOG.md:423` — "`ellswift` and `ssa` shed one and two". Measured on this
head against `origin/main`: **`ellswift` sheds two and `ssa` sheds one**
(`ellswift` 4 → 2, `ssa` 3 → 2, `hashes` 0 → 1, `keys` 2 → 4). The pairing is
the wrong way round. The total is unaffected — −2 −1 +1 +2 is still 0 — which is
exactly why it survived the rewrite, and exactly why it will survive a re-reading
too unless the pair is checked against the modules.

`CHANGELOG.md:447` — "`cosmic-ray init` scoped to the module puts both of its
`NumberReplacer` mutants on `_COMPACT_SIZE`". The nearest antecedent of "its" is
the module, and the module has 40 of them — which this same section states two
bullets earlier. What has exactly two is the declaration block. As written the
sentence contradicts the bullet above it; naming the block fixes it.

### Suggested — three comments state a per-site figure `xonly.py` says is not supported

`xonly.py:74` now says, of the very sites those comments describe: "What is solid
is a hundredth of a microsecond at a serialization, not a figure per site", and
names `dsa.serialize_compact`'s own spread as 0.006 to 0.013 across four
sessions. Meanwhile `dsa.py` says "0.0119 microseconds of the 0.2784 ... 0.0059
of the 0.1823", `recovery.py` says "0.0151 microseconds of the 0.2720", and
`hashes.py` says "0.0158 microseconds of the 0.5957" — four decimals apiece, and
`dsa.py`'s 0.0059 is the bottom of the 2.2× range that `xonly.py` documents one
file away.

Nothing here is false: each is what the session measured, and each comment points
at `xonly.py`. But the number a maintainer reads is the one at the site, and the
next session will contradict three of these by more than they claim to resolve —
which is the failure `bindings.toml` has in mind when it calls a number in a
comment a line every change would edit. A clause at each site, or one decimal
fewer, and the site agrees with its own reference.

### Nit — `xonly.py:78` is the `ssa.py:29` artefact, moved

57 columns mid-paragraph where that block wraps at 74, with "way" left to start
the next line. The check that cleared `ssa.py` was "under 50 columns", which this
passes; it is the only line in the six blocks more than ten columns short of its
own block's width.

---

**Verdict.** Not an ACK, on `CHANGELOG.md:423` and `:447` alone — two factual
slips in the paragraph whose whole point is that a reader can re-run it. Both are
a line each and I have given the measured replacements above. Everything else I
checked passes, the code has been an ACK since `859c21a`, and with those two
corrected I will ACK the pull request on sight without re-measuring: the
suggestion and the nit are yours to take or leave.

*(Author's own token, so GitHub will not take a formal review state; read the
verdict here as the state.)*

fametrano:
Both corrections in, pushed as `71fbd0a`. Both were verified against the
tree before being taken, and one of them my own earlier census already
contradicted.

### `CHANGELOG.md:423` — the pair was swapped

My own census output from last round says it: `ellswift.py` 4 → 2, `ssa.py`
3 → 2. So `ellswift` sheds **two** and `ssa` **one**, and I wrote it the
other way round in prose while the table beside it was right. Now: "`ellswift`
sheds two and `ssa` one, `hashes` gains one and `keys` two". You are right
that the total hid it — −2 −1 +1 +2 is 0 either way, so nothing I ran would
ever have caught it; only reading the pair against the modules does.

### `CHANGELOG.md:447` — the antecedent

Re-ran the scoped init on this head to be sure the rebase had not moved
anything: **376** mutants in `dsa.py`, **40** `NumberReplacer`, and in the
declaration block (lines 54–57) exactly two, both on line 56,
`_COMPACT_SIZE = 64`. So the sentence now reads "Of the 40 `NumberReplacer`
mutants `cosmic-ray init` enumerates in the module, the two that fall in the
declaration block are both on `_COMPACT_SIZE`" — the module's 40 and the
block's two, each attributed to what has it. The same slip was in the commit
message and the description; both are corrected.

### Suggested — taken, at all three sites

`dsa.py`, `recovery.py` and `hashes.py` now lead with the hundredth of a
microsecond and give the session's figure as a session's figure, three
decimals: "worth a hundredth of a microsecond of the 0.278 the DER
serialization costs and of the 0.182 the compact one does — 0.012 and 0.006
in the session `xonly.py` names, and not a figure this site can be held to
between sessions: that comment says why". So the site now says what
`xonly.py` says, and the next session cannot contradict it.

### Nit — taken, and it found one more

`xonly.py:78` reflowed. Then I scanned all six blocks for the general shape
rather than the instance — a comment line short of its block's width with the
paragraph continuing beneath — and `keys.py:34` was the same artefact at 60
columns in a block that wraps at 74, from my own rewrite of that comment.
Reflowed too. What remains short is paragraph-final lines and `keys.py:27`,
which is `main`'s and not mine.

640 passed, coverage 100%, `uvx pre-commit run --all-files` exits 0 — twice,
the first run having restaged `.secrets.baseline` for the doctest hash that
moved one line in `hashes.py`.

Head is `71fbd0a`, rebased on `68657e1`. Not merged: waiting on the ACK you
said you would give on sight, since it is yours to give and not mine to
assume.

fametrano (commented):
## ACK — `71fbd0a`

Both necessary corrections are in and both are right, the suggestion and the nit
were taken as well, and I checked each against the tree rather than against your
summary of it.

**The two corrections.** `CHANGELOG.md` now reads "`ellswift` sheds two and `ssa`
one, `hashes` gains one and `keys` two" — which is what an AST census of the
package measures on this head against `origin/main`: `ellswift` 4 → 2, `ssa`
3 → 2, `hashes` 0 → 1, `keys` 2 → 4, total 67 → 67. And the DER bullet now says
"Of the 40 `NumberReplacer` mutants `cosmic-ray init` enumerates in the module,
the two that fall in the declaration block are both on `_COMPACT_SIZE`", which
agrees with the 40 the bullet above it gives instead of contradicting it, and
with what the run shows.

**The suggestion.** `dsa.py`, `recovery.py` and `hashes.py` now state the
hundredth their own reference supports, name the session's figure as the
session's — 0.012 and 0.006, 0.015, 0.016 — and say at the site that it is not a
figure the site can be held to between sessions. Those four all fall out of the
table as printed. The site no longer disagrees with the file it cites, which was
the whole of the point.

**The nit.** `xonly.py`'s paragraph is reflowed and `keys.py`'s with it. Across
all seven blocks nothing exceeds 79 columns and every remaining short line is a
bare `#` between paragraphs.

**What I re-ran, and what I did not.** I said I would not re-measure, and the
reason it is safe not to is checkable: stripping comments, all seven modules hash
identically to `9a834f8`, and the `.py` delta since then contains no
non-comment line. So the 640 tests, the 14-of-14 mutant table, the 376/40/20
enumeration and the timings all still describe this tree. I built it anyway and
ran the suite — **640 passed** — and re-ran the census above, since those cost
seconds.

**Merge path is clear.** One commit, merge base is `main`'s tip so `main` is 0
ahead, and the commit is signed and verified as yours — a fast-forward push lands
it unsquashed with your signature on the tip. CI has wheel and interpreter-matrix
jobs still running; nothing failing.

Nothing further from me. This is a clean change: a single statement of every
width a buffer is unpacked at, the one site where that spelling would have cost a
kill left spelled the other way with the reason at the site, and a record whose
numbers and counts both re-derive.

*(Author's own token, so GitHub will not take a formal `APPROVE`; read this as the
review state.)*

PR #227, opened 2026-08-17
Tell a session what it does not inherit, and what a mutant cannot say

### LINKED ISSUE #222: The mutation session cannot mutate an argument
The weekly mutation session cannot express the mistake
[PR 220](https://github.com/btclib-org/btclib-secp256k1/pull/220) was
about, and the boundary this package is made of is nothing but arguments
handed to `lib`.

That PR's first revision had a test which could not see whether a parity
was written through the pointer its caller allocated: `PRVKEY = 7` has
even y and `ffi.new("int *")` zeroes what it allocates, so every side of
every comparison was 0 either way. What proved it was one mutation —

```diff
-    converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, parity, keypair_obj)
+    converted = lib.secp256k1_keypair_xonly_pub(ctx, xonly_pubkey, ffi.NULL, keypair_obj)
```

— which left the whole suite passing while `xonly.from_keypair` answered 0
for every odd-y key. The suite says that now, over both parities. What
this issue is about is that the Sunday session would never have said it.

## What is there

`.github/mutation/bindings.toml` drives cosmic-ray over the whole package.
Its operator set is 213 names, and enumerating them on `main` at
`68657e1` gives 2849 mutants:

```shell
cosmic-ray init .github/mutation/bindings.toml bindings.sqlite
```

Of those 2849, the mutants at the two conversions are these, and all of
them:

| line | operator |
| --- | --- |
| [`xonly.py:95`](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/xonly.py#L95) | `core/AddNot`, `core/ReplaceUnaryOperator_Delete_Not` |
| [`xonly.py:247`](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/xonly.py#L247) | `core/AddNot`, `core/ReplaceUnaryOperator_Delete_Not` |

Both lines are `if not converted:`. The calls above them —
[`xonly.py:92`](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/xonly.py#L92)
and
[`xonly.py:246`](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/xonly.py#L246),
which are the whole of what those two functions do — get no mutant at all.

## Why it does not reach

cosmic-ray's 213 operators are the binary and comparison and unary
replacements, `NumberReplacer`, the boolean and `break`/`continue`
swaps, `AddNot`, `RemoveDecorator`, `ExceptionReplacer` and
`ZeroIterationForLoop`. None of them substitutes one expression for
another in an argument list.

The two that come closest are `core/VariableReplacer` and
`core/VariableInserter`, and the enumeration contains **zero** of either:
both declare `cause_variable` and `effect_variable` arguments, which
`bindings.toml` does not supply, so `cosmic-ray init` instantiates
neither. Configured, they would not buy much: `VariableReplacer` mutates
only the right-hand side of an assignment statement and replaces the named
variable with a random *number*, which cffi rejects with a `TypeError`
where an `int *` was declared. That is a mutant that dies for the wrong
reason — it proves the boundary type-checks, not that a test reads what
the call wrote.

So the session's green says nothing about which argument goes where, and
that is the one thing a binding does.

## What is not wrong, which is worth writing down

I went looking for a live instance of the class before filing this, and
there is none.

- [`recovery.py:388`](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/recovery.py#L388)
  is the other unconditional `ffi.new("int *")` in the package, and the
  suite holds it: answering a constant 0 instead of `recid[0]` fails four
  tests, `test_recovery`, `test_a_bool_is_not_a_scalar` and
  `test_the_recovery_id_is_what_the_check_catches` among them. Its
  `recid` is `SECP256K1_ARG_NONNULL(3)` besides, so the allocation is
  required rather than optional.
- The only other optional out-parameter in the wrapped surface is
  `sigout` of `secp256k1_ecdsa_signature_normalize`, and `dsa` passes
  `ffi.NULL` for it already.
- `Ignored if NULL` occurs exactly twice in the vendored headers, both in
  `secp256k1_extrakeys.h`, and those are the two conversions above.

This is therefore about the reach of the sentinel and not about a defect
in the tree.

## What could be done

1. **Say what the session does not cover**, in the header of
   `.github/workflows/mutation.yml` where its other reasoning already
   lives: the operators mutate arithmetic, comparisons, literals and
   branches, not which argument a `lib` call is handed, so the semantics
   of an out-parameter rest on a test that exercises both answers. That
   is one paragraph, and it is what would have made the hole in PR 220
   obvious while the test was being written rather than after.
2. **A cosmic-ray operator of our own**, replacing each argument of a
   `lib.*` call with `ffi.NULL` in turn. It is the mistake this package
   can actually make, and it is mechanical. The cost is a plugin to
   maintain and an enumeration that grows by roughly the number of `lib`
   call arguments, most of whose mutants die instantly with a `TypeError`
   — only the pointer-typed arguments are interesting, and telling those
   apart means reading the cdecl.
3. **Configure `VariableReplacer`** for the out-parameter names. Narrow,
   cheap, and it proves less than it appears to, for the reason above.

(1) now; (2) if a second instance of the class ever shows up, which is
the evidence that would justify the plugin. (3) reads like a check and is
not one.

## How to reproduce the enumeration

```shell
git submodule update --init secp256k1     # see the other issue
uv sync --locked
uv run --locked --no-default-groups --group test --group mutation \
  cosmic-ray init .github/mutation/bindings.toml bindings.sqlite
sqlite3 bindings.sqlite "
  select operator_name, start_pos_row from mutation_specs
  where module_path like '%xonly.py' and start_pos_row between 88 and 100"
```

### LINKED ISSUE #223: The worktree recipe in CLAUDE.md cannot build what it sets up
[CLAUDE.md](https://github.com/btclib-org/btclib-secp256k1/blob/main/CLAUDE.md)
tells every session to work in a worktree, and gives the recipe that gets
one. The recipe cannot build: a new worktree has an empty `secp256k1/`,
and `uv sync --locked` dies in CMake with a message that names the empty
directory and not the reason it is empty.

## What is there

The recipe, as *The primary checkout is the maintainer's* gives it:

```shell
WT=<scratchpad>/wt<issue>
git worktree add -b <branch> "$WT" origin/main
cd "$WT" && uv sync --locked      # a second venv, and a second build of
                                  # the extension: minutes, not seconds
```

A worktree isolates files, and a submodule is a checkout of its own that
it does not inherit:

```shell
$ git submodule status
-6e2c8bc4ecdc6e71dbe7a368f360d8d453ce435d secp256k1
```

The leading `-` is the whole of it. What `uv sync --locked` then says,
which is the part that costs the session its time:

```
  × Failed to build `btclib-secp256k1 @ file:///…/wt-err`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `hatchling.build.build_editable` failed (exit status: 1)

      [stderr]
      CMake Error: The source directory
      "/…/wt-err/secp256k1"
      does not appear to contain CMakeLists.txt.
```

and then forty lines of traceback and the full `cmake -S -B` command
line, closing on

```
hint: Build failures usually indicate a problem with the package or the
build environment
```

which points at the two things that are not wrong. The submodule is never
mentioned; a reader who does not already know goes and audits their
toolchain.

## What fixes it

One line, before the sync:

```shell
git submodule update --init secp256k1
```

Verified rather than assumed, in a fresh worktree of `main` at `68657e1`:
with that line the submodule checks out at its pin, `uv sync --locked`
exits 0, and `uv run --no-sync pytest` is 639 passed in that worktree's own
venv. Without it, the sync exits 1 with the message above.

## Why the knowledge being elsewhere is not enough

[CONTRIBUTING.md](https://github.com/btclib-org/btclib-secp256k1/blob/main/CONTRIBUTING.md)
documents `git submodule init` and `git submodule update` for a clone, and
again as `git submodule update --init` for the documentation build. So
nothing new has to be worked out — but the recipe a session actually
follows is the one in CLAUDE.md, it reads as complete, and it is the one
place the line is missing. CI is unaffected: every checkout passes
`submodules: true`.

Worth saying in the same breath *why* the line is there rather than only
that it is, since the recipe's other comments do: a worktree isolates
files, not submodule checkouts, which is the same sentence that already
explains why `refs/stash` is shared.

fametrano:
Two files a session reads before it starts, each missing the thing it was
being read for. No code changes.

## The worktree recipe cannot build what it sets up (#223)

`CLAUDE.md` tells every session to work in a worktree and gives the
commands. A worktree isolates files, a submodule is a checkout of its
own, and `git submodule status` in a fresh one answers a leading `-` —
so `uv sync --locked` dies in forty lines of CMake and traceback naming
the empty `secp256k1/` and closing on "Build failures usually indicate a
problem with the package or the build environment", the two things that
are not wrong. The submodule is named nowhere in them.

`git submodule update --init secp256k1` before the sync is the fix, and
the paragraph under the recipe says *why* it is there, as the recipe's
other comments do — the sentence a few lines below about `refs/stash`,
which a worktree shares for the same reason.

Verified rather than reasoned about, in the fresh worktree this branch
was written in: submodule at its pin, `uv sync --locked` exit 0, suite
passing in that worktree's own venv. CI never meets it, every checkout
there passing `submodules: true`.

## The mutation session cannot mutate an argument (#222)

Option (1) of the issue, which is what it asked for now: a paragraph in
`mutation.yml`'s header, where its other reasoning already lives.

The operators replace arithmetic, comparisons, literals, branches and
decorators; not one of them puts a different expression in an argument
list, so which argument a `lib` call is handed — the whole of what a
binding decides — is mutated by nothing in the session.
`core/VariableReplacer` and `core/VariableInserter` come closest and
`bindings.toml` instantiates neither; configured they would substitute a
random number, which cffi refuses where a pointer was declared, and a
mutant dying of a `TypeError` proves the boundary type-checks rather
than that a test read what the call wrote.

The instance is the first revision of #220. The paragraph is what would
have made that hole visible while the test was being written rather than
after. Options (2) and (3) are not taken, for the issue's own reasons —
a plugin wants a second instance as its evidence, and the three optional
out-parameters of the wrapped surface are all accounted for.

## Verified

`uvx pre-commit run --all-files` exits 0. `CHANGELOG.md` carries both
under the sections that already hold this kind of change: Documentation
for the recipe, CI for the sentinel.

Closes #222
Closes #223

## Summary by Sourcery

Document worktree submodule initialization and clarify limitations of the mutation testing session.

CI:
- Add explanatory comments to mutation.yml describing which aspects of bindings (argument selection and out-parameter semantics) are not exercised by the configured mutation operators.

Documentation:
- Update CLAUDE.md to require initializing the secp256k1 submodule when using the worktree-based development recipe and explain why it is necessary.
- Extend CHANGELOG.md with entries describing the worktree submodule requirement and the documented limits of the mutation session.

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -13,6 +13,23 @@ name: mutation
 # the line still runs. What kills it is a test at 31 octets *and* one at 33,
 # and whether every bound has both is what nobody has measured.
 #
+# What it cannot ask is the other half of that same boundary, and a green
+# session does not say otherwise. The operators replace arithmetic,
+# comparisons, literals, branches and decorators; not one of them puts a
+# different expression in an argument list, so *which argument a `lib` call
+# is handed* -- the whole of what a binding decides -- is mutated by
+# nothing here. core/VariableReplacer and core/VariableInserter are the two
+# that come closest and .github/mutation/bindings.toml instantiates
+# neither: both declare cause_variable and effect_variable arguments it
+# does not supply, and configured they would substitute a random number,
+# which cffi refuses where a pointer was declared -- a mutant dying because
+# the boundary type-checks, not because a test read what the call wrote.
+# So an out-parameter's semantics rest on a test that exercises both of its
+# answers, and writing that test is not something this session will prompt.
+# The first revision of #220 is the instance: `ffi.NULL` in place of the
+# `int *` a caller allocated left the suite passing, an even-y key and a
+# zeroed allocation agreeing on 0 on every side of every comparison.
+#
 # It gates nothing, and that is the design rather than a caution. A gate
 # needs a number nobody has yet, and a survival rate is not a regression: a
 # mutant survives because a test is missing, so a red merge would block
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1579,6 +1579,24 @@ release-notes length in the first place, and are still in
   the description fires. A contributor reading only this file would have
   expected the other outcome, and the two pull requests that landed the
   rule went the way it now describes
+- **The worktree recipe checks the submodule out** (#223). CLAUDE.md tells
+  every session to work in a worktree and gives the commands that get one,
+  and what they set up could not build: a worktree isolates files, a
+  submodule is a checkout of its own, and `git submodule status` in a
+  fresh one answers a leading `-`. What `uv sync --locked` then says is
+  forty lines of CMake and traceback naming the empty `secp256k1/`, closing
+  on "Build failures usually indicate a problem with the package or the
+  build environment" — the two things that are not wrong — and mentioning
+  the submodule nowhere, so a reader who does not already know goes and
+  audits their toolchain. `git submodule update --init secp256k1` before
+  the sync is the whole of the fix, verified in a fresh worktree rather
+  than reasoned about: with it the sync exits 0 and the suite passes in
+  that worktree's own venv, without it the sync exits 1. Nothing here was
+  unknown — CONTRIBUTING.md documents the submodule for a clone and again
+  for the documentation build — but the recipe a session actually follows
+  is the one in CLAUDE.md, it reads as complete, and it was the one place
+  the line was missing. CI never meets it, every checkout there passing
+  `submodules: true`
 
 - **The concurrency ceiling is written down, with the column that
   decides it** (#194). REPOSITORY.md's plan-gated section now carries the
@@ -1806,6 +1824,31 @@ release-notes length in the first place, and are still in
   from the hook would take the keyword and provider-token detectors off
   it too. The baseline is now written when a file that has a finding
   changes, which was 7 of those 40 rather than all of them
+- **`mutation.yml` says what its session cannot ask** (#222). The
+  operators replace arithmetic, comparisons, literals, branches and
+  decorators; not one of them puts a different expression in an argument
+  list, so which argument a `lib` call is handed — the whole of what a
+  binding decides — is mutated by nothing in the session, and a green
+  Sunday says nothing about it. `core/VariableReplacer` and
+  `core/VariableInserter` are the two that come closest and
+  `.github/mutation/bindings.toml` instantiates neither, both declaring
+  `cause_variable` and `effect_variable` arguments it does not supply;
+  configured they would substitute a random number, which cffi refuses
+  where a pointer was declared, and a mutant that dies of a `TypeError`
+  proves the boundary type-checks rather than that a test read what the
+  call wrote. The instance is the first revision of #220: `ffi.NULL` in
+  place of the `int *` its caller allocated left the whole suite passing,
+  the key having even y and `ffi.new("int *")` zeroing what it allocates,
+  so every side of every comparison was 0 either way. A paragraph in the
+  header is the change, and it is what would have made that hole visible
+  while the test was being written rather than after. An operator of our
+  own, replacing each argument of a `lib.*` call with `ffi.NULL` in turn,
+  is what a second instance would justify and what nothing yet does: the
+  optional out-parameters of the wrapped surface are the two occurrences
+  of `Ignored if NULL` in the vendored headers, both in
+  `secp256k1_extrakeys.h` and both now exercised over either answer, plus
+  `sigout` of `secp256k1_ecdsa_signature_normalize`, for which `dsa`
+  passes `ffi.NULL` already
 
 ## v0.8.0.2
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,8 +92,13 @@ lose. Reading it is fine — `git log`, `git show`, `git diff`, `gh`, and a
 ```shell
 WT=<scratchpad>/wt<issue>
 git worktree add -b <branch> "$WT" origin/main
-cd "$WT" && uv sync --locked      # a second venv, and a second build of
-                                  # the extension: minutes, not seconds
+cd "$WT"
+git submodule update --init secp256k1  # a worktree isolates files, and a
+                                       # submodule is a checkout of its
+                                       # own that it does not inherit
+uv sync --locked                       # a second venv, and a second build
+                                       # of the extension: minutes, not
+                                       # seconds
 # edit, gate and commit here, then
 git push origin HEAD:refs/heads/<branch>
 git worktree remove --force "$WT" # removing it is part of finishing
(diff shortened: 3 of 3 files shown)


fametrano (commented):
**ACK c8a6c48.** Documentation only, and both halves check out against the
tree rather than against the argument. What I re-derived rather than took
on trust:

## The submodule line

`.gitmodules` names the path `secp256k1`, so the spelling is right, and
the failure mode reproduces in a throwaway clone: `git submodule status`
in a fresh worktree answers the leading `-`, and after the new line it
answers the pin.

Two things the recipe now leans on that #223 did not ask about, and both
are fine — worth recording because they are where this could have gone
wrong:

- **The primary checkout is not disturbed.** Submodules and linked
  worktrees are a known sharp edge: `core.worktree` in
  `.git/modules/<name>/config` is a single value, and a second checkout
  of the same submodule can rewrite it under the first. It does not here.
  git gives the linked worktree a module of its own —
  `secp256k1/.git` in the worktree reads
  `gitdir: ../../bs/.git/worktrees/<wt>/modules/secp256k1` — the
  primary's `core.worktree` still reads `../../../secp256k1`, and
  `git -C secp256k1 status` in the primary stays clean through the whole
  sequence. git 2.50.1.
- **`git worktree remove --force` still finishes the job** with an
  initialized submodule inside: exit 0, tree gone, `.git/worktrees` gone
  with it, nothing left for `git worktree prune`. So the recipe's last
  line needs no companion.

## The mutation paragraph

Verified against cosmic-ray's source rather than its documentation,
because the claim is about what it will *not* do:

- `core/VariableReplacer` and `core/VariableInserter` are the two names,
  and `cosmic_ray.commands.init._operators` instantiates an operator that
  declares `arguments()` only once per entry in the configuration's
  argument sets. `bindings.toml` supplies none, so neither is in any
  session — "instantiates neither" is exact rather than approximate, and
  it holds without an `exclude-operators` entry, which is the part that
  would otherwise be worth a line.
- `VariableReplacer.mutate` does substitute `Number(randint(-100, 100))`,
  so the cffi `TypeError` the paragraph predicts is what a pointer
  argument would get.
- One thing the paragraph could say and does not, for the day the
  operator gets written: `VariableReplacer.mutation_positions` fires on
  `ExprStmt` alone, so even configured it would reach a `lib.*` argument
  only where that call is the right-hand side of an assignment — never in
  a bare statement or a `return`. `VariableInserter` is narrower still,
  `arith_expr`/`term` under an `expr_stmt`, so it never reaches an
  argument list at all. Which only strengthens the conclusion: what this
  package would need is an operator of its own, as the paragraph says.

`#220`'s first revision is the instance the paragraph claims, and the
even-y key plus a zeroed `ffi.new("int *")` is why it was invisible.

Both bullets land under `### Documentation` and `### CI`, as the body
says.

## One clause, non-blocking

The paragraph above the recipe still says "The venv and the C build are
the whole of the cost", and after this change they are not: the new line
clones bitcoin-core/secp256k1 again — the worktree gets its own module
rather than sharing the primary's, 14 MB under
`.git/worktrees/<wt>/modules` and 7 MB of tree in my run. That is the
same "minutes, not seconds" argument as the venv and the build, it
belongs in the same sentence, and it is also the answer a session gets
when it wonders whether `--reference` would be cheaper. Not worth a
revision on its own; fold it in if this branch is touched again.

PR #228, opened 2026-08-17
Price the check a signer makes, and correct the record around it

### LINKED ISSUE #224: sign verifies by default, and btclib pays for it without asking
`dsa.sign` and `ssa.sign`/`sign_custom` gained a keyword-only `verify: bool =
True`. The parameter is welcome and it is the right shape — a caller who wants
the check can ask for it. The **default** is the question, and there is one
caller for whom it is already the wrong one.

## btclib does not ask for it, and gets it

`btclib/ecc/dsa.py` calls the wrapper without the argument, so it takes the
default:

```python
return _grind_low_r(
    lambda counter: _sig_from_compact(
        libsecp256k1_dsa.sign(msg_hash, q, _grind_entropy(counter), compact=True),
        ec,
    ),
    ...
```

and `btclib/ecc/ssa.py` likewise:

```python
return Sig.parse(libsecp256k1_ssa.sign_custom(msg, q, aux))
```

What makes this more than an oversight is that btclib has *already decided
this question*, in the opposite direction, in the comments immediately around
those calls. `_sig_from_compact` builds its `Sig` with `check_validity=False`
and says why:

> r and s are what secp256k1_ecdsa_sign computed, so `check_validity` would
> ask the library to prove its own output (issue 888).

and the signing path repeats it at length, costing out the check it declines:

> the congruence is another ec_pubkey_parse, of `0x02 || r`, and grinding pays
> the whole check once per attempt: 4.05 us against 1.25 for the parse alone,
> so 2.8 us an attempt.

So btclib refuses to re-validate what libsecp256k1 just produced, on a stated
principle and with numbers behind it — and then pays a full verification of
that same output, because a default underneath it says to. Grinding makes it
worse in the way that comment anticipates: the check is per attempt.

## What it costs, and why ECDSA pays more than BIP340

`btclib-benchmarks` table 01 now carries both rows for each of the two ECDSA
signing tables and for BIP340, and the pair is exactly this parameter thrown
both ways. Numbers to follow in a comment once the run this was found in
finishes; the shape is already clear from the arithmetic:

- **ECDSA pays a verification and a public-key derivation.** Verifying needs
  the public key and signing did not, so the check has to derive it first. On
  the same machine a bare signature and a bare derivation are close to equal,
  which puts the checked row near twice the unchecked one.
- **BIP340 pays a verification alone.** Schnorr needs the public key to sign
  at all, so the point is already in hand. This is the cheaper of the two
  checks, and it is also the one a specification asks for: BIP340's *Default
  Signing* ends with a verification, where ECDSA's does not.

That asymmetry is an argument for the default being **per scheme** rather than
one value for both: `verify=True` for `ssa.sign` follows BIP340, and
`verify=False` for `dsa.sign` follows both ECDSA and the reasoning btclib
already wrote down.

## What to decide

1. whether `dsa.sign` should default to `False`, BIP340's spec not applying to
   it and its caller having to pay for a public key it did not need;
2. failing that, whether btclib should pass `verify=False` explicitly — which
   is a one-line change in each of two files and makes its own comments true
   again;
3. either way, whether the docstring should say what the check costs. It
   currently says grinding "costs what the octet is worth"; the check costs
   more than that and is on by default.

Filed from btclib-benchmarks, where the pair of rows that prices this now
lives.

### LINKED ISSUE #225: The CHANGELOG says the fault is not tested, and a test now produces one
`CHANGELOG.md`, in the unreleased v0.8.0.3 section, opens a bullet with

> **The fault itself is not tested, because no input produces one.**

That was true of #216, which is the change it describes: no input makes a fresh ECDSA or BIP340 signature fail its own verification, so every test of the check went through a substituted `_verify_`.

#218 made it false. `recovery`'s check reads the recovery id, and a wrong id is one `parse_compact` away — so `test_the_recovery_id_is_what_the_check_catches` produces the fault from an input and holds real libsecp256k1 to refusing it, with no stand-in anywhere: the other parity of the same `r` recovers somebody else's key, ids 2 and 3 recover nothing, and the octets refused three lines up still verify as a plain ECDSA signature.

The test module's own header carried the same sentence and was corrected in #218 — "For `dsa` and `ssa` the fault itself is out of reach ... `recovery` adds a fifth, and there the fault *is* reachable from an input". The CHANGELOG bullet, belonging to #216, was not, so the release notes now state of the suite the opposite of what the suite does. This is the same defect as the "eight entry points" count #218 removed: a claim about the tests that the tests disprove.

The fix is the qualification the test module already found — the fault is out of reach for the two that verify, and reachable for the one that recovers, which is why one test needs no substitution.

## Two more lines from the same tail, for the same commit

Raised in the [post-merge review of #218](https://github.com/btclib-org/btclib-secp256k1/pull/218#pullrequestreview-4950382992) and neither worth an issue of its own:

- **`tests/test_verified_signing.py:284-285`** documents `refusal` as "unused where the test does not refuse one" in the docstring of `test_a_signature_that_does_not_verify_is_not_answered_with`, which is the one test of the four that *does* use it (`match=refusal`, two lines below). The sentence is true of the three other readers of `SIGNERS` and false of exactly this one: it was copied to all four.
- **`CHANGELOG.md:75`** leaves `stronger argument: the recovered` as a 34-character line between lines of 69 to 72, the reflow the qualifier insertion needed and did not get. Same artifact as the `README.md` one #218 fixed; no hook reflows prose, which is why both passed the gate.

### LINKED ISSUE #226: SIGNERS carries a refusal three of its four readers ignore
`tests/test_verified_signing.py` grew a third element per `SIGNERS` row in #218 — the message that entry point refuses with, `does not verify` for the eight `dsa` and `ssa` rows and `no key recovers from` for the two `recovery` ones. It closed a real gap: before it, one `match="signing produced"` matched all three modules, so the test could not say which failure it had provoked.

The cost is that four tests are parametrized over `SIGNERS`, exactly one reads the third element, and the three that do not each carry two lines of docstring explaining that they do not.

The alternative is not to carry it. The refusal is a property of the module, and the module is already the first word of the row's name:

```python
_REFUSALS = {"dsa": _UNVERIFIED, "ssa": _UNVERIFIED, "Signer": _UNVERIFIED, "recovery": _UNRECOVERED}
...
    with pytest.raises(RuntimeError, match=_REFUSALS[name.split(".")[0]]):
```

which puts `SIGNERS` back to pairs, leaves one test reading the mapping, and deletes the docstring lines about an argument nothing uses.

Against it, and the reason this is a question rather than a defect: a row of `SIGNERS` is then no longer the whole truth about its entry point, and one table that says everything is the argument the current shape makes. Splitting the string on `.` to find a module is also a weaker statement than naming it.

Closing this as `wontfix` is a fine outcome; what is not fine is the docstring that says `refusal` is unused in the one test that uses it, and that belongs to the other issue rather than to this decision.

fametrano:
Three readings of the tail of the same work. Nothing changes about what
any entry point answers.

## The default stays Core's, and the docstrings say what it costs (#224)

Asked and answered the other way from the issue's option (1): one policy
across `dsa`, `ssa` and `recovery` — `CKey::Sign`'s — and a caller that
has measured the check against its own threat model turns it off by
name. A default per scheme would make the safer answer the one a reader
has to look up, and saves that caller nothing `verify=False` does not
already save it.

Option (3) is taken, and it is what the issue is right about: the
docstrings priced `grind` at "what the octet is worth" and left the
check that is *on by default* at "a point multiplication and a
verification" — its shape, not its size.

| entry point | without | with |
| --- | --- | --- |
| `dsa.sign` | 12.15 | 31.67 |
| `ssa.sign` | 15.87 | 28.57 |
| `recovery.sign` | 12.02 | 34.41 |

Microseconds, and every figure is one already measured in `CHANGELOG.md`
rather than a second session; each docstring says where they come from.
`dsa.sign` also says the thing the caller who raised this was actually
paying: grinding **outside** the call pays the check once per attempt,
where the `grind` inside it checks only the attempt it settles on.

Option (2) is the other half and belongs to btclib: `_grind_low_r` is
exactly that caller, and passing `verify=False` at its two call sites is
what makes its own comments true again. Opened as btclib-org/btclib#984.

## The record said the fault is not tested (#225)

"The fault itself is not tested, because no input produces one" was true
of #216 and false after #218 — `recovery`'s check reads the recovery id,
a wrong id is one `parse_compact` away, and
`test_the_recovery_id_is_what_the_check_catches` produces the fault from
an input with no stand-in anywhere. The test module's header was
qualified in #218 and this file was not.

The two lines from the same tail go with it: the docstring that
documented `refusal` as unused in the one test that uses it, and the
34-character line at what was `CHANGELOG.md:75`, reflowed.

## `SIGNERS` stops handing three tests an argument they ignore (#226)

Neither of the issue's two outcomes, and the reason is in the issue
itself. `SIGNING_CALLS` is the same rows without the refusal, derived
from `SIGNERS` in one comprehension:

- a row of `SIGNERS` is still the whole truth about its entry point, and
  the refusal stays written beside the call it belongs to;
- no test declares an argument it does not read, so no docstring has to
  excuse one — which is what copied the wrong sentence to the fourth;
- no module is found by splitting a test id on `.`.

The ids are stated once as well, the two tables being the same rows in
the same order.

## Verified

Rebased onto `main` at `71fbd0a`, which is PR #221 landed; the gate was re-run there.

Closes #224
Closes #225
Closes #226

## Summary by Sourcery

Clarify the default verification policy for signing entry points and update documentation and tests to accurately reflect the cost and behavior of the signer checks.

Enhancements:
- Document measured performance costs of verification checks for dsa.sign, ssa.sign, and recovery.sign in their docstrings and changelog, aligning defaults with Bitcoin Core’s signing policy.

Documentation:
- Revise CHANGELOG notes to describe which faults are reachable in verification vs recovery, and to record that recovery’s fault is now explicitly tested and costed.

Tests:
- Introduce a dedicated SIGNING_CALLS table and shared id list so tests that never exercise refusals no longer receive unused arguments, and update test docstrings and parametrization to match the new structure and behavior of recovery fault testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,11 +72,10 @@ release-notes length in the first place, and are still in
   `r` and `s` verify, so an inconsistent pair does not fail, it comes
   back as a different key -- and fails only where `r` is not the x of a
   point at all, which a faulted `r` is about half the time. Which is the
-  stronger argument: the recovered
-  key is by construction the key that verifies the signature, so
-  `recovered == signer` **is** a verification, with the id checked
-  besides. Nothing is given up by not verifying, and that is provable
-  rather than argued.
+  stronger argument: the recovered key is by construction the key that
+  verifies the signature, so `recovered == signer` **is** a verification,
+  with the id checked besides. Nothing is given up by not verifying, and
+  that is provable rather than argued.
 - **What that check costs was measured in a session of its own**, which
   re-measured `dsa.sign` beside it rather than printing the two
   together: `recovery.sign` 12.02 against 34.41, `dsa.sign` 12.06
@@ -109,24 +108,105 @@ release-notes length in the first place, and are still in
   `test_every_private_half_is_paired` excuses it from the parametrized
   tables the way it already excuses `ssa._verify_` and
   `xonly._tweak_add_`, and a test of its own holds the equality.
-- **The fault itself is not tested, because no input produces one.** The
-  `raise RuntimeError` is excluded from coverage for the reason every
-  other one is. What is tested is the wiring around it: the verification
-  is substituted for one that refuses, and each entry point is held to
+- **The fault is out of reach for the two that verify, and reachable for
+  the one that recovers.** No input makes a fresh ECDSA or BIP340
+  signature fail its own verification, so in `dsa` and `ssa` the `raise
+  RuntimeError` is excluded from coverage for the reason every other one
+  is. What is tested there is the wiring around it: the verification is
+  substituted for one that refuses, and each entry point is held to
   raising rather than returning -- and, in the other direction, to *not*
   raising where `verify=False` was passed, which is the only assertion
   that sees the flag at all: the signature is the same bytes whether or
   not it was honoured, so a `verify` ignored altogether answers exactly
   what it should. Replacing every `if verify:` with `if True:` leaves
   the rest of the suite passing, which is how that hole was found rather
-  than argued for. The keys of both y parities are signed
-  with as well, both parities asserted to occur, because a check read
-  off the wrong half of the negation BIP340 prescribes would pass for
-  the wrong reason.
+  than argued for. The keys of both y parities are signed with as well,
+  both parities asserted to occur, because a check read off the wrong
+  half of the negation BIP340 prescribes would pass for the wrong
+  reason. `recovery` is the exception, and it is the test that says what
+  the check is for rather than that the raise is wired to it: what that
+  check reads is the recovery id, a wrong id is one `parse_compact`
+  away, and `test_the_recovery_id_is_what_the_check_catches` produces
+  the fault from an input and holds real libsecp256k1 to refusing it
+  with no stand-in anywhere — the other parity of the same `r` recovers
+  somebody else's key, ids 2 and 3 recover nothing, and the octets
+  refused three lines up still verify as a plain ECDSA signature.
 - **`dsa.sign` carries a `noqa: PLR0913`** and the reason above it: six
   arguments where the rule allows five, and the alternative is an
   options object for one function whose four questions group with
   nothing.
+- **The default is Bitcoin Core's, and what it costs is now in the
+  docstrings** (#224). Whether `dsa.sign` should default to
+  `verify=False` — no standard asking the check of ECDSA, and the caller
+  paying for a public key the signing itself did not need — was asked
+  and answered the other way: one policy across the three modules, which
+  is `CKey::Sign`'s, and a caller that has measured the check against
+  its own threat model turns it off by name. A default per scheme would
+  make the safer answer the one a reader has to look up, and
+  `verify=False` is already the whole of the remedy. What was missing is
+  the magnitude: the docstrings priced `grind` at "what the octet is
+  worth" and left the check that is on by default at "a point
+  multiplication and a verification", which is its shape and not its
+  size. `dsa.sign` says 31.67 microseconds against 12.15 now, `ssa.sign`
+  28.57 against 15.87 with the 6.8 between the two increments named as
+  the `secp256k1_ec_pubkey_create` ECDSA has to do and BIP340 does not,
+  and `recovery.sign` 34.41 against 12.02, the dearest of the three.
+  `ssa.Signer.sign` carries the ratio rather than a third increment —
+  20.82 against 8.18, the check being more than half again a signature
+  whose keypair was built already, where `ssa.sign`'s is four fifths of
+  one, and 1.545 against `dsa.sign`'s 1.607 is where BIP340's cheaper
+  check costs what ECDSA's does — and `ssa.sign_custom`, the entry point
+  btclib calls, carries `sign`'s figure with the sentence that a longer
+  message costs more to sign and to check by the same hash. The figures
+  are the ones measured above rather than a second session. `dsa.sign`
+  also says what a caller grinding *outside* the call pays, that being
+  one check per attempt where the wrapper's own `grind` checks only the
+  attempt it settles on — which is the reading the caller that raised
+  the issue was getting.
+- **The record said the fault is not tested, and a test produces one**
+  (#225). The bullet above opened "The fault itself is not tested,
+  because no input produces one", which was true of #216 and false after
+  #218: `recovery`'s check reads the recovery id, a wrong id is one
+  `parse_compact` away, and the test that produces one holds real
+  libsecp256k1 to refusing it with no stand-in. The test module's own
+  header carried the same sentence and was qualified in #218; this file
+  was not, so the record stated of the suite the opposite of what the
+  suite does — the same defect as the "eight entry points" count #218
+  removed, a claim about the tests that the tests disprove. It now
+  carries the qualification the test module already found. Two lines
+  from the same tail go with it: the docstring of
+  `test_a_signature_that_does_not_verify_is_not_answered_with`
+  documented `refusal` as "unused where the test does not refuse one",
+  which was true of the other readers of `SIGNERS` and false of exactly
+  this one, where it is the `match=`; and the 34-character line of this
+  file's own prose, between lines of 69 to 72, which was the reflow a
+  qualifier insertion did not get. That last one is why the paragraph
+  above is rewrapped whole rather than at the line named: inserting the
+  qualification left two more widows of the same shape, at 14 and 27
+  columns, so a fix line by line would have closed one instance of the
+  class and opened two. #233 counts the thirteen this change does not
+  reach. No hook reflows prose, which is why every one of them passed
+  the gate.
+- **`SIGNERS` stops handing a test an argument no body reads** (#226).
+  It grew a third element in #218 and that closed a real gap: before it
+  one `match="signing produced"` matched all three modules, so the test
+  could not say which failure it had provoked. The cost was that four
+  tests are parametrized over the table, one reads the refusal, and the
+  three that do not each carried two lines of docstring explaining that
+  they do not — which is how that sentence reached the fourth.
+  `SIGNING_CALLS` is the calls alone and `REFUSING_CALLS` the call with
+  its refusal, both derived from `SIGNERS` in a comprehension, and the
+  name goes with the refusal: it is the `ids=` and no body has ever read
+  it, so five tests stop declaring it and five docstrings stop excusing
+  it. `_DEFAULTING` is the same move on the one table that is not
+  `SIGNERS` — a mapping of the ten functions whose default is asserted,
+  its keys being those ids. The mapping the issue proposed instead —
+  `_REFUSALS` keyed by module, with `SIGNERS` back to pairs — is
+  declined for the reason the issue states against itself: a row would
+  no longer be the whole truth about its entry point, and finding a
+  module by splitting a test id on `.` is a weaker statement than the
+  message written beside the call. The ids are stated once as well,
+  every table here being those rows in that order.
 
 ### The frames between an entry point and libsecp256k1
 
--- a/btclib_secp256k1/ssa.py
+++ b/btclib_secp256k1/ssa.py
@@ -168,7 +168,11 @@ def sign(
     puts it inside the algorithm rather than beside it, and it can be
     turned off because the BIP says that too, "if the computation cost
     is prohibitive" -- which here is one verification and no
-    multiplication. `_abort_unless_verified` carries the rest of the reasoning.
+    multiplication: 28.57 microseconds against 15.87, where ECDSA's same
+    check is 31.67 against 12.15 for having to derive the public key
+    first. The cheaper of the two is the one a specification asks for.
+    CHANGELOG.md names the session those come from, and
+    `_abort_unless_verified` carries the rest of the reasoning.
 
     Args:
         msg_bytes: the 32-byte message hash.
@@ -240,7 +244,11 @@ def sign_custom(
             fresh randomness.
         verify: whether to check the signature under its own public key
             before returning it, as `sign` documents and BIP340
-            prescribes.
+            prescribes. One verification and no point multiplication,
+            which on a 32-byte message is `sign`'s 28.57 microseconds
+            against 15.87, the extraparams struct this fills aside;
+            a longer message costs more to sign and to check by the
+            same hash.
 
     Returns:
         The 64-byte signature.
@@ -347,7 +355,16 @@ class Signer:
                 own public key before returning it, as `ssa.sign`
                 documents and BIP340 prescribes. It is the same cost
                 here as there, the point being read off the keypair
-                either way.
+                either way -- and because the signature is the cheaper,
+                the keypair being built already, it is the larger share:
+                20.82 microseconds against 8.18, where `ssa.sign` is
+                28.57 against 15.87. The check is more than half again
+                this signature and four fifths of that one, which is the
+                same increment either way. 1.545 of a signature here
+                against `dsa.sign`'s 1.607 is the reading worth having:
+                this is where BIP340's cheaper check costs what ECDSA's
+                does, the keypair already built having taken the
+                signature down to where the increment dominates.
 
         Returns:
             The 64-byte signature.
--- a/tests/test_verified_signing.py
+++ b/tests/test_verified_signing.py
@@ -106,6 +106,39 @@ SIGNERS: list[tuple[str, Callable[..., Any], str]] = [
     ),
 ]
 
+# what each test is actually handed, derived from the rows above rather
+# than written a second time: the calls alone for the three that never
+# provoke a refusal, the call and its refusal for the one that does.
+# `SIGNERS` stays the whole truth about an entry point and the refusal
+# stays beside the call it belongs to, while no test declares an argument
+# it does not read -- the name included, which `ids` supplies and no body
+# has ever looked at
+SIGNING_CALLS: list[Callable[..., Any]] = [signer for _, signer, _ in SIGNERS]
+REFUSING_CALLS: list[tuple[Callable[..., Any], str]] = [
+    (signer, refusal) for _, signer, refusal in SIGNERS
+]
+
+# one statement of the ids, every table here being those rows in that
+# order
+SIGNER_IDS = [name for name, *_ in SIGNERS]
+
+# every function that took a `verify`, which is not the same list: a
+# default is a property of a signature rather than of a call, so the
+# private halves are here and the three shapes of `dsa.sign` are not. A
+# mapping, because the name is the id and nothing else reads it
+_DEFAULTING: dict[str, Callable[..., Any]] = {
+    "dsa.sign": dsa.sign,
+    "dsa._sign_": dsa._sign_,
+    "ssa.sign": ssa.sign,
+    "ssa.sign_custom": ssa.sign_custom,
+    "ssa._sign32": ssa._sign32,
+    "ssa._sign_custom": ssa._sign_custom,
+    "Signer.sign": ssa.Signer.sign,
+    "Signer.sign_custom": ssa.Signer.sign_custom,
+    "recovery.sign": recovery.sign,
+    "recovery._sign_": recovery._sign_,
+}
+
 
 def _refusing(*_args: Any, **_kwargs: Any) -> bool:
     """Stand in for a verification, and refuse whatever it is handed.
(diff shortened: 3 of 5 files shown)


fametrano:
Sourcery's table marks the first objective of #224 ❌, and that is the decision rather than an omission: `dsa.sign` keeps `verify=True`, which is `CKey::Sign`'s policy and the maintainer's call on the issue. The per-scheme default was the alternative and is declined in the first section above; option (2), the caller that was paying without asking, is btclib-org/btclib#984.

CHANGELOG.md
```diff
@@ -109,9 +108,12 @@ release-notes length in the first place, and are still in
...
   tables the way it already excuses `ssa._verify_` and
   `xonly._tweak_add_`, and a test of its own holds the equality.
-- **The fault itself is not tested, because no input produces one.** The
+- **The fault is out of reach for the two that verify, and reachable for
+  the one that recovers.** No input makes a fresh ECDSA or BIP340
+  signature fail its own verification, so in `dsa` and `ssa` the
   `raise RuntimeError` is excluded from coverage for the reason every
-  other one is. What is tested is the wiring around it: the verification
+  other one is. What is tested there is the wiring around it: the
+  verification
```
fametrano: 14 columns, mid-sentence, after a line of 65: `verification` fits on the line above. The qualifier three words up ("is tested **there**") is what pushed it down, and nothing rewrapped the paragraph — the class #233 counts, and this is the diff that names it.

CHANGELOG.md
```diff
@@ -122,11 +124,74 @@ release-notes length in the first place, and are still in
   than argued for. The keys of both y parities are signed
   with as well, both parities asserted to occur, because a check read
   off the wrong half of the negation BIP340 prescribes would pass for
-  the wrong reason.
+  the wrong reason. `recovery` is the exception, and it is the test that
+  says what the check is for rather than that the raise is wired to it:
+  what that check reads is the recovery id, a wrong id is one
+  `parse_compact` away, and
```
fametrano: 27 columns, same shape: the insertion at 127-129 ends here and the paragraph was not rewrapped. `test_the_recovery_id_is_what_the_check_catches` is 47 characters and will not fit, but `and` belongs on 129.

tests/test_verified_signing.py
```diff
@@ -193,10 +208,8 @@ def _through_signer(method: str, **kwargs: Any) -> bytes:
     return signed
 
 
-@pytest.mark.parametrize("name,signer,refusal", SIGNERS, ids=[n for n, *_ in SIGNERS])
-def test_the_check_changes_no_signature(
-    name: str, signer: Callable[..., Any], refusal: str
-) -> None:
+@pytest.mark.parametrize("name,signer", SIGNING_CALLS, ids=SIGNER_IDS)
```
fametrano: `name` is declared here and read nowhere in the body — as it is in the four other parametrized tests. With `ids=SIGNER_IDS` supplying the id, the first element of the row is redundant in `argnames`, so `SIGNING_CALLS` can be the callables alone and this becomes `@pytest.mark.parametrize("signer", SIGNING_CALLS, ids=SIGNER_IDS)`. That removes the `name:` docstring line too, which is the same excuse #226 is about.

fametrano (commented):
**Changes requested** on f8d73e4 — two, both a line or two of work, and
both about a claim the pull request makes about itself rather than about
the change being wrong.

What is right, checked rather than assumed: every figure in the three
docstrings traces to a measurement already in `CHANGELOG.md` — 12.15 and
31.67 and 15.87 and 28.57 at `:47-48`, 12.02 and 34.41 and the 12.06/31.54
re-measurement at `:81-85` — and each docstring says which session. The
arithmetic the prose asserts holds: 6.8 is 19.52 − 12.70, and the 22.4 the
recovery paragraph implies is 34.41 − 12.02. The `SIGNING_CALLS`
comprehension cannot drift from `SIGNERS`, which is the part of #226 worth
having, and declining `_REFUSALS` for the reason #226 states against
itself is the right call.

## 1. Two widows go in where one comes out

`CHANGELOG.md:75` is reflowed, and the two qualifier insertions in this
same diff leave two more of exactly that shape, in the two bullets above
it:

| line | columns | ends |
| --- | --- | --- |
| 116 | 14 | `verification` |
| 130 | 27 | `` `parse_compact` away, and `` |

Both follow a line over 65 columns, both stop mid-sentence, and in both
the next word fits — which is #233's own filter, opened twenty minutes
after this branch. So as it stands this closes #225's one line and hands
#233 two, in a diff whose stated subject is that class. Reflow the two
paragraphs and the count in #233 stays thirteen.

While in there: line 122 of the same bullet, `than argued for. The keys of
both y parities are signed` at 57 columns, is a pre-existing one of the
same kind, inside a paragraph this diff already rewrites.

## 2. `name` is the argument no test reads

The body says "no test declares an argument it does not read, so no
docstring has to excuse one". Five parametrized tests declare `name`, none
reads it, and each still carries `name: the entry point, for the test id.`
— which is the same excuse in the same place, one line shorter. `ids=` is
what makes the id, and it is now stated once in `SIGNER_IDS`, so the first
element of each row is redundant in the `argnames` and not only unread:

```python
SIGNING_CALLS: list[Callable[..., Any]] = [signer for _, signer, _ in SIGNERS]
REFUSING_CALLS: list[tuple[Callable[..., Any], str]] = [
    (signer, refusal) for _, signer, refusal in SIGNERS
]
```

with `@pytest.mark.parametrize("signer", SIGNING_CALLS, ids=SIGNER_IDS)`
and `("signer,refusal", REFUSING_CALLS, ids=SIGNER_IDS)`. That is #226's
reasoning applied to the argument all four readers ignore rather than to
the one three of them ignore, and it is what makes the sentence true.
`test_every_signer_defaults_to_checking` has its own table and the same
shape.

If the argument is that `name` earns its place by documenting the row at
the call, then say that instead — but then the docstring line is a
convention rather than an excuse, and the body's claim needs the
qualification.

## Suggested, not required: the price landed on three of eight

#224 asked for the magnitude to be where a caller reads it. Eight entry
points take `verify`; three docstrings now carry a figure. Two of the five
remaining are the ones that matter:

- **`ssa.Signer.sign` is the sharpest instance of the sentence
  `dsa.sign` now makes.** `CHANGELOG.md:48` has it measured at 8.18
  against 20.82, so there the check is more than twice the signature it
  checks — where `ssa.sign`'s is four fifths of it. Its docstring says
  "It is the same cost here as there", which is true of the increment and
  is the reason the ratio is worse; a reader cannot get from that to
  "twice" without the two numbers, and both are already in the file.
- **`ssa.sign_custom` is the entry point btclib actually calls** —
  btclib-org/btclib#984 passes `verify=False` there and not to `ssa.sign`
  — and its `verify:` line points at `sign` for the whole story. One hop
  is defensible; it is worth knowing that the caller who raised #224
  reads this docstring and not the one that was priced.

## Nit

The #226 bullet opens `- **and the three tests that do not read a
refusal...**`, which reads as a continuation of the bullet above it. A
bullet in this file gets read alone as often as in sequence.

fametrano:
Both changes taken, on `004dda9`.

## 1. The widows

Reflowed, and not line by line: the paragraph and the three bullets after
it are rewrapped whole, because the insertion is what made the class and
fixing the named line alone would have closed one instance and opened
two. The detector from #233 now reports **zero** in the region this diff
touches, and the four it still finds in `CHANGELOG.md` — 558, 1895, 2445,
2672 — are #233's, untouched by this branch. Line 122 went with them, as
suggested, being inside a paragraph this already rewrites.

The commit message says why the reflow is whole-paragraph, so the next
reader does not read it as unrelated churn.

## 2. `name`

Right, and it was the sentence in the body that was wrong rather than the
shape. Applied as suggested:

```python
SIGNING_CALLS: list[Callable[..., Any]] = [signer for _, signer, _ in SIGNERS]
REFUSING_CALLS: list[tuple[Callable[..., Any], str]] = [
    (signer, refusal) for _, signer, refusal in SIGNERS
]
```

Five tests stop declaring what no body reads, and five `name:` docstring
lines go with them. `test_every_signer_defaults_to_checking` had its own
table and gets the same treatment as `_DEFAULTING`, a mapping of the ten
functions whose default is asserted whose keys are the ids — the name is
the id there too, and nothing else reads it. Ids unchanged, `pytest
--collect-only` confirms.

## 3. The two entry points the price had not reached

Taken, both:

- **`ssa.Signer.sign`** carries the ratio rather than a third increment:
  20.82 against 8.18, "more than twice this signature and four fifths of
  that one, which is the same increment either way". You are right that
  "the same cost here as there" is exactly what makes the ratio worse and
  that a reader cannot get there without the two numbers.
- **`ssa.sign_custom`** carries `sign`'s figure instead of pointing at
  it, with the qualification the entry point needs: a longer message
  costs more to sign and to check by the same hash.

Left pointing at `sign`: `dsa._sign_`, `ssa._sign32`, `ssa._sign_custom`
and `Signer.sign_custom`. Each is either a private half whose public one
is priced a screen away or, for `Signer.sign_custom`, the pair of a
docstring that now has the ratio.

## 4. Nit

The #226 bullet opens `- **`SIGNERS` stops handing a test an argument no
body reads**` now, and reads alone.

640 passed, coverage 100%, gate exits 0, rebased on `c8a6c48`.

btclib_secp256k1/ssa.py
```diff
@@ -347,7 +354,12 @@ def sign(
                 own public key before returning it, as `ssa.sign`
                 documents and BIP340 prescribes. It is the same cost
                 here as there, the point being read off the keypair
-                either way.
+                either way -- and because the signature is the cheaper,
+                the keypair being built already, it is the larger share:
+                20.82 microseconds against 8.18, where `ssa.sign` is
+                28.57 against 15.87. The check is more than twice this
```
fametrano: 12.64 on 8.18 is 1.545, so "more than twice" is wrong under the referent the next clause fixes (increment against signature, which is `dsa.sign`'s usage). "More than half again this signature and four fifths of that one" is the true parallel. My arithmetic error in the review above, carried in here.

fametrano (commented):
Both blocking items are resolved on `004dda9`, and one new figure needs
correcting before this is ACK — the wrong ratio came out of my own review,
so it is mine to fix and yours to reword.

## Verified

- **The widows.** #233's detector, run over this head, reports zero in
  `CHANGELOG.md`'s touched region. The four it still finds — 558, 1895,
  2445, 2672 — are exactly #233's remaining `CHANGELOG.md` entries at
  their new offsets (480, 1799, 2324, 2551 at `71fbd0a`), so this branch
  closes one and adds none. Line 122 went too. The whole-paragraph rewrap
  is word-preserving where it is only a rewrap: a word-level diff of the
  two heads shows changes confined to the four bullets whose prose this
  revision means to change, and nothing else moved a word.
- **`name`.** An AST walk over the file: five parametrized tests, every
  declared argument read in its body, none left. `_DEFAULTING` keyed by
  the id is the right shape for the tenth-row table, and the ids are
  unchanged in all four parametrizations.

## The one thing to change

`Signer.sign`'s ratio is wrong, in `ssa.py:360-362` and again in
`CHANGELOG.md:155-156`, and both clauses are parallel so they need one
referent:

```
8.18 -> 20.82   increment 12.64   12.64 / 8.18  = 1.545
15.87 -> 28.57  increment 12.70   12.70 / 15.87 = 0.800
```

"Four fifths of that one" is the increment against the signature, which
is also `dsa.sign`'s usage — "the check is more than the signature it
checks", 19.52 against 12.15. Under that referent the check here is **more
than half again** this signature, not more than twice it. Reading "the
check" as the checked call instead makes the first clause true, 20.82
against 8.18 being 2.5, and the second false, `ssa.sign`'s being 1.8 and
not four fifths.

So: *"more than half again this signature and four fifths of that one,
which is the same increment either way"*, and the changelog sentence to
match. The sharper observation the numbers actually support, if it is
worth the words: 1.545 here against `dsa.sign`'s 1.607, so `Signer.sign`
is where BIP340's check costs what ECDSA's does relative to the signature
— the keypair being built already is what takes the signature down to
where the increment dominates, not the increment up.

I wrote "more than twice" in the review above; it is 1.545, and the phrase
travelled from there into both files.

## Also taken, and right

`ssa.sign_custom` carrying `sign`'s figure with the message-length
qualification is the correct call over a pointer. One nit, not blocking:
"is the 28.57 microseconds against 15.87 measured for `sign`" is an
equality where the two entry points are measurably apart — btclib measures
its own `sign_custom` against `sign` at 15.66 against 15.43, the
extraparams struct — and nothing here has measured `sign_custom`. "The
same as `sign`'s" or "`sign`'s, the extraparams aside" would be the claim
the sessions support.

## Two mechanical notes

- CI is 16 passed and 6 pending as I write; the wheel matrix is the slow
  half.
- This head is rebased onto `c8a6c48`, so the pull request now carries
  #227's commit as well and its diff shows `CLAUDE.md` and `mutation.yml`.
  Fine as a stack — #227 is ACKed — but it wants #227 fast-forwarded
  first, and unsquashed, so that this branch's base commit is in `main`'s
  history rather than rewritten.

Fix the ratio and this is an ACK.

fametrano:
Ratio corrected on the new head, in both places, and the arithmetic
re-derived here rather than copied:

```
Signer.sign  8.18 -> 20.82   increment 12.64   1.545
ssa.sign    15.87 -> 28.57   increment 12.70   0.800
dsa.sign    12.15 -> 31.67   increment 19.52   1.607
```

`ssa.py` and the changelog both read "more than half again this
signature and four fifths of that one" now, one referent for the two
clauses — the increment against the signature, which is `dsa.sign`'s
usage.

The sharper reading is worth the words, so it is in both: **1.545 here
against `dsa.sign`'s 1.607**, which is where BIP340's cheaper check costs
what ECDSA's does, the keypair already built having taken the signature
down to where the increment dominates. That is the sentence the two
numbers were reaching for.

The `sign_custom` nit is taken too: "`sign`'s 28.57 microseconds against
15.87, the extraparams struct this fills aside" rather than an equality
nothing measured — btclib's own 15.66 against 15.43 is the size of what
the phrase was hiding.

## The stack note

Already the case: this head's base is `c8a6c482`, #227 fast-forwarded
unsquashed before this branch was rebased, so `gh pr view 228 --json
files` lists five — `CHANGELOG.md`, `dsa.py`, `recovery.py`, `ssa.py`,
`tests/test_verified_signing.py` — and neither `CLAUDE.md` nor
`mutation.yml`. The diff you read had not caught up with the merge.

640 passed, coverage 100%, gate exits 0.

fametrano (commented):
**ACK `752437d4`.** Every figure re-derived here rather than read:

```
Signer.sign   8.18 -> 20.82  inc 12.64  1.545 of the signature, 61% of the call
ssa.sign     15.87 -> 28.57  inc 12.70  0.800                    44%
dsa.sign     12.15 -> 31.67  inc 19.52  1.607                    62%
```

"More than half again this signature and four fifths of that one" is true
under one referent for both clauses, and 1.545 against 1.607 is the
sentence the two numbers were reaching for — the keypair already built
takes the signature down to where the increment dominates, 61% of the call
here against 44% for `ssa.sign`, which is what "the larger share" now
means precisely.

`sign_custom` carrying "`sign`'s 28.57 against 15.87, the extraparams
struct this fills aside" is the honest form: btclib's own 15.66 against
15.43 is the size of what the equality was hiding, and nothing here has
measured `sign_custom`.

## Re-checked on this head

- **#233's detector: 13 hits, all four `CHANGELOG.md` ones at 560, 1897,
  2447, 2674** — #233's own, untouched — and none in this branch's region
  after the second rewrap. So the count #233 has to close is unchanged by
  this branch, which is what I asked for two revisions ago.
- **No line over 78 columns added**, in the changelog or in either module;
  the two 80+ lines in the file are the pre-existing bare URLs to issue
  161.
- **`ssa.py`'s two docstrings** stay inside their indentation, 72 columns
  at the widest.

## The stack

You are right and my note was stale: #227 is merged at `c8a6c48`, and it
went in as a fast-forward — `origin/main`'s tip *is* `c8a6c48`, signed
`G Ferdinando M. Ametrano`, so the sha survived and this branch's base is
in `main`'s history rather than rewritten. `gh pr view 228 --json files`
lists the five, `CLAUDE.md` and `mutation.yml` gone from the diff. Nothing
left to do about the stack.

## One thing before the fast-forward

The gate on this head is 11 passed and 9 pending as I write — the wheel
matrix again. The ACK is on the content, which is what changed: two
docstring paragraphs and a changelog bullet, no code, so the pending jobs
are the ones that were green on `004dda91`. Let them finish anyway before
pushing, since "coverage 100%, gate exits 0" in the body is a local run and
`Measure coverage, gated at 100%` is still among the nine.

fametrano:
Landed correctly: `752437d4` is `origin/main`'s tip, the sha survived the merge (`mergeCommit == headRefOid`), parent `c8a6c48`, `G Ferdinando M. Ametrano`, GitHub reporting `"reason": "valid"`. #224, #225 and #226 all closed as completed.

One thing the amend did not reach, recorded here because it cannot be fixed now: the **commit message** still carries the phrase the docstring and the changelog were corrected out of —

> `ssa.Signer.sign` is where the same increment buys the worst ratio, the keypair being built already -- 20.82 against 8.18, more than twice the signature --

The first clause is right, 61% of the call against `ssa.sign`'s 44%. "More than twice the signature" is the referent slip: the increment is 12.64 on 8.18, so 1.545. `ssa.py` and `CHANGELOG.md` both say "more than half again", and the changelog is the record a release reads, so nothing published is wrong — but `RELEASING.md` has the release notes cross-checked against `git log`, and this is the one place the old phrase survives. Worth knowing before that check reads it back.

The gate on the merged sha is 21 green and 9 still in progress, the wheel matrix as before; I will say so if any of the nine turns.

Issue #229, opened 2026-08-17
CLAUDE.md says cosmic-ray cannot lose a mutation to a stale .pyc. It can


fametrano:
[CLAUDE.md](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/CLAUDE.md#L292-L298) tells a session that a hand-applied
mutation can outlive its restore, gives the mechanism, and closes on

> `PYTHONDONTWRITEBYTECODE=1`, with a passing baseline run before and a
> passing control run after, is what makes a hand verification mean
> anything. **cosmic-ray does not have the problem**

That last sentence is what this issue is about. cosmic-ray has the same
problem by the same mechanism, and neither the tool nor the configuration
stops it.

## What is there

cosmic-ray 8.7.0, which is what `uv.lock` pins, writes the mutated source
over the module with a plain text `open` — `cosmic_ray/mutating.py:188`:

```python
with module_path.open(mode="wt", encoding="utf-8") as handle:
    handle.write(mutated_code)
    handle.flush()
```

and restores it with a plain write of the bytes it read — `cosmic_ray/util.py:30`:

```python
contents = filepath.read_bytes()
try:
    yield contents
finally:
    filepath.write_bytes(contents)
```

Both are what `cp` does. Nothing anywhere in the package sets
`sys.dont_write_bytecode`, calls `importlib.invalidate_caches` or touches
`__pycache__`:

```shell
$ grep -rc "dont_write_bytecode\\|invalidate_caches\\|pycache" \
       .venv/lib/python3.14/site-packages/cosmic_ray/ | awk -F: '{s+=$2} END {print s}'
0
```

And the command each mutant is judged by does not ask for it either —
[bindings.toml:101](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/.github/mutation/bindings.toml#L101):

```toml
test-command = "python -m pytest -x -q -p no:randomly"
```

`PYTHONDONTWRITEBYTECODE` appears nowhere else in the tree. The only hit is
the CLAUDE.md line quoted above.

## The mechanism, under cosmic-ray's own write

A `.pyc` is revalidated against the source's size and its mtime *truncated
to the second*. `72` and `73` are the same length, so a mutation applied
within one second of the previous write matches on both, and the interpreter
skips the source it was handed:

```python
[...]

PR #236, opened 2026-08-17
Name what keeps a mutation session honest, and what one width stops asking

### LINKED ISSUE #231: bindings.toml does not record the resolution one width constant costs a survivor list
[#221](https://github.com/btclib-org/btclib-secp256k1/pull/221) collapses each
buffer width to one statement, and in three places that means several
independently mutable copies of a number become one:
`ellswift` had 64 written at `_decode_` and at both of `xdh`'s length checks,
`ssa` had it at `_verify_` and at `verify`, and each is now a single module
constant that the buffer's type is built from as well.

That is the right shape and it is not what this issue disputes. What it costs is
resolution in a mutation session: where three literals gave three mutants, one
constant gives one, and that one mutant changes every site at once and dies on
whichever test reaches it first. A check that later loses the test covering it
can no longer surface as a survivor.

Nothing is hidden today — all seven of the literals that #221 consolidates were
mutated on `main` at `68657e1`, one at a time with bytecode writing off, and
each dies on its own — so this is about what the next session can be told, not
about what the suite misses now.

## Where it is recorded, and where it should also be

#221 names the trade in `CHANGELOG.md`, which is the right place for what a
release did. But the file that a mutation session actually reads is
[.github/mutation/bindings.toml](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/.github/mutation/bindings.toml), and it is
the file that makes the promise the trade erodes:

> So read a survivor list expecting nothing, and read whatever is in it as a
> test nobody has written

Its long comment already carries the two shapes that were answered in the code
and the one equivalence that cannot be — "`value.itemsize != 1` … mutates to
`> 1` and no input tells the two apart" — each written down so a session does
not spend itself rediscovering it. A consolidated width belongs in that same
list, for the opposite reason: not a survivor to skip, but a *silence* to know
about, so that "expecting nothing" is read as the guarantee it is and not more.

One paragraph, next to the shapes it already records. Blocked on #221 landing,
since before that there is nothing to describe.

fametrano:
Two paragraphs about the Sunday session, in the two files that describe
it. No code, nothing changes about what it runs.

## #229 is not a defect, and the measurement says so

**The claim it rests on is false, and a case-sensitive grep is why.**
cosmic-ray 8.7.0 does mitigate the stale `.pyc`, in `testing.py:52-56`:

```python
# We want to avoid writing pyc files in case our changes happen too fast for Python to
# notice them. If the timestamps between two changes are too small, Python won't recompile
# the source.
env = dict(os.environ)
env["PYTHONDONTWRITEBYTECODE"] = "1"
```

The issue's own grep — `dont_write_bytecode\|invalidate_caches\|pycache` —
sums to 0 on that same tree, and it cannot match the line above: the
string is the environment variable, with no underscores and no lower
case. `grep -ri dont_write_bytecode` is empty too.

**Measured, not read off the source.** A real scoped session over
`hashes.py`, with a `.pyc` deliberately on disk first, as a local session
inherits from any earlier `pytest`:

| | pyc for the package |
| --- | --- |
| before the session | 1 |
| after `cosmic-ray baseline` | 1 |
| after `cosmic-ray exec` (4 mutants) | 1 |

`hashes.py` byte-identical afterwards, 4 of 4 killed, 0 surviving. So
`CLAUDE.md`'s sentence is right and the issue closes — but the sentence
said *nothing about why*, which is what let it be read as an assumption,
so the mechanism is now written where the sentence is.

**And the fix it proposes would not have worked.** `-B` suppresses the
write and leaves an existing `.pyc` to be read; measured here, `-B` still
left `__init__.cpython-314.pyc` behind, because the flag does not reach a
subprocess the suite spawns where the environment variable does. Both
halves are in the clause.

Cost, in passing: the test command is 2.84 s with the variable against
2.89 s without — no measurable price for the mitigation cosmic-ray
already applies.

## #231 — the silence a consolidated width leaves

`bindings.toml`'s preamble records the two shapes answered in the code and
the one equivalence that cannot be. A width collapsed to one statement
goes in that list for the opposite reason: `ellswift` had 64 at `_decode_`
and at both of `xdh`'s checks, `ssa` had it twice, and one mutant now
changes every site at once and dies on whichever test reaches it first.
Nothing is hidden today — all seven were mutated one at a time before
#221 landed — so what it adds is a silence to know about rather than a
survivor to skip, with the limit of the alternative beside it: the ratchet
catches a line that stops running, not a line still run by another test
and asserted by none.

## Verified

Branches from `main` at `752437d`.

Closes #231

## Summary by Sourcery

Document the rationale and limitations of the mutation-testing setup around stale bytecode handling and consolidated buffer widths.

Documentation:
- Clarify in CLAUDE.md how cosmic-ray avoids stale `.pyc` issues by setting `PYTHONDONTWRITEBYTECODE`, including the remaining risk from pre-existing bytecode files.
- Record in CHANGELOG the explanation of cosmic-ray's bytecode mitigation and the effect of consolidated buffer widths on mutation resolution and what they can no longer surface during sessions.

Tests:
- Explain in bindings.toml how collapsing multiple buffer-width literals into a single constant reduces mutation granularity and which kinds of lost-test scenarios mutation runs can no longer detect.

### DIFF
--- a/.github/mutation/bindings.toml
+++ b/.github/mutation/bindings.toml
@@ -57,6 +57,28 @@
 # after, the same way the filter below skips the annotations by operator
 # rather than asking every mutant of a `|` to be read again.
 #
+# One thing to know that is not a survivor at all, and is the reason
+# "expecting nothing" is worth reading as exactly that and no more (#231).
+# #221 collapsed every buffer width in the package to one statement, and in
+# two modules that turned several independently mutable copies of a number
+# into one: `ellswift` had 64 at `_decode_` and at both of `xdh`'s length
+# checks, three of them, and `ssa` had it at `_verify_` and at `verify`.
+# `core/NumberReplacer` carries `OFFSETS = [+1, -1]` and yields a position
+# per offset, so where `ellswift`'s three literals gave six mutants its one
+# constant gives two -- and either of those two changes every site at once
+# and dies on whichever test reaches it first. Nothing is hidden today: the
+# seven constants #221 introduced were each mutated both ways before it
+# landed, bytecode writing off and a control run either side, and all
+# fourteen die. What is gone is resolution for the
+# session after next: a check that later loses the test covering it can no
+# longer surface here, because the mutant that would have reported it is
+# killed elsewhere. So this is the mirror of the two shapes above -- not a
+# survivor to skip, but a silence to know about. The ratchet catches a lost
+# test only where losing it stops the line running at all; a check still
+# executed by some other test and asserted by none is what neither
+# question sees, and one constant is now the whole of what would have
+# asked about it.
+#
 # That is a statement about the shapes, not a promise about the rate:
 # what the next scheduled run reports is what it reports, and it is not
 # restated here, a number in a comment being a line every change to the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1703,6 +1703,24 @@ release-notes length in the first place, and are still in
   `macos.yml`'s cells is `windows.yml`'s too, and `test.yml`'s aggregate
   job explains its name by there being a second workflow over this matrix
   shape, which is now a third
+- **cosmic-ray's own defence against a stale `.pyc` is named rather than
+  asserted** (#229). CLAUDE.md said "cosmic-ray does not have the
+  problem" after two paragraphs on how a hand-applied mutation loses
+  one, and did not say why — which is what let the sentence be read as
+  an assumption. It is `cosmic_ray/testing.py`, which sets
+  `PYTHONDONTWRITEBYTECODE` in the environment of the test command under
+  a comment giving this very reason, so no `.pyc` is written during a
+  session; measured by looking at `btclib_secp256k1/__pycache__` before
+  and after a real `cosmic-ray exec` over `hashes.py`, one file before
+  and the same one after, and the baseline step writes none either. Two
+  things go in the same clause because each is what the next reader
+  would otherwise re-derive: grepping the package for
+  `dont_write_bytecode` finds nothing and proves nothing, the string
+  being the environment variable's own spelling; and no flag stops a
+  *read*, so a `.pyc` from an earlier `pytest` is still used where its
+  size and truncated mtime match — the same window as the hand case,
+  which is why that recipe wants the control run and not only the
+  variable.
 
 ### CI
 
@@ -1929,6 +1947,23 @@ release-notes length in the first place, and are still in
   `secp256k1_extrakeys.h` and both now exercised over either answer, plus
   `sigout` of `secp256k1_ecdsa_signature_normalize`, for which `dsa`
   passes `ffi.NULL` already
+- **`bindings.toml` records the resolution a consolidated width costs**
+  (#231). Its preamble already carries the two shapes that were answered
+  in the code and the one equivalence that cannot be, each written down
+  so a session does not spend itself rediscovering them; a width
+  collapsed to one statement belongs in that list for the opposite
+  reason. Two modules had several mutable copies of one number —
+  `ellswift` three, at `_decode_` and at both of `xdh`'s checks, and
+  `ssa` two, at `_verify_` and at `verify` — and `core/NumberReplacer`
+  carries `OFFSETS = [+1, -1]`, a position per offset, so `ellswift`'s
+  six mutants are two now and either of them changes every site at once,
+  dying on whichever test reaches it first. A check that later loses the
+  test covering it can therefore no longer surface as a survivor.
+  Nothing is hidden today, the seven constants #221 introduced having
+  each been mutated both ways before it landed and all fourteen dying,
+  and the paragraph says what the ratchet does and does not catch
+  instead: a line that stops running is caught, a line still run by
+  another test and asserted by none is what neither question sees.
 
 ## v0.8.0.2
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -310,7 +310,18 @@ findings.
   python reuses the mutated bytecode — silently, in the next unrelated run.
   `PYTHONDONTWRITEBYTECODE=1`, with a passing baseline run before and a
   passing control run after, is what makes a hand verification mean
-  anything. cosmic-ray does not have the problem
+  anything. **cosmic-ray does not have the problem, and the mechanism is
+  worth naming rather than trusting** (#229): `cosmic_ray/testing.py`
+  sets that very variable in the environment of the test command, under a
+  comment giving this reason, so a session writes no `.pyc` at all — the
+  baseline included, both measured by looking at `__pycache__` before and
+  after a real `cosmic-ray exec`. Grepping the package for
+  `dont_write_bytecode` finds nothing and means nothing: the string it
+  sets is the environment variable's own spelling. What no flag stops is
+  a *read*, so a `.pyc` already on disk from an earlier `pytest` is still
+  used where its size and truncated mtime match — which is the same
+  window as above, and the reason the hand recipe wants the control run
+  and not only the variable
 - **`detect-secrets` runs twice, over two baselines**: the tree with the
   entropy detectors on, and `tests/*.csv`/`tests/*.json` with them off,
   those files being hex and nothing else. Adding to either side fails its


.github/mutation/bindings.toml
```diff
@@ -57,6 +57,26 @@
 # after, the same way the filter below skips the annotations by operator
 # rather than asking every mutant of a `|` to be read again.
 #
+# One thing to know that is not a survivor at all, and is the reason
+# "expecting nothing" is worth reading as exactly that and no more (#231).
+# #221 collapsed every buffer width in the package to one statement, and in
+# three places that turned several independently mutable copies of a number
```
fametrano: Two counts here do not survive a recount:

- **"three places"** is two. Against `71fbd0ac^` the mutable copies were `ellswift` 3 and `ssa` 2; `dsa` and `recovery` had one each and `hashes` none (its 32 was inside `"char[32]"`). #221's own message: "`ellswift` had three independently mutable copies of 64 and `ssa` two".
- **"three literals gave three mutants"** is six, and "one constant gives one" is two: `core/NumberReplacer` has `OFFSETS = [+1, -1]` and yields a position per offset. #221 uses the factor correctly -- "seven literals ... fourteen mutants between them".

fametrano (commented):
**Changes requested** on `77c38eb`. The `#229` half is right and I could
not make it wrong; the `#231` paragraph has two counts that do not survive
a recount, and understates the silence it exists to describe.

## Verified: the cosmic-ray half

Better than the body claims, in one respect. The measurement is one
session's behaviour; the mechanism makes it a property:

- `cosmic_ray/testing.py:56` sets `env["PYTHONDONTWRITEBYTECODE"] = "1"`
  under the comment at 52-54 giving exactly this reason.
- `grep -rin dont_write_bytecode` over the installed package is empty, so
  the issue's grep could not have hit and the clause saying so is worth its
  line.
- **The baseline goes through the same call**, which the body asserts and
  measures but does not source: `cli.py`'s `baseline` builds a
  `WorkItem(mutations=[], job_id="baseline")` and hands it to
  `cosmic_ray.commands.execute`, which reaches `mutating.py:75`'s
  `run_tests` — the one function that builds that environment. So "the
  baseline included" holds by construction and not only in the run that was
  watched. Worth the half-clause, since it is what makes the sentence
  general.

## 1. "In three places" is two

`bindings.toml`, new paragraph: "in three places that turned several
independently mutable copies of a number into one: `ellswift` had 64 at
`_decode_` and at both of `xdh`'s length checks, `ssa` had it at
`_verify_` and at `verify`."

Two are named because two is the count. Recounted against `71fbd0ac^`, the
mutable copies — literals a `NumberReplacer` could reach, so not the ones
inside `"char[64]"` — were:

| module | copies before | after |
| --- | --- | --- |
| `ellswift` | 3 (`octets` at `_decode_`, and both of `xdh`'s) | 1 |
| `ssa` | 2 (`octets` at `_verify_` and at `verify`) | 1 |
| `dsa` | 1 | 1 |
| `recovery` | 1 | 1 |
| `hashes` | 0 (the 32 was inside `"char[32]"`) | 1 |

`#221`'s own commit message says it in those words: "`ellswift` had three
independently mutable copies of 64 and `ssa` two". No third place.

## 2. "Three mutants" is six, and "one" is two

Same sentence: "Where three literals gave three mutants, one constant gives
one."

`core/NumberReplacer` has `OFFSETS = [+1, -1]` and yields one position per
offset, so it emits **two** mutants per numeric literal. Three literals
gave six; one constant gives two. The ratio the sentence is about is
unaffected, which is why it reads plausibly — but this file is where a
session reads counts, `#221` states the factor correctly in the same
breath ("seven literals that already sat at the validation sites, fourteen
mutants between them"), and the preamble two paragraphs up already trades
in exact mutant counts ("365 - 13 = 352").

## 3. The silence is larger than the paragraph, and predates the change

This is the one worth a revision rather than a numeral. The paragraph makes
the silence a consequence of `#221`, and its own examples are the two
smallest instances. Counting every width constant standing behind more than
one *check* today:

| constant | check sites |
| --- | --- |
| `xonly._XONLY_SIZE` | 5 — `xonly.py:194, 439, 551, 585, 665` |
| `ellswift._ENCODING_SIZE` | 3 — `139, 226, 227` |
| `ssa._SIGNATURE_SIZE` | 2 — `529, 574` |
| `silentpayments.LABEL_SIZE` | 2 — `697, 791` |

Twelve sites behind four constants, where the paragraph names five behind
two. And the largest of them is not `#221`'s doing: `xonly` and
`silentpayments` "already had" this spelling, in `#221`'s own words, so one
mutant has stood for `xonly`'s five length checks since before that change
— and `#221`'s verification, "all seven of the consolidated literals were
mutated one at a time", covers the literals it consolidated and says
nothing about those five.

Which inverts what a session takes away. As written it reads "this change
cost some resolution, and here is the little of it"; what is true is "one
mutant per width is the shape of this package, `xonly` most of all, and
`#221` brought two more modules into it". The paragraph's subject should be
the shape, since that is what the next survivor list will and will not be
able to say.

Everything else here is right, and the two-shapes-plus-one-silence framing
is the right place for it.

fametrano:
Both counts recounted and both wrong as I had them. Fixed on `72be84f`, in
the paragraph, the changelog bullet and the commit message.

**"three places" is two.** An AST count of the int literals that are a
buffer width, at `71fbd0ac^` and at `71fbd0ac`:

| module | before | after |
| --- | --- | --- |
| `ellswift` | 3 × 64 | 1 |
| `ssa` | 2 × 64 | 1 |
| `dsa` | 1 × 64 | 1 |
| `recovery` | 1 × 64 | 1 |
| `hashes` | 0 | 1 |

So two modules had *several* copies collapsed, which is what the sentence
was about, and the list under it already named exactly those two — the
numeral contradicted its own example.

**And the factor is two.** `cosmic_ray/operators/number_replacer.py`:
`OFFSETS = [+1, -1]`, and `mutation_positions` yields a position per
offset. So `ellswift`'s three literals gave **six** mutants and its one
constant gives **two**. The paragraph now says that, with the operator's
own line as the reason rather than a bare number, and #221's "seven
literals … fourteen mutants" is the arithmetic it should have matched from
the start.

While correcting it, one more claim of mine was loose and is now precise:
"all seven of those literals were mutated one at a time" conflated the
seven *constants* #221 introduced with the literals it consolidated. It is
the seven constants, each mutated both ways, all fourteen dying — which is
#221's table exactly.

Rebased on `752437d`, still one commit.

fametrano (commented):
**Correcting my own finding 3 before you act on it.** I called `xonly`'s
five `_XONLY_SIZE` sites "check sites". Four of them are not checks. Read
rather than grepped:

| site | what it is |
| --- | --- |
| `xonly.py:439` `tweak_add_check` | `octets(…, _XONLY_SIZE)` — a width refusal |
| `xonly.py:194` `from_pubkey` | `if len(…) == _XONLY_SIZE:` — **dispatch**: 32 is the x, anything else goes to `keys.parse` |
| `xonly.py:551` `parse` | `if len(…) != _XONLY_SIZE:` — **dispatch** to `keys._parsed`, so a 33- or 65-octet sec key is accepted, not refused |
| `xonly.py:585` `_parsed` | the same dispatch, answering None |
| `xonly.py:665` `to_pubkey` | **dispatch**: 32 gets `0x02 ‖ x`, otherwise the point is already lifted |

That changes the size of what I claimed twice over. A dispatch is
exercised from both sides by every ordinary test — 32-octet x-only keys and
33- or 65-octet sec keys go through all four constantly — so a mutated
`_XONLY_SIZE` sends each of them down the wrong branch and dies loudly.
"Twelve sites behind four constants" counted those four as if they were
refusals, and they are not comparable to `ellswift`'s three `octets(…)`.

The accurate count of **hard width refusals** standing behind one constant:

| constant | refusals |
| --- | --- |
| `ellswift._ENCODING_SIZE` | 3 — `139, 226, 227` |
| `ssa._SIGNATURE_SIZE` | 2 — `529, 574` |
| `silentpayments.LABEL_SIZE` | 2 — `697, 791` |
| `xonly._XONLY_SIZE` | 1 — `439` |

So your paragraph's two examples are the two largest, which I said they
were not. What is left of the finding is one omission and one distinction,
both smaller than what I posted:

- **`silentpayments.LABEL_SIZE` is a third instance**, two refusals behind
  one constant — `parse_label` and `_fill_label_cache` — and it predates
  `#221`, that module having had the spelling already. So "in three places"
  would even be *right* about the shape if it counted this one, which is a
  more useful repair than deleting the numeral.
- **`xonly`'s four dispatch branches are a different shape** and worth at
  most a clause: one constant decides which arm four entry points take, and
  a mutant of it cannot isolate one of them either — but nothing there is a
  test nobody wrote, because both arms are load-bearing in the ordinary
  tests.

Findings **1** and **2** stand as posted: "three places" against two, and
three-literals-give-six-mutants, both against `#221`'s own message and
`NumberReplacer`'s `OFFSETS = [+1, -1]`.

fametrano (commented):
**ACK `72be84f2`.** Both counts corrected, and the third correction is one
I had not asked for and is the right one.

Re-derived on this head rather than read:

- **"two modules"** — against `71fbd0ac^`, the mutable copies were
  `ellswift` 3, `ssa` 2, `dsa` 1, `recovery` 1, `hashes` 0. Two modules had
  *several*, and the list under the numeral now agrees with it.
- **The factor** — `OFFSETS = [+1, -1]` with a position yielded per offset,
  so `ellswift`'s three literals gave six and its one constant gives two.
  Citing the operator's own line rather than the bare number is better than
  the number: it is the thing that would go stale if cosmic-ray changed it.
- **"the seven constants #221 introduced … all fourteen die"** — this is the
  correction you made unprompted and it checks out exactly. At `71fbd0ac`
  the int-literal width constants are seven: `dsa._COMPACT_SIZE`,
  `ellswift._ENCODING_SIZE`, `hashes._HASH_SIZE`,
  `keys._COMPRESSED_SIZE`, `keys._UNCOMPRESSED_SIZE`,
  `recovery._COMPACT_SIZE`, `ssa._SIGNATURE_SIZE`. `dsa._DER_SIZE` is
  outside the count and has to be, being `ffi.sizeof` of a cdecl — which is
  `#221`'s own documented exception, so the seven land on the right side of
  it. Seven constants, two mutants each, fourteen: `#221`'s arithmetic
  exactly, which is what the old sentence was reaching for.
- The commit message carries both corrections, so `git log` and the file
  will not disagree the way `752437d`'s did.

The `#229` half I could not fault on the first read and it is unchanged: the
mechanism is `testing.py:56` under its own comment, the issue's grep could
not have hit, and the baseline reaches the same `run_tests` by construction.

On my finding 3, which you did not take and should not have: the sentence is
scoped to what `#221` collapsed, and that is two modules exactly, so it is
true as it stands. My suggestion — that `silentpayments.LABEL_SIZE` is a
third instance of the shape and predates the change — was about widening the
subject from the change to the shape, which is a different paragraph and not
a defect in this one. If it is ever worth writing, `xonly`'s four dispatch
branches belong in it too, and they are not checks; I said they were, and
that was the part of my review that was wrong rather than optional.

CI on this head is 25 green with nothing pending.

Issue #230, opened 2026-08-17
The secrets baseline is rewritten by a line that moved, and --slim is what stops it


fametrano:
`.secrets.baseline` is rewritten whenever a line that carries a flagged
pseudo-secret moves, and the rewrite carries a fresh `generated_at` with it.
Over [#221](https://github.com/btclib-org/btclib-secp256k1/pull/221) that
happened on all four pushes, and the only thing it recorded each time was that
one entry's `line_number` had gone 40, 48, 49 as a comment block above it grew:

```diff
         "filename": "btclib_secp256k1/hashes.py",
         "hashed_secret": "a608d88cb7e04b46b5e44fa00bb7c01b7748e6f0",
         "is_verified": false,
-        "line_number": 48
+        "line_number": 49
       }
-  "generated_at": "2026-08-17T09:42:14Z"
+  "generated_at": "2026-08-17T10:49:53Z"
```

Nothing about the secrets changed in any of the four. The hook was not
complaining either — a moved line does not fail `detect-secrets-hook`, it just
means the committed baseline no longer matches what a regeneration would
produce, so a regeneration lands in the diff.

## What stops it

`detect-secrets scan --slim`, which is in v1.5.0 and is exactly for a baseline
that gets committed: it omits both fields. A results entry carries

    ['filename', 'hashed_secret', 'is_verified', 'line_number', 'type']

without the flag and

    ['filename', 'hashed_secret', 'is_verified', 'type']

with it, and `generated_at` is absent from a slim baseline altogether. The hook
takes one, including after the flagged line has moved:

```shell
$ detect-secrets scan --slim f.py > .secrets.baseline
$ detect-secrets-hook --baseline .secrets.baseline f.py ; echo $?
0
$ # prepend a line, so the secret is now on line 2, unchanged
$ detect-secrets-hook --baseline .secrets.baseline f.py ; echo $?
0
```

The change would be the two commands in
[CONTRIBUTING.md:423](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/CONTRIBUTING.md#L423) and
[:429](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/CONTRIBUTING.md#L429), and one regeneration of each baseline.

## The trade, which is the reason this is a question and not a defect

[...]

PR #237, opened 2026-08-17
Rewrap what an insertion left short, and say what the sentences omitted

### LINKED ISSUE #232: silentpayments names its two widths publicly where the other six modules name them privately
Six modules name a buffer width privately —
[`xonly._XONLY_SIZE`](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/xonly.py#L42), and the five
[#221](https://github.com/btclib-org/btclib-secp256k1/pull/221) adds in `dsa`,
`recovery`, `hashes`, `ssa` and `ellswift`, plus `keys`'s two. The seventh does
not:

```python
SUMMARY_SIZE = ffi.sizeof("secp256k1_silentpayments_prevouts_summary")
LABEL_SIZE = 33
```

— [silentpayments.py:43-44](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/silentpayments.py#L43-L44),
both public, and `_LABEL_BUFFER_TYPE` beside them private like every other
buffer type in the package.

## Why this is a question

The split looks intentional and the code half-says so. `SUMMARY_SIZE` is a
length a *caller* cannot work out — it is `ffi.sizeof` of a struct, not a
number any BIP writes down — and a caller holding a serialized summary needs
it; a doctest shows it being used that way, at
[silentpayments.py:473](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/silentpayments.py#L473):

```
>>> len(summary) == silentpayments.SUMMARY_SIZE
```

`LABEL_SIZE` is the 33 bytes of a compressed point, which a caller could
equally want, and `tests/test_modules.py` reaches for both through the module
name. So the rule that would explain the whole package is something like *a
width a caller needs is public; a width only a buffer needs is not* — which is
a different rule from the one
[xonly.py](https://github.com/btclib-org/btclib-secp256k1/blob/68657e14c47c9d7a2ff29a39181c38c8b310ac7c/btclib_secp256k1/xonly.py#L44-L57) states, that being about
whether anything unpacks the buffer.

The module's own comment explains why the two widths *exist* and where each
comes from, but not why these two are the package's only public ones. One
clause would settle it, and after #221 — which makes "the six modules read
alike" a claim in `CHANGELOG.md` — the seventh is the one a reader will ask
about.

If the answer is "public on purpose, for the reason the doctest shows", then
saying that where the constants are declared is the whole of the change, and
this closes.

### LINKED ISSUE #233: Thirteen more lines stopped short where a qualifier went in
[#225](https://github.com/btclib-org/btclib-secp256k1/issues/225) closed with
one line reflowed, `CHANGELOG.md:75`, and named the class in passing: "no hook
reflows prose, which is why both passed the gate". This is that class, counted.

A widow is what a mid-block insertion leaves behind: the qualifier goes in, the
line it lands on stops short, and nothing rewraps the paragraph. It is
invisible in a diff of the change that made it, because the change is a line
that was already going to be rewritten.

## What is there

At [`71fbd0a`](https://github.com/btclib-org/btclib-secp256k1/tree/71fbd0ac),
fourteen lines are shorter than 46 columns, follow a line of 65 or more, and
end **mid-sentence** with a next word that would have fitted under 72. That
last condition is what separates an artifact from a choice: a line ending in
`.`, `:` or `—` is somebody starting a clause on its own line, and those are
not counted here.

| file | line | the line | the word that fitted |
| --- | --- | --- | --- |
| `CHANGELOG.md` | 75 | `stronger argument: the recovered` | `key` |
| `CHANGELOG.md` | 480 | `rounds — 500 000 calls for the` | `primitive,` |
| `CHANGELOG.md` | 1799 | `base64 detector reads. A` | `` `pragma:`` |
| `CHANGELOG.md` | 2324 | `name is unchanged,` | `which` |
| `CHANGELOG.md` | 2551 | `` `.venv-3.10`, which is what`` | `` `UV_PROJECT_ENVIRONMENT=…`` |
| `README.md` | 31 | `As used by the` | `[btclib](…)` |
| `README.md` | 459 | `MuSig2 is not` | `wrapped,` |
| `README.md` | 1061 | `is for, and` | `[the` |
| `CONTRIBUTING.md` | 615 | `still` | `` `gh`` |
| `RELEASING.md` | 5 | `the latter produced. Where` | `it` |
| `RELEASING.md` | 107 | `section of both against `git log`, drop the` | `` `(work`` |
| `RELEASING.md` | 377 | `same time, headed` | `` `##`` |
| `HISTORY.md` | 280 | ``arithmetic. Like `xonly.from_prvkey` it`` | `is` |
| `SECURITY.md` | 15 | `ours to fix, and it has its own` | `[security` |

`CHANGELOG.md:75` is
[#228](https://github.com/btclib-org/btclib-secp256k1/pull/228)'s, so thirteen
remain. `CONTRIBUTING.md:615` is the extreme: five columns.

## How the count was made

```python
import re, subprocess

W = 72
for path in ("CHANGELOG.md", "README.md", "CONTRIBUTING.md", "CLAUDE.md",
             "REPOSITORY.md", "RELEASING.md", "HISTORY.md", "SECURITY.md",
             "scripts/README.md"):
    text = subprocess.run(["git", "show", f"origin/main:{path}"],
                          capture_output=True, text=True).stdout
    lines, infence = text.split("\n"), False
    for i, ln in enumerate(lines[:-1]):
        if ln.lstrip().startswith("```"):
            infence = not infence
            continue
        if infence:
            continue
        prev, nxt = (lines[i - 1] if i else ""), lines[i + 1]
        s, n, p = ln.strip(), nxt.strip(), prev.strip()
        if not s or not n or not p:
            continue
        if re.match(r"^\s*([-*]|\d+\.|#|\||>|<!--)", nxt):
            continue
        if s.startswith(("#", "|", ">", "<!--", "//")):
            continue
        if len(ln) > 46 or len(prev) < 65:
            continue
        if len(ln) + 1 + len(n.split()[0]) > W:
            continue
        if re.search(r"[.:;!?—]$", s):     # a clause ended: a choice, not an artifact
            continue
        print(f"{path}:{i + 1}  ({len(ln)} cols)  {s!r} + {n.split()[0]!r}")
```

Loosening any of the three thresholds finds more and finds noise with it: at 55
columns instead of 46 the same scan reports 46 lines, most of them a sentence
deliberately begun on a line of its own. 46 and "mid-sentence" is where every
hit above reads as a defect on inspection, and that is the whole of the
justification for those numbers — they are a filter, not a rule.

## What could be done

1. **Reflow the thirteen**, and nothing else. Half an hour, no tooling, and the
   class comes back the next time a qualifier goes into a paragraph.
2. **A hook.** `prettier` is already here and already rewrites in place — but
   only over `yaml and jsonc`, as the hook's own name says. Giving it markdown
   with `--prose-wrap always` would reflow every prose file in the repository
   in one commit, which is a diff nobody can review against a claim that only
   whitespace moved, and it would also rewrap the paragraphs that are wrapped
   the way they are on purpose. Not obviously worth it, and not reversible
   cheaply.
3. **The scan above as a `local` hook**, failing on a widow rather than fixing
   it. It is a check somebody can act on — the fix is one line — and it does
   not touch what it does not report. What it costs is the thresholds: they are
   a heuristic, and a false positive is a `# noqa`-shaped exception in prose,
   which there is no spelling for.

(1) is worth doing whichever of the others follows it. (3) is the one that
would have caught #225's line at the commit that wrote it; whether a heuristic
belongs in a gate that is otherwise exact is the question this issue is really
asking.

### LINKED ISSUE #234: The worktree recipe costs a submodule clone the sentence under it does not count
Raised in the [review of #227](https://github.com/btclib-org/btclib-secp256k1/pull/227#pullrequestreview-4954072831)
as non-blocking, and filed rather than folded in because that branch landed.

The paragraph under the worktree recipe in `CLAUDE.md` says:

> The venv and the C build are the whole of the cost, and they buy the thing
> that matters […]

After #223 they are not the whole of it. The recipe's new line clones
bitcoin-core/secp256k1 again: a linked worktree gets a submodule module of its
own rather than sharing the primary checkout's — `secp256k1/.git` in the
worktree reads `gitdir: ../../bs/.git/worktrees/<wt>/modules/secp256k1` — which
was measured at 14 MB under `.git/worktrees/<wt>/modules` and 7 MB of tree
(git 2.50.1).

That is the same "minutes, not seconds" argument the venv and the build get,
it belongs in the same sentence, and stating it answers the question a session
asks next: whether `--reference` against the primary's object store would be
cheaper, and whether sharing that store is safe across worktrees.

Two facts from the same review, worth keeping with it because they are what
makes the recipe safe and neither is obvious:

- **the primary checkout is not disturbed.** `core.worktree` in
  `.git/modules/<name>/config` is a single value and a second checkout of one
  submodule can rewrite it under the first — but git gives the linked worktree
  its own module, the primary's `core.worktree` still reads
  `../../../secp256k1`, and `git -C secp256k1 status` there stays clean through
  the whole sequence;
- **`git worktree remove --force` still finishes** with an initialized
  submodule inside: exit 0, tree gone, `.git/worktrees` gone with it, nothing
  left for `git worktree prune`. The recipe's last line needs no companion.

One clause, and whichever of those two is worth a sentence beside it.

fametrano:
Prose only. Nothing changes about what any call answers.

## #233 — ten lines, twelve blocks, and three that are not defects

Each widow is what a qualifier inserted into a paragraph leaves behind
when nothing rewraps the block. **The unit of repair is the block**, which
#228 established the hard way: fixing the one line #225 named left two
more of exactly that shape.

| file | blocks rewrapped |
| --- | --- |
| `CHANGELOG.md` | 4 |
| `README.md` | 3 |
| `RELEASING.md` | 3 |
| `CONTRIBUTING.md` | 1 |
| `HISTORY.md` | 1 |

Word for word identical afterwards, and **asserted rather than
eyeballed**: the rewrap compares the joined words of each block before and
after and refuses to write where they differ. A markdown link is kept
whole for it — its inner spaces hidden from `textwrap` and restored —
because a link is one token to a reader and several to a wrapper.

**Three of the thirteen are not defects.** A line before a link of 63, 84
and 125 columns is short because nothing can follow it:

```
README.md:1059   'for, and' + [the libsecp256k1 wrappers table](…)  -> joining needs 134
RELEASING.md:7   '… Both go through' + [Trusted Publishing](…)      -> needs 102
SECURITY.md:15   '… it has its own' + [security policy](…)           -> needs 116
```

The filter measured `[security` where the token is the whole
`[security policy](url)`. With links kept whole those three blocks rewrap
to exactly what they already say, so ten is the count of real ones and
those three are the shape the heuristic cannot see.

**No hook**, which is the question the issue says it is really asking.
The thresholds are a heuristic and a false positive has no `# noqa` to be
marked with in prose; a gate exact everywhere else is the wrong place for
a check that is nearly right, and `.yamllint.yaml` already records what
that trade costs — a rule disabled at 46 findings that were all pins.

## #234 — the recipe costs a clone the sentence did not count

"The venv and the C build are the whole of the cost" stopped being true
when the submodule line went in: a linked worktree gets a module of its
own rather than sharing the primary checkout's, 14 MB under
`.git/worktrees/<wt>/modules` and 7 MB of tree. Counted now, which also
answers what a session asks next — `--reference` is cheaper and buys a
shared object store git deliberately does not give a linked worktree.

The two facts from that review go with it: `core.worktree` is a single
value a second checkout of one submodule can rewrite under the first, and
does not here; and `git worktree remove --force` finishes with an
initialized submodule inside, so the recipe's last line needs no
companion.

## #232 — why two widths are public

Not `xonly.py`'s rule, which is about a buffer the module fills and
unpacks. These are lengths the module answers a caller with.
`SUMMARY_SIZE` has to be public: `ffi.sizeof` of a struct, written down by
no BIP, moving when libsecp256k1 moves, unknowable to a caller holding
what `prevouts_summary` returned. `LABEL_SIZE` is 33 and could be worked
out — what keeps it public is that it has been since 0.8.0, and taking a
released name private to tidy an asymmetry is a break charged to a reader
who never asked. `keys._COMPRESSED_SIZE` is the same 33 and stays private:
nothing there checks a caller's argument against it, where this checks
both the label handed in and the one written back.

## Verified

Branches from `main` at `752437d`; it shares no line with #236, which
appends to the same two changelog sections.

Closes #232
Closes #233
Closes #234

## Summary by Sourcery

Clarify documentation prose around worktree costs and public buffer widths while reflowing markdown paragraphs to remove short, mid-sentence lines.

Documentation:
- Rewrap multiple markdown paragraphs across changelog, README, CONTRIBUTING, RELEASING and HISTORY to eliminate mid-sentence short lines without changing their wording.
- Document the additional cost of a submodule clone and related git worktree behaviour in the contributor workflow guide.
- Explain in-code and in the changelog why `silentpayments` exposes two public buffer widths while other modules keep theirs private.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -104,9 +104,31 @@ git push origin HEAD:refs/heads/<branch>
 git worktree remove --force "$WT" # removing it is part of finishing
 ```
 
-The venv and the C build are the whole of the cost, and they buy the thing
-that matters: a commit cannot contain work that was never in it, and the
-maintainer's branch does not move under them.
+The venv, the C build and a second clone of the submodule are the whole
+of the cost, and they buy the thing that matters: a commit cannot
+contain work that was never in it, and the maintainer's branch does not
+move under them. The clone is the third of those because a linked
+worktree gets a submodule module of its own rather than sharing the
+primary checkout's — `secp256k1/.git` there reads `gitdir:
+…/.git/worktrees/<wt>/modules/secp256k1` — which was measured at 14 MB
+under `.git/worktrees/<wt>/modules` and 7 MB of tree. `--reference`
+against the primary's module is what a session asks about next, and it
+works: `git submodule update --init --reference
+<primary>/.git/modules/secp256k1 secp256k1` leaves the module directory
+exactly where git puts it, writes one `objects/info/alternates` pointing
+at the primary's `objects`, and measures **128 KB** against those 14 MB,
+with the primary's own `core.worktree` untouched and its submodule
+clean. What it costs is the pointer: the worktree's submodule then has
+no copy of its own objects, so a `git gc` or a repack in the primary —
+or moving or deleting it — can leave this one unable to find them. What
+git is keeping apart by giving each linked worktree a module of its own
+is the submodule's *state*, so that two worktrees can have it checked
+out at two commits; the object store sitting inside that module is a
+consequence of where the state lives rather than a refusal to share
+objects, which is why `--reference` is allowed to share them and why it
+changes nothing about `core.worktree`. For a recipe whose last line
+removes the worktree anyway that is a trade worth declining, and
+declining knowingly is the point of the paragraph.
 
 **The submodule line is what makes the rest of the recipe run**, and
 leaving it out costs a session rather than a build: `git submodule status`
@@ -117,6 +139,17 @@ the two things that are not wrong. CI never meets it, every checkout there
 passing `submodules: true`. It is the same sentence as the one below about
 `refs/stash`: a worktree isolates files, and neither a submodule checkout
 nor a ref is one.
+
+**Two things the recipe leans on, both measured rather than assumed**, and
+worth knowing because a submodule inside a linked worktree is a known sharp
+edge. `core.worktree` in `.git/modules/<name>/config` is a single value, so
+a second checkout of one submodule can rewrite it under the first — and
+does not here, git giving the linked worktree its own module: the primary's
+`core.worktree` still reads `../../../secp256k1` and `git -C secp256k1
+status` there stays clean through the whole sequence. And `git worktree
+remove --force` still finishes with an initialized submodule inside: exit
+0, tree gone, `.git/worktrees` gone with it, nothing left for
+`git worktree prune`. So the recipe's last line needs no companion.
 
 **Never `git stash`, in the primary checkout or in a worktree:
 `refs/stash` is shared.** A worktree isolates files, not refs, so
--- a/btclib_secp256k1/silentpayments.py
+++ b/btclib_secp256k1/silentpayments.py
@@ -39,7 +39,26 @@ from .context import ctx
 # summary's is asked of the struct rather than written down, so that a
 # libsecp256k1 that changes it changes this too; the label's is the 33
 # bytes of a compressed point, which is its serialization and not the 68
-# of the object holding it
+# of the object holding it.
+#
+# Public, where the widths of the other six modules are private, and the
+# question those answer is a different one: `xonly.py` states its own for
+# the buffer it unpacks, and nothing outside the module needs it. This
+# module answers a caller two lengths (#232). `SUMMARY_SIZE` is the one
+# that has to be public: it is `ffi.sizeof` of a struct, no BIP writes it
+# down, it moves when libsecp256k1 moves, and a caller holding what
+# `prevouts_summary` returned can learn how many octets that is nowhere
+# else -- its docstring shows the check. `LABEL_SIZE` is 33 because a
+# label is a compressed point, so a caller could work it out where it
+# could not work the other out; what keeps it public is that it has been
+# since 0.8.0, and taking a released name private to tidy an asymmetry is
+# a break charged to a reader who never asked. `keys._COMPRESSED_SIZE` is
+# the same 33 and stays private: nothing there checks a caller's argument
+# against it, where two callers' arguments are checked against this one --
+# `parse_label`'s octets and the labels `_fill_label_cache` takes for
+# `_scan_outputs_`. What `serialize_label` does with it is not that: it
+# unpacks the buffer this module declared, which is exactly what `keys`
+# does with its own
 SUMMARY_SIZE = ffi.sizeof("secp256k1_silentpayments_prevouts_summary")
 LABEL_SIZE = 33
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -555,13 +555,13 @@ release-notes length in the first place, and are still in
   int and build the type from it. The int is the spelling that says what
   the number is, and it is the one the whole package uses now.
 - **What the eight cost, and what they cost now.** `main`'s spelling of
-  each call written out and alternated against the shipped one in a single
-  process over one build on the tree this branch lands as, minimum of 15
-  rounds — 500 000 calls for the
-  primitive, 300 000 for the serializations, 20 000 for the four hosts —
-  with every pair asserted to answer alike before it is timed, and a noise
-  row for each site rather than for one of them. An Apple M5, macOS 26.6,
-  arm64, CPython 3.14.6, microseconds per call:
+  each call written out and alternated against the shipped one in a
+  single process over one build on the tree this branch lands as,
+  minimum of 15 rounds — 500 000 calls for the primitive, 300 000 for
+  the serializations, 20 000 for the four hosts — with every pair
+  asserted to answer alike before it is timed, and a noise row for each
+  site rather than for one of them. An Apple M5, macOS 26.6, arm64,
+  CPython 3.14.6, microseconds per call:
 
   | | main | now | noise |
   | --- | --- | --- | --- |
@@ -1749,6 +1749,70 @@ release-notes length in the first place, and are still in
   unset, so the loss would show as a diff of nothing. Nothing is lost
   today, all 53 and all 466 findings being unverified and none carrying
   `is_secret`.
+- **Ten lines that stopped short mid-sentence are the paragraphs they
+  sit in, rewrapped** (#233). Each is the artifact an insertion leaves
+  in a block nothing reflows: the qualifier goes in, the line it lands
+  on stops early, and the diff of the change that made it shows a line
+  that was going to be rewritten anyway. The unit of repair is the block
+  and not the line, which #228 established the hard way — fixing the one
+  line #225 named left two more of the same shape. Twelve blocks moved
+  across `CHANGELOG.md`, `README.md`, `CONTRIBUTING.md`, `RELEASING.md`
+  and `HISTORY.md`, every one of them word-for-word identical to what it
+  said before, which is asserted rather than eyeballed: the rewrap
+  compares the joined words of the block before and after and refuses to
+  write where they differ. Three of the thirteen the issue counted are
+  not defects at all and stay: a line before a markdown link of 63, 84
+  and 125 columns is short because nothing can follow it, and the filter
+  measured the link's first word rather than the link. No hook does
+  this, which #233 asked about and answers no: the thresholds are a
+  heuristic, a false positive has no `# noqa` to be marked with in
+  prose, and a gate that is exact everywhere else is the wrong place for
+  one that is nearly right.
+- **The worktree recipe costs a submodule clone, and the sentence under
+  it counts it now** (#234). It said the venv and the C build "are the
+  whole of the cost" and after the submodule line they are not: a linked
+  worktree gets a module of its own rather than sharing the primary
+  checkout's, measured at 14 MB under `.git/worktrees/<wt>/modules` and
+  7 MB of tree, which is the same "minutes, not seconds" argument the
+  other two get. `--reference` is what a session asks about next and is
+  measured rather than reasoned about: it works, leaves the module
+  directory where git puts it, writes one `objects/info/alternates` at
+  the primary's `objects`, and comes to 128 KB against those 14 MB with
+  the primary's `core.worktree` untouched. What it costs is that pointer
+  — no copy of its own objects, so a `git gc` or a repack in the
+  primary, or moving it, can leave the worktree's submodule unable to
+  find them. What git keeps apart by giving each linked worktree a
+  module of its own is the submodule's *state*, so two worktrees can
+  have it at two commits; the object store inside that module follows
+  from where the state lives rather than from a refusal to share
+  objects, which is why `--reference` may share them and leaves
+  `core.worktree` alone. For a recipe that removes the worktree anyway
+  that is a trade worth declining knowingly. Two facts measured while
+  checking that go with it, both about a sharp edge: `core.worktree` is
+  a single value that a second checkout of one submodule can rewrite
+  under the first, and does not here; and `git worktree remove --force`
+  finishes with an initialized submodule inside, exit 0 and nothing left
+  for `git worktree prune`, so the recipe's last line needs no
+  companion.
+- **`silentpayments` says why its two widths are the package's only
+  public ones** (#232). Six modules name a buffer width privately and
+  this one does not, which after #221 made "the six modules read alike"
+  a claim a reader would test against the seventh. The rule is not the
+  one `xonly.py` states about its own: those are the size of a buffer
+  the module fills and unpacks, where these are lengths this module
+  answers a caller with. `SUMMARY_SIZE` has to be public — `ffi.sizeof`
+  of a struct, written down by no BIP, moving when libsecp256k1 moves,
+  and unknowable to a caller holding what `prevouts_summary` returned.
+  `LABEL_SIZE` is 33 and could be worked out, so what keeps it public is
+  that it has been since 0.8.0 and taking a released name private to
+  tidy an asymmetry is a break charged to a reader who never asked.
+  `keys._COMPRESSED_SIZE` is the same 33 and stays private: two callers'
+  arguments are checked against this one, `parse_label`'s octets and the
+  labels `_fill_label_cache` takes for `_scan_outputs_`, where nothing
+  in `keys` checks an argument against a width at all. What
+  `serialize_label` does with it is the other kind of use and is named
+  as such — unpacking the buffer this module declared, which is exactly
+  what `keys` does with its own.
 
 ### CI
 
@@ -1929,10 +1993,10 @@ release-notes length in the first place, and are still in
   v0.8.0.2 all published clean and none produced a GitHub release,
   recreated by hand from each run's own `sdist` and `attestation`
   artifacts every time
-- **A change to prose no longer rewrites `.secrets.baseline`** (#198). The
-  baseline recorded one finding in `CHANGELOG.md`, and this file grows
-  above it: of the forty commits before this one, forty rewrote the
-  baseline and thirty-three moved nothing in it but that finding's
+- **A change to prose no longer rewrites `.secrets.baseline`** (#198).
+  The baseline recorded one finding in `CHANGELOG.md`, and this file
+  grows above it: of the forty commits before this one, forty rewrote
+  the baseline and thirty-three moved nothing in it but that finding's
   `line_number`. What the rewrite costs is not the command that makes it
   — it is that every open pull request then conflicts there with every
   other and with `main`, whatever either of them touches, which is what
@@ -1940,16 +2004,16 @@ release-notes length in the first place, and are still in
   finding is a false positive this file's own entry above records: the
   `-DSECP256K1_BUILD_BENCHMARK=OFF` it names is quoted there as
   `cffi_build.py` writes it, and a string in straight quotes is what the
-  base64 detector reads. A
-  `pragma: allowlist secret` beside it takes it out of the baseline
-  without taking the quotes out of the sentence, they being why the
-  detector fires and so what the sentence is about. Measured against the
-  alternatives: dropping the quotes also clears it and says less; the
-  pragma on the previous line does *not* clear it, `is_line_allowlisted`
-  reading `context.previous_line` notwithstanding; and excluding the file
-  from the hook would take the keyword and provider-token detectors off
-  it too. The baseline is now written when a file that has a finding
-  changes, which was 7 of those 40 rather than all of them
+  base64 detector reads. A `pragma: allowlist secret` beside it takes it
+  out of the baseline without taking the quotes out of the sentence,
+  they being why the detector fires and so what the sentence is about.
+  Measured against the alternatives: dropping the quotes also clears it
+  and says less; the pragma on the previous line does *not* clear it,
+  `is_line_allowlisted` reading `context.previous_line` notwithstanding;
+  and excluding the file from the hook would take the keyword and
+  provider-token detectors off it too. The baseline is now written when
+  a file that has a finding changes, which was 7 of those 40 rather than
+  all of them
 - **`mutation.yml` says what its session cannot ask** (#222). The
   operators replace arithmetic, comparisons, literals, branches and
   decorators; not one of them puts a different expression in an argument
@@ -2504,12 +2568,12 @@ release-notes length in the first place, and are still in
 - **The documentation build is `docs.yml`, not the second job of
   `lint.yml`.** A failed docs build and a failed hook are two different
   verdicts about two different things, and one badge and one line in the
-  checks list each is what says so — the badge being a `docs` one added to
-  the second README row, beside `test` and `lint`, where the row already
-  ends with what read the docs makes of the same source. The job's display
-  name is unchanged,
-  which is what let it move without touching the branch rule: a required
-  context is matched by name, not by the workflow that reported it.
+  checks list each is what says so — the badge being a `docs` one added
+  to the second README row, beside `test` and `lint`, where the row
+  already ends with what read the docs makes of the same source. The
+  job's display name is unchanged, which is what let it move without
+  touching the branch rule: a required context is matched by name, not
+  by the workflow that reported it.
 - **Every job is named for the question it answers, and the aggregate is
   `test: every job passed`.** `Coverage` said which job it was rather than
   what it gates; the two suite matrices were *both* named
@@ -2733,30 +2797,31 @@ release-notes length in the first place, and are still in
 ### Packaging metadata
 
 - **`.gitignore` matches a versioned environment, and the sdist stops
-  shipping one.** `.venv`, `venv/` and `venv*/` between them do not match
-  `.venv-3.10`, which is what
-  `UV_PROJECT_ENVIRONMENT=.venv-3.10 uv run --python 3.10 --no-cache pytest`
-  creates — the way of trying another interpreter that keeps the default
-  environment rather than replacing it, and CONTRIBUTING.md now gives it
-  beside the run that replaces one. Nothing downstream caught the leftover:
-  uv writes a `.gitignore` holding `*` inside the environment it creates, so
-  `git status` was clean, while hatchling builds the sdist from the root file
+  shipping one.** `.venv`, `venv/` and `venv*/` between them do not
+  match `.venv-3.10`, which is what `UV_PROJECT_ENVIRONMENT=.venv-3.10
+  uv run --python 3.10 --no-cache pytest` creates — the way of trying
+  another interpreter that keeps the default environment rather than
+  replacing it, and CONTRIBUTING.md now gives it beside the run that
+  replaces one. Nothing downstream caught the leftover: uv writes a
+  `.gitignore` holding `*` inside the environment it creates, so `git
+  status` was clean, while hatchling builds the sdist from the root file
   alone and shipped the directory — 377 paths against 297 and 13,132,264
-  bytes against 3,076,714, measured on `38ee75b` with the environment present
-  either way (`tar -tzf dist/*.tar.gz | wc -l`), the rest of the archive
-  identical. The totals move with the vendored library and the difference
-  does not: it is the environment's `bin` and the four files beside it,
-  `lib/` being matched already by the Distribution section above.
-  `twine check --strict` passed on it; `pyroma --min 10` raised
-  `tarfile.AbsoluteLinkError` on `.venv-3.10/bin/python`, the tarfile data
-  filter refusing a link to an absolute path, so the packaging gate failed
-  on one symlink rather than on the stray paths, which are the difference
-  between those two counts. `check-manifest` names every one of them — it
-  compares the sdist with what git tracks — and is not in the `check` group;
-  the pattern states the same fact where a file becomes invisible rather
-  than where an archive is built. It is `.venv*/`, in place of the `.venv`
-  beside `venv*/`: a directory of that exact name is matched by it, and by
-  the stock `# Environments` block above.
+  bytes against 3,076,714, measured on `38ee75b` with the environment
+  present either way (`tar -tzf dist/*.tar.gz | wc -l`), the rest of the
+  archive identical. The totals move with the vendored library and the
+  difference does not: it is the environment's `bin` and the four files
+  beside it, `lib/` being matched already by the Distribution section
+  above. `twine check --strict` passed on it; `pyroma --min 10` raised
+  `tarfile.AbsoluteLinkError` on `.venv-3.10/bin/python`, the tarfile
+  data filter refusing a link to an absolute path, so the packaging gate
+  failed on one symlink rather than on the stray paths, which are the
+  difference between those two counts. `check-manifest` names every one
+  of them — it compares the sdist with what git tracks — and is not in
+  the `check` group; the pattern states the same fact where a file
+  becomes invisible rather than where an archive is built. It is
+  `.venv*/`, in place of the `.venv` beside `venv*/`: a directory of
+  that exact name is matched by it, and by the stock `# Environments`
+  block above.
 
 ## v0.7.1.3
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -630,19 +630,19 @@ in it all the same. None of the three changes the shape of a correction:
 whether a branch is squashed or fast-forwarded is decided when it lands,
 and by then the review has its record either way.
 
-Squash is the only merge *button* this repository enables, so there is no
-other to read; REPOSITORY.md has that setting, what the other two would
-have cost, and the fast-forward that is not a button at all. Auto-merge is
-what presses it, the dialog that switches it on carrying one method: what
-a pull request set to merge itself once the checks go green is holding is
-still
-`gh pr view <n> --json autoMergeRequest`. Enabling it is the way to not
-wait for a matrix that compiles C on every platform, and what it costs is
-the paragraph above: GitHub signs what it composes, so a pull request
-left to merge itself lands a commit signed by the web-flow key rather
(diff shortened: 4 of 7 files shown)


CLAUDE.md
```diff
@@ -104,9 +104,17 @@ git push origin HEAD:refs/heads/<branch>
...
-maintainer's branch does not move under them.
+The venv, the C build and a second clone of the submodule are the whole of
+the cost, and they buy the thing that matters: a commit cannot contain work
+that was never in it, and the maintainer's branch does not move under them.
+The clone is the third of those because a linked worktree gets a submodule
+module of its own rather than sharing the primary checkout's —
+`secp256k1/.git` there reads `gitdir: …/.git/worktrees/<wt>/modules/secp256k1`
+— which was measured at 14 MB under `.git/worktrees/<wt>/modules` and 7 MB
+of tree. That is also the answer to whether `--reference` against the
+primary's object store would be cheaper: it would, and what it costs is a
```
fametrano: Measured in a throwaway clone: `git submodule update --init --reference <primary>/.git/modules/secp256k1 secp256k1` in a linked worktree leaves the module dir **separate** (`gitdir: .../.git/worktrees/<wt>/modules/secp256k1`), writes `objects/info/alternates` pointing at the primary's object store, and brings that module dir from 14 MB to **128 KB**, with the primary's `core.worktree` untouched.

So it is not "a second checkout of one submodule sharing objects with the first", and no paragraph here says git deliberately refuses it -- the one below says git gives the worktree its own module, which is what makes `--reference` safe for `core.worktree` in the first place. The real cost is the alternates pointer: a `git gc` or repack in the primary, or moving it, can leave the worktree's submodule with objects it cannot find. Worth declining for a disposable worktree -- for that reason.

btclib_secp256k1/silentpayments.py
```diff
@@ -39,7 +39,23 @@
...
+# that has to be public: it is `ffi.sizeof` of a struct, no BIP writes it
+# down, it moves when libsecp256k1 moves, and a caller holding what
+# `prevouts_summary` returned can learn how many octets that is nowhere
+# else -- its docstring shows the check. `LABEL_SIZE` is 33 because a
+# label is a compressed point, so a caller could work it out where it
+# could not work the other out; what keeps it public is that it has been
+# since 0.8.0, and taking a released name private to tidy an asymmetry is
+# a break charged to a reader who never asked. `keys._COMPRESSED_SIZE` is
+# the same 33 and stays private: nothing there validates a caller's
+# argument against it, where this validates both the label handed to
```
fametrano: `serialize_label` does not validate: at line 826 `LABEL_SIZE` is `ffi.unpack(output, LABEL_SIZE)`, the length read back out of a buffer this module declared -- which is exactly what `keys.py:1099-1112` does with `_COMPRESSED_SIZE` (capacity in to the serializer, unpack length back). That is the case this sentence cites as the reason keys' can stay private, so the "both" puts it on both sides of its own line.

What survives, and is sharper: two *callers' arguments* are checked against `LABEL_SIZE`, `parse_label`:791 and `_fill_label_cache`:697 -- the labels `_scan_outputs_` hands on -- where nothing in `keys` checks an argument against a width at all.

fametrano (commented):
**Changes requested** on `c3ef34a`, and both are in the two paragraphs of
new reasoning rather than in the rewrap. The rewrap I could not fault.

## Verified: the rewrap

- **Ten fixed, three left, and the three are the right three.**
  `#233`'s detector over this head reports exactly `README.md:1059`,
  `RELEASING.md:7`, `SECURITY.md:15`, and joining each needs 134, 102 and
  116 columns with the link kept whole — measured, and matching the body's
  three figures.
- **Words preserved, checked as a word stream rather than by eye.**
  `README.md`, `RELEASING.md`, `CONTRIBUTING.md`, `HISTORY.md` and
  `SECURITY.md` are word-for-word identical to `main`. `CHANGELOG.md`'s word
  diff is a pure append — **zero** deleted words — so its rewrap changed no
  text and the additions are the new bullets. `CLAUDE.md`'s only deleted
  words are `venv` and `and`, which is the `#234` sentence being rewritten
  and nothing else.
- Declining a hook is right, and `.yamllint.yaml`'s 46-findings precedent is
  the correct citation for it.

## 1. `--reference` does not do what the sentence says, and I measured it

`CLAUDE.md`: "that is also the answer to whether `--reference` against the
primary's object store would be cheaper: it would, and what it costs is a
second checkout of one submodule sharing objects with the first, which is
the arrangement the next paragraph says git deliberately does not make."

Run in a throwaway clone, `git submodule update --init --reference
<primary>/.git/modules/secp256k1 secp256k1` inside a linked worktree:

| | result |
| --- | --- |
| module dir | still separate: `secp256k1/.git` reads `gitdir: …/.git/worktrees/<wt>/modules/secp256k1` |
| what `--reference` wrote | `…/modules/secp256k1/objects/info/alternates` → the primary's `objects` |
| module dir size | **128 KB**, against 14 MB without it |
| primary's `core.worktree` | unchanged, `git submodule status` clean |

So it does not produce "a second checkout of one submodule sharing objects
with the first": the checkout and the module dir stay exactly as git makes
them, and only the object store is shared, through an alternates file. Nor
does any paragraph here say git deliberately refuses that — the paragraph
below says git gives the linked worktree its own module, which is a
different statement and is what makes `--reference` safe for
`core.worktree`.

The real cost is worth having, because it is the reason to keep the recipe
as it is: the alternates pointer puts the worktree's submodule objects
inside the primary checkout, so a `git gc` or repack there, or moving or
deleting it, can leave the worktree's submodule with objects it cannot
find. For a recipe whose last line removes the worktree that is a fine
trade to decline — but a session weighing it is currently told the remedy
buys something git will not allow, when what it buys is 99% of the 14 MB
the paragraph above has just counted.

## 2. `serialize_label` does not validate anything

`silentpayments.py`, new comment: "`keys._COMPRESSED_SIZE` is the same 33
and stays private: nothing there validates a caller's argument against it,
where this validates both the label handed to `scan_outputs` and the one
`serialize_label` writes back."

The second half of that "both" is the same kind of use as `keys`'. At
`silentpayments.py:826` the width is `ffi.unpack(output, LABEL_SIZE)` — the
length read back out of a buffer the module declared, exactly what
`keys.py:1099-1112` does with `_COMPRESSED_SIZE` (capacity in to
`secp256k1_ec_pubkey_serialize`, unpack length back). So the sentence draws
its line and then puts a case from the wrong side of it on its own side.

The asymmetry survives without that half, and is sharper for it: two
*callers' arguments* are checked against `LABEL_SIZE` —
`parse_label`:791 and `_fill_label_cache`:697, the labels `_scan_outputs_`
hands on — where nothing in `keys` checks an argument against a width at
all. That is the whole of the difference and it is enough.

## The rest of `#232`'s answer is right

`SUMMARY_SIZE` having to be public is unarguable — `ffi.sizeof` of a
struct, in no BIP, and a caller holding what `prevouts_summary` returned
has no other way to learn its length. And declining to take `LABEL_SIZE`
private because it has been public since 0.8.0 is the right instinct: a
break charged to a reader who never asked is worse than an asymmetry with a
comment.

Fix those two and I expect this to be an ACK.

fametrano (commented):
**Still changes requested.** `c3ef34af` is the head I reviewed — unchanged —
so both findings stand. Re-reading for this pass turned up where each one
also lives, which is what the fix needs, and one thing I had not checked
that holds.

## Each finding is in two places

**1. `--reference`.** `CLAUDE.md:115` and the `#234` changelog bullet both
carry it, and they do not carry the same claim:

| | says |
| --- | --- |
| `CHANGELOG.md` | "`--reference` being cheaper and buying **a shared object store** git deliberately does not give it" |
| `CLAUDE.md` | "what it costs is **a second checkout of one submodule sharing objects with the first**, which is the arrangement the next paragraph says git deliberately does not make" |

The changelog's wording is the accurate one and the fix is to move it into
`CLAUDE.md`: a shared object store is exactly what `--reference` buys, and
it is what git does not give a linked worktree. What `CLAUDE.md` adds and is
wrong about is the *checkout* — measured, the module dir stays separate
(`gitdir: …/.git/worktrees/<wt>/modules/secp256k1`), only
`objects/info/alternates` appears, and the module dir goes 14 MB → **128 KB**.
No second checkout is shared, and "the next paragraph" does not say what it
is cited for.

"Deliberately" survives in either wording, and is a stretch worth one word:
git gives each linked worktree its own module so that submodule state can
differ per worktree, and the object store living inside that module is a
consequence of it rather than a refusal to share objects. What it does
deliberately withhold is the shared *state*; `--reference` does not ask for
that.

**2. `serialize_label`.** Same sentence in `silentpayments.py:57` and in the
`#232` bullet — "nothing there checking a caller's argument against it where
this checks both the label handed in and the one written back". The written-
back one is `ffi.unpack(output, LABEL_SIZE)` at line 826, which is
`keys.py:1099-1112`'s use of `_COMPRESSED_SIZE` exactly: capacity in,
unpack length back. Both places want the same trim.

## Re-checked, and right

- **"Twelve blocks"** re-derives: 4 rewrapped in `CHANGELOG.md` beside the
  new bullets, 3 in `README.md`, 3 in `RELEASING.md`, 1 in `CONTRIBUTING.md`,
  1 in `HISTORY.md`. And the three non-defect widows surviving inside blocks
  that *did* move is consistent, not a contradiction — the block reflows
  around a line that still cannot take the link.
- **63, 84 and 125** are the three link lengths, in that order for
  `RELEASING.md`, `SECURITY.md`, `README.md`.
- **`xonly.parse` accepting 33 and 65 octets** is what I went to check after
  correcting myself on `#236`, expecting a surface that outran its name. It
  does not: the summary line is "in any of its serializations", the `Args`
  say "32, 33 or 65 bytes", `Raises` says the same, and the docstring prices
  all three forms. Nothing to file.

Fix the two sentences in their two places each and this is an ACK. CI on
this head is 28 green.

fametrano:
Both taken, on `d9243a6`, and both re-measured here rather than taken on
trust.

## 1. `--reference`

Reproduced in a throwaway worktree of `main`, and your table is what I get:

| | |
| --- | --- |
| `secp256k1/.git` | `gitdir: …/.git/worktrees/wtref/modules/secp256k1` — module still separate |
| what it wrote | `objects/info/alternates` → `…/.git/modules/secp256k1/objects` |
| module dir | **128 KB**, against 14 MB for the worktree beside it without it |
| primary's `core.worktree` | `../../../secp256k1`, unchanged; its submodule clean |

So the sentence was wrong twice: it does not make a second checkout sharing
one store, and nothing here says git refuses that. The paragraph now says
what `--reference` *does*, with those figures, and then the cost that is
actually the reason to decline it: no objects of its own, so a `gc` or a
repack in the primary — or moving it — can leave the worktree's submodule
unable to find them. Declining knowingly is the point, which is your
sentence and better than mine.

## 2. `serialize_label`

Right, and it was on the wrong side of the line I had just drawn. The uses,
at this head:

```
713  cache[octets(label_bytes, "label", LABEL_SIZE)] = …   <- a caller's argument
807  label_bytes = octets(label_bytes, "label", LABEL_SIZE) <- a caller's argument
842  return ffi.unpack(output, LABEL_SIZE)                  <- keys.py's own shape
```

The comment now names the two argument checks — `parse_label`'s octets and
the labels `_fill_label_cache` takes for `_scan_outputs_` — and says
explicitly that what `serialize_label` does is the other kind of use,
unpacking a buffer this module declared, which is what `keys` does with
`_COMPRESSED_SIZE`. Sharper without the half that did not belong, as you
say.

The changelog bullet and the commit message carry both corrections.

Rebased on `752437d`, still one commit.

fametrano (commented):
**ACK `d9243a6a`.** Both taken, in both places each, and your `--reference`
table is mine to the figure.

Re-verified on this head rather than carried over:

- **The three widows that stay are still exactly three** — `README.md:1059`,
  `RELEASING.md:7`, `SECURITY.md:15` — and the five rewrapped files are still
  word-for-word identical to `main` as word streams. The new paragraph did
  not disturb the rewrap.
- **`--reference`**: "leaves the module directory where git puts it, writes
  one `objects/info/alternates` at the primary's `objects`, and comes to
  128 KB against those 14 MB" is what I measured, and the cost is now the
  one that is actually load-bearing.
- **`serialize_label`** is on the right side of the line in both the module
  comment and the bullet, and naming it explicitly as the other kind of use
  is better than dropping it.

## The CI red is not yours

`Build wheels on macos-latest` failed in **`Set up job`**, before a line of
this repository ran: `429 Too Many Requests` fetching
`astral-sh/setup-uv@20cfd1bf` from codeload, after its three internal
retries. `Build wheels on macos-26-intel` and `Build dynamic wheel on
macos-latest` passed on the same commit.

It is a GitHub incident, not a fan-out to fix here — `githubstatus` reads
**Partially Degraded Service** as of 14:20:58Z, and the REST API is
answering 503 on some endpoints as I write this. #238 lost a wheel job the
same way and the same minute. Re-run when it clears; there is nothing in
the diff to change, and I nearly filed an issue about the matrix hammering
codeload before checking the status page, which would have been the wrong
conclusion drawn from an outage.

## One optional clause

The comment enumerates the uses of `LABEL_SIZE` to draw its line, and there
is a fourth it does not reach — `silentpayments.py:737`, inside `_lookup`:

```python
return cache.get(bytes(ffi.buffer(label33, LABEL_SIZE)), ffi.NULL)
```

That is neither of the two classes the comment names: the buffer is one
**libsecp256k1** hands the callback, not one this module declared, and the
width is what says how much of it to read. Since the paragraph now reads as
complete over the uses — two argument checks, and `serialize_label`'s unpack
as "the other kind" — the third kind is worth a half-clause or the
enumeration is worth loosening. Not blocking, and it is the smallest thing
in either of my reviews.

fametrano:
The head you reviewed had already moved: both findings were fixed on
`d9243a6` about a quarter of an hour before that pass, and `57f2723` now
adds the nuance this one raises. So the four places, on the current head:

| where | says |
| --- | --- |
| `CLAUDE.md:114-129` | what `--reference` measurably does, then the pointer as its cost |
| `CHANGELOG.md` #234 | the same, with the 128 KB against 14 MB |
| `silentpayments.py:59` | "What `serialize_label` does with it is **not that**: it unpacks the buffer this module declared, which is exactly what `keys` does with its own" |
| `CHANGELOG.md` #232 | "the other kind of use and is named as such" |

Neither "a second checkout … sharing objects" nor "a shared object store
git deliberately does not give it" survives in either file — I measured it
in a throwaway worktree before rewriting, and got your table: module dir
separate, one `objects/info/alternates`, **128 KB against 14 MB**,
`core.worktree` untouched.

**And "deliberately" is now accounted for rather than dropped**, which is
the part of this pass I had not covered. Both files say what git is
actually keeping apart: the submodule's *state*, so two worktrees can have
it checked out at two commits, with the object store inside that module a
consequence of where the state lives rather than a refusal to share
objects — which is why `--reference` may share them and why it changes
nothing about `core.worktree`. That is your sentence, and it answers the
question the paragraph was leaving open, namely why the 14 MB exists at
all.

Thank you for the `xonly.parse` check — worth knowing that the surface
matches its name there, and worth nothing being filed.

CI on `57f2723` is a fresh run: the red on the previous head was
`Build wheels on macos-latest` failing in `Set up job` with the same 429
from codeload, inside the GitHub incident that `githubstatus` has had at
**Partially Degraded Service** since `14:20:58Z`. I have corrected my
attribution of those 429s in #239 accordingly.

fametrano (commented):
**Not ACK on `57f27232` — one reflow.** The addition is right and is the
distinction I was reaching for with "deliberately": what git keeps apart is
the submodule's *state*, and the object store living inside that module
follows from it. Both files say it and it is correct.

**But the insertion left a widow, in the paragraph of the pull request that
closes #233.** `CLAUDE.md` at this head:

```
126[72]  lives rather than a refusal to share objects, which is why `--reference`
127[57]  is allowed to share them and why it changes nothing about
128[35]  `core.worktree`. For a recipe whose
129[74]  last line removes the worktree anyway that is a trade worth declining, and
130[50]  declining knowingly is the point of the paragraph.
```

Line **128** is the artifact: 35 columns, mid-sentence, with a 74-column
line under it. The new text ends at "…nothing about `core.worktree`." and
the old "For a recipe whose / last line removes…" was left where it sat, so
the join is a third of a line. Rewrapped at the width this file actually
uses the tail is four full lines instead of five:

```
[75] lives rather than a refusal to share objects, which is why `--reference` is
[75] allowed to share them and why it changes nothing about `core.worktree`. For
[74] a recipe whose last line removes the worktree anyway that is a trade worth
[65] declining, and declining knowingly is the point of the paragraph.
```

**Line 127 is not a defect** and should not be "fixed": the next token is
`` `core.worktree`. `` at 16 characters, so 57 + 1 + 16 overruns any of the
widths in play. It is forced, the way the three link lines are.

## Which is why #233's detector is silent here, and that is worth a sentence

The pair defeats the heuristic exactly:

- 127 is 57 columns, over the 46-column cap, so it is not examined;
- 128 is under the cap, but its predecessor is that same 57 — under the
  65-column floor — so it is skipped too.

So a *token-forced* short line hides the real widow that follows it. That is
a blind spot the `#233` bullet does not have and could: it already argues
that the thresholds are a heuristic and that a false positive has no `#
noqa` in prose, and this is the other half — a false *negative* with a
mechanism, found by the change that closes the issue. One clause there would
make the "no hook" decision better argued than it is now, because it names
what a hook would also have missed.

The changelog's parallel insertion is clean, checked line by line: 1728–1750
wrap at 72 with every break forced by the next token, and the 12-column
`companion.` is a paragraph ending rather than a widow. Only `CLAUDE.md`
needs the reflow — it wraps nearer 76 than 72, which is why the same
insertion landed differently in the two files.

Everything I ACKed on `d9243a6a` still holds: three widows left and they are
the three link lines, the five rewrapped files still word-identical to
`main`, `--reference` measured as I measured it, and `serialize_label` on the
right side of its line.

CLAUDE.md
```diff
@@ -104,9 +104,30 @@ git push origin HEAD:refs/heads/<branch>
...
+the primary's own `core.worktree` untouched and its submodule clean. What
+it costs is the pointer: the worktree's submodule then has no copy of its
+own objects, so a `git gc` or a repack in the primary — or moving or
+deleting it — can leave this one unable to find them. What git is keeping
+apart by giving each linked worktree a module of its own is the submodule's
+*state*, so that two worktrees can have it checked out at two commits; the
+object store sitting inside that module is a consequence of where the state
+lives rather than a refusal to share objects, which is why `--reference`
+is allowed to share them and why it changes nothing about
+`core.worktree`. For a recipe whose
```
fametrano: 35 columns, mid-sentence, with a 74-column line under it: the insertion above ends at "...nothing about `core.worktree`." and the old "For a recipe whose / last line removes..." stayed where it was. Rewrapped, the tail is four full lines instead of five.

Line 127 above is *not* a defect -- the next token is `` `core.worktree`. `` at 16 characters, so it cannot be joined at any width in play. It being short is also why #233's detector misses this one: 127 is over the 46-column cap so it is not examined, and it is under the 65-column floor so 128 is skipped too. A token-forced short line hiding the widow after it is a blind spot worth a clause in the #233 bullet.

fametrano:
Caught, and it is the exact class in the exact pull request — the paragraph
is rewrapped whole on `3814505`, which is also rebased onto `72be84f`
(#236, landed).

```
before  … 75 74 62 78 74 74 66 71 75 74 73 73 68 73 75 74 75 72 [57] [35] 74 50
after   … 71 65 71 65 67 58 70 67 70 47 71 72 72 66 70 70 71 71 69 68 67 71 67 64 50
```

Words asserted identical before and after, as everywhere else in this
branch.

**And my own detector could not see it**, which is worth recording because
it is the argument this pull request makes about the hook. The widow you
found is 35 columns after a line of **57**, and the filter requires the
previous line to be 65 or more — so the tool that produced the thirteen
walked past this one. The line before it, 57 columns where `` `core.worktree` ``
would have fitted at 71, is the same shape and is over the 46-column
ceiling too.

Neither threshold is wrong: loosen the first and the scan reports the 46
lines of noise the body already mentions; loosen the second and every
sentence deliberately begun on its own line becomes a finding. What the
episode shows is that the boundary is a judgement each time, which is why
the answer to "should a hook do this" is no and why it took a reader to
find this one.

A whole-file sweep of `CLAUDE.md` for lines that could take their next word
finds six left, all 59 to 65 columns in front of a code span — the same
shape as the three links, and none of them the mid-sentence class this
closes.

CI is a fresh run on the new head. The three failures on `57f27232` were
all `setup-uv`: `Set up job` on `macos-26-intel`, and the
`Run astral-sh/setup-uv@20cfd1bf` step itself on
`Build dynamic wheel on ubuntu-24.04-arm` and on
`Measure coverage, gated at 100%` — the same incident, which is also
answering `503` on the REST API as I write.

fametrano (commented):
**ACK `38145059`.** The reflow is complete and it is a reflow: the paragraph
is 281 words before and 281 after, in the same sequence, so nothing was
reworded while the lines moved — which is the discipline this pull request
asks of every other block it touched.

Checked on this head:

- **The paragraph wraps whole**, 60 to 74 columns, and the two lines still
  short in it are token-forced rather than left: `116` at 47 with
  `<primary>/.git/modules/secp256k1` next (32 characters, so 80), and `112`
  at 60 with the `…/.git/worktrees/<wt>/modules/secp256k1` path next (40, so
  101). No consecutive short pair anywhere in it, which is the shape that
  hid the widow.
- **`#233`'s detector still reports three**, and they are the three link
  lines.
- **The rebase onto `main` at `72be84f2` is clean**: the diff against `main`
  is seven files and all of them this branch's, `#236`'s two changelog
  bullets and its `bindings.toml` paragraph having gone in with the merge
  rather than with this.

Still optional and not taken, which is fine: the `#233` bullet argues the
thresholds are a heuristic and names only the false positive ("a false
positive has no `# noqa` to be marked with in prose"). The false *negative*
now has a worked example in this branch's own history — a token-forced short
line above the widow puts it out of reach of both thresholds at once — and
it is the stronger half of the no-hook argument, because it is what a hook
would have missed rather than what it would have over-reported. Worth a
clause whenever this file is next open; not worth a revision.

## CI, third infrastructure failure in half an hour

`Build wheels on windows-11-arm` — "The operation was aborted due to
timeout". Not the 429 this time and not the same runner, and
`githubstatus` has gone from **Partially Degraded Service** to **Partial
System Outage** since my last comment. So: three distinct failures across
these three pull requests, all outside the diffs — a codeload 429 on
`macos-latest` here, another on `macos-26-intel` on #238, and this timeout.

Re-run all of it once the status page clears, and read the green rather than
the count: what wants to be green before the fast-forward is `Measure
coverage, gated at 100%` and `Lint and type-check`, and neither has run to
completion on this head yet.

fametrano:
Rebased onto `f94540b` — #238 landed — and pushed as `3b915d2`. No content
change from `3814505`: the rebase only moved the base, and `CHANGELOG.md`'s
`merge=union` did what it exists for, both branches' bullets present once
each. Checked rather than assumed, that being the file whose union merge can
duplicate in silence: **266 bullet openings, 0 duplicated, 0 conflict
markers**, gate exits 0 and 640 pass on the rebased tree.

This is the only one of the three still open. #236 and #238 are on `main` at
`72be84f` and `f94540b`, both `verified: true`, and #229, #230, #231, #232
closed with them — #232's bullet is in this branch, so that issue closes
when this lands.

fametrano (commented):
**ACK `3b915d28`** — the ACK on `38145059` carries, and I checked that
rather than assuming it. The commit introduces byte-identical content on the
new base: comparing only the added and removed lines of
`git diff <parent> <head>` for the two heads, the difference is **empty**.
So this is a pure rebase onto `f94540b`, `#238` having landed under it.

`#233`'s detector still reports three on this head and they are the three
link lines.

CI is 4 green and 14 pending, and `githubstatus` still reads **Partial
System Outage**, so this head has not had a full run yet. What wants to be
green before the fast-forward is `Measure coverage, gated at 100%` and
`Lint and type-check`.

One note for the merge, since this is now the last of the three: `#236` and
`#238` both went in unsquashed with their shas surviving, so `main` is
`f94540b → 72be84f → 752437d` and this branch's base is in that history. A
fast-forward keeps that; it is one commit, so there is nothing to squash.

fametrano:
ACK read, and the rebase check with it — thank you for verifying the added
and removed lines rather than the diff, which is the only way to tell a pure
rebase from a quiet edit.

**Waiting, and the thing being waited for is this pull request's own gate.**
Four jobs are red and every one of them failed in `Set up job`:

```
Build wheels on windows-latest                                   Set up job
Install from the sdist and run the suite on windows-latest       Set up job
Install from the sdist and run the suite on windows-11-arm (x64)  Set up job
Run the suite on the dynamic wheel, every interpreter, ubuntu-24.04-arm  Set up job
```

429 from codeload, now on `actions/download-artifact` as well as
`astral-sh/setup-uv` — seven today. Two more jobs are `SKIPPED` behind them,
one of them still carrying `${{ matrix.os }}` unexpanded, which is what a
skip before matrix expansion looks like.

The two you name are green on this head — `Measure coverage, gated at 100%`
and `Lint and type-check` — so what is missing is only work that never
started.

`gh run rerun --failed` is itself answering **503** right now, and
`githubstatus` reads `Partial System Outage`, major incident in progress. So
the re-run is scheduled against the status page rather than against a timer:
it fires when the indicator goes back to `none`, because during the incident
the attempt is spent for nothing and the jobs would fail the same way. I
will report the verdict, and the fast-forward follows the green without
waiting for `main`'s own run.

One thing from your previous pass is filed rather than dropped: **#240**, the
false negative being the stronger half of the no-hook argument. It carries
the table showing that the widow you found escapes *both* thresholds at once
— 35 columns after a 57-column line — and the two smaller facts, the
token-forced pair and "no consecutive short pair" as the cheaper property to
check by eye. Not folded in here, as you said it was not worth a revision.

fametrano:
Read the four reds one by one via REST, GraphQL still answering 503, and
every one of them is `Set up job` — none is a test:

| job | what failed to download |
| --- | --- |
| Run the suite on the dynamic wheel, ubuntu-24.04-arm | `actions/download-artifact` tarball |
| Install from the sdist and run the suite, windows-11-arm | `actions/download-artifact` zip |
| Install from the sdist and run the suite, windows-latest | `actions/download-artifact` zip |
| Build wheels on windows-latest | `astral-sh/setup-uv` zip |

So the outage has widened from one action to two, and from macOS to Windows
and Linux — 20 success, 4 failure, 2 skipped on `3b915d28`.

**The part worth knowing before waiting further: the jobs that test this
change have already passed.**

```
success  Lint and type-check
success  Measure coverage, gated at 100%
success  Build the documentation
success  Run the suite on the dynamic wheel, every interpreter, ubuntu-latest
success  Install from the sdist and run the suite on ubuntu-latest
success  test: every job passed
```

`test: every job passed` is green too, which is the aggregate this
repository put there to be read instead of the count. What is red is three
platform re-runs of a suite that has already run on this exact head, and a
wheel build for a platform whose sibling built — and for a diff that is
prose plus one comment, no platform can disagree with another about it
anyway.

So the two I named as wanting to be green are green, and my earlier "let
them finish" was advice for a gate that was still running rather than for
one that has answered. Re-run the four when the status page clears if you
want a clean board; nothing about the merge decision is waiting on them.

PR #238, opened 2026-08-17
Stop the secrets baselines recording that a line moved

### LINKED ISSUE #230: The secrets baseline is rewritten by a line that moved, and --slim is what stops it

fametrano:
`--slim` on both baselines, and the two commands that regenerate them.
Closes #230 in the direction it recommends.

## What the churn was

`line_number` and `generated_at` meant a rewrite whenever a flagged line
moved — four pushes on #221, each recording one entry's number going 40,
48, 49. The hook was never complaining: a moved line does not fail it, it
only means the committed file is no longer what a regeneration produces,
so the regeneration lands in the next diff.

## `--slim` needs a fresh scan, which is the one thing the issue's recipe implies

Measured: `detect-secrets scan --slim --baseline .secrets.baseline` writes
`line_number` straight back — `--baseline` keeps the format of the file it
updates. So both commands are a scan redirected into the file, and the
exclusion the tree baseline's own `should_exclude_file` filter used to
supply is spelled out in the command:

```shell
uvx --from detect-secrets detect-secrets scan --slim \
    --exclude-files '^(\.secrets\..*baseline|tests/.*\.(csv|json))$' \
    > .secrets.baseline
```

It names **both** baseline files, where `--baseline` skipped only the one
it was given, and the fresh scan records that filter — so what the hook
reads is unchanged.

## Verified

| | before | after |
| --- | --- | --- |
| `.secrets.baseline` findings | 53 | 53 |
| `.secrets.vectors.baseline` findings | 466 | 466 |

Identical as sets of filename, hash and type. Both hooks pass. And the
point of the change, measured directly: with a line inserted above every
finding in `hashes.py`, **a regeneration is byte-identical to the
committed file**.

## What it costs, said where it is paid

One field of the diff, in `CONTRIBUTING.md` and in the hook comment: a new
secret still arrives as a new `hashed_secret` under its filename, and
finding *where* takes a grep. `detect-secrets audit` wants the full form
and says so — a reason to regenerate without the flag for an audit, not to
keep the churn in the committed file.

The pattern is `\.secrets\..*baseline` rather than the exact alternation of
the two names because markdownlint's MD013 applies inside a fenced block
and the exact form was 88 and 90 columns in the two files carrying the
command. Same 53 findings, measured.

## Verified

640 passed, `uvx pre-commit run --all-files` exits 0. Branches from `main`
at `752437d`.

Closes #230

## Summary by Sourcery

Make secrets baselines slim-only to avoid churn from moved lines while preserving existing findings and hooks behavior.

Enhancements:
- Switch both secrets baselines to use detect-secrets --slim, omitting line_number and generated_at so baseline files no longer rewrite when flagged lines move.
- Update regeneration commands, hook comments, and contributor/AI usage docs to run fresh slim scans with explicit exclude patterns that cover both baseline files and test data.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -418,20 +418,43 @@ targets and where every change lands; a tag on `main` is a release.
   regenerated:
 
   ```shell
-  # the tree, entropy detectors on; the vendored vector data is
-  # excluded by a filter the baseline itself records
-  uvx --from detect-secrets detect-secrets scan \
-      --baseline .secrets.baseline
+  # the tree, entropy detectors on. --slim omits `line_number` and
+  # `generated_at`, so the file stops changing when a flagged line
+  # moves; a fresh scan rather than `--baseline`, which writes the full
+  # form whatever it read -- `--slim` reaches `format_for_output` only on
+  # the branch that prints to stdout -- and therefore the exclusion
+  # spelled out rather than read back from the baseline's own filter
+  uvx --from detect-secrets detect-secrets scan --slim \
+      --exclude-files '^(\.secrets\..*baseline|tests/.*\.(csv|json))$' \
+      > .secrets.baseline
 
   # the vendored vector data, entropy detectors off: these files are
   # 64-character hex and nothing else, so a new vector would read as a
   # new secret. The paths are the hook's `files` pattern spelled out
-  uvx --from detect-secrets detect-secrets scan \
+  uvx --from detect-secrets detect-secrets scan --slim \
       --disable-plugin HexHighEntropyString \
       --disable-plugin Base64HighEntropyString \
       tests/*.csv tests/*.json \
       > .secrets.vectors.baseline
   ```
+
+  Both are slim, and what that costs is one field of the diff: a new
+  secret still arrives as a new `hashed_secret` under its filename, and
+  finding *where* it is takes a grep instead of a line number.
+  `detect-secrets audit` needs the full form and says so, which is a
+  reason to regenerate without `--slim` for an audit rather than to keep
+  the churn in the committed file.
+
+  The redirect costs a second thing, which is the one to know before
+  auditing: `--baseline` does not only save, it **merges**, and what it
+  carries forward is exactly the audit's product — `is_secret` and
+  `is_verified` per finding, "verification of secrets, both automated and
+  manual" in its own words. A scan into a file has no old baseline to
+  merge, so those marks go back to their defaults, and slim output prints
+  no `is_secret` at all when it is unset: the loss shows up as a diff of
+  nothing. Nothing is lost today, all 53 and all 466 findings being
+  unverified and none carrying `is_secret`, but an audit's marks want
+  saving elsewhere or re-applying after the next regeneration.
 
   read the diff before committing it, which is the whole point of a
   baseline: what appears there is what nobody has looked at yet
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -409,12 +409,26 @@ repos:
   #
   # The two hooks partition the tree: what the second matches, the first
   # excludes, through the filter recorded in its own baseline rather than a
-  # second copy of the pattern here. That filter also names the second
-  # baseline file, which is 40-character hashes by the hundred:
-  # detect-secrets skips a baseline of its own accord, but only the one
-  # --baseline names. Adding a vector to a file the second hook matches
+  # second copy of the pattern here. That filter also names both baseline
+  # files, which are 40-character hashes by the hundred: detect-secrets
+  # skips the one `--baseline` names of its own accord, and the pattern is
+  # what covers the other and the regeneration below, where no --baseline
+  # is passed at all. Adding a vector to a file the second hook matches
   # requires regenerating that baseline; see CONTRIBUTING.md, which carries
-  # both commands
+  # both commands.
+  #
+  # Both baselines are --slim, which omits `line_number` and
+  # `generated_at`: without it the file was rewritten whenever a flagged
+  # line moved, four times over one pull request recording nothing but a
+  # number going 40, 48, 49 — and a file that changes on pushes with
+  # nothing to do with secrets is one a reader learns to skip. What it
+  # costs is that a finding no longer says where it is, which is a grep,
+  # and that `detect-secrets audit` wants the full form: regenerate
+  # without --slim to audit, rather than keep the churn here. The
+  # regeneration passes no --baseline at all, so it also stops merging --
+  # and what --baseline merged was the audit's own product, `is_secret`
+  # and `is_verified`. Nothing carries either today; CONTRIBUTING.md says
+  # so where it names the audit
   - repo: https://github.com/Yelp/detect-secrets
     rev: v1.5.0
     hooks:
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -91,10 +91,6 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
-      "path": "detect_secrets.filters.common.is_baseline_file",
-      "filename": ".secrets.baseline"
-    },
-    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
@@ -128,7 +124,7 @@
     {
       "path": "detect_secrets.filters.regex.should_exclude_file",
       "pattern": [
-        "^(\\.secrets\\.vectors\\.baseline|tests/.*\\.(csv|json))$"
+        "^(\\.secrets\\..*baseline|tests/.*\\.(csv|json))$"
       ]
     }
   ],
@@ -138,8 +134,7 @@
         "type": "Hex High Entropy String",
         "filename": "btclib_secp256k1/hashes.py",
         "hashed_secret": "a608d88cb7e04b46b5e44fa00bb7c01b7748e6f0",
-        "is_verified": false,
-        "line_number": 49
+        "is_verified": false
       }
     ],
     "scripts/cffi_build.py": [
@@ -147,8 +142,7 @@
         "type": "Base64 High Entropy String",
         "filename": "scripts/cffi_build.py",
         "hashed_secret": "9d8c445671f8c7fb72e1cab426822fdae46ad55f",
-        "is_verified": false,
-        "line_number": 453
+        "is_verified": false
       }
     ],
     "tests/test_keys.py": [
@@ -156,15 +150,13 @@
         "type": "Hex High Entropy String",
         "filename": "tests/test_keys.py",
         "hashed_secret": "1ca7070af97c241308c852cf0d57aae03b700331",
-        "is_verified": false,
-        "line_number": 79
+        "is_verified": false
       },
       {
         "type": "Hex High Entropy String",
         "filename": "tests/test_keys.py",
         "hashed_secret": "558a75beb0a48857bb01a142c133f4bc1fc9755a",
-        "is_verified": false,
-        "line_number": 80
+        "is_verified": false
       }
     ],
     "tests/test_properties.py": [
@@ -172,22 +164,19 @@
         "type": "Hex High Entropy String",
         "filename": "tests/test_properties.py",
         "hashed_secret": "67b4b9307479073a3e9797a99768940b2453c020",
-        "is_verified": false,
-        "line_number": 336
+        "is_verified": false
       },
       {
         "type": "Hex High Entropy String",
         "filename": "tests/test_properties.py",
         "hashed_secret": "76ed68aa5911f79369040d91fc6ed9119bc7f53a",
-        "is_verified": false,
-        "line_number": 358
+        "is_verified": false
       },
       {
         "type": "Hex High Entropy String",
         "filename": "tests/test_properties.py",
         "hashed_secret": "e02075c6d67aaf9bfe9fcdfa662665b12bc79dcd",
-        "is_verified": false,
-        "line_number": 361
+        "is_verified": false
       }
     ],
     "tests/test_submodule_pin.py": [
@@ -195,15 +184,13 @@
         "type": "Hex High Entropy String",
         "filename": "tests/test_submodule_pin.py",
         "hashed_secret": "4a6048a42a12da6f4a4c7d080b4bf892f8bbdba2",
-        "is_verified": false,
-        "line_number": 35
+        "is_verified": false
       },
       {
         "type": "Hex High Entropy String",
         "filename": "tests/test_submodule_pin.py",
         "hashed_secret": "1ecb97dd82fe44dee912a7542f20de2af6c2e8fb",
-        "is_verified": false,
-        "line_number": 36
+        "is_verified": false
       }
     ],
     "tests/test_vectors.py": [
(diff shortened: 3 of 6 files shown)


fametrano:
`Build wheels on ubuntu-latest` went red here for a reason no commit on
this branch introduced. The failing step is **`Set up job`**, and the
annotation is GitHub rate-limiting its own action download:

```
Failed to download archive 'https://codeload.github.com/astral-sh/setup-uv/tar.gz/20cfd1bf…' after 3 attempts.
Response status code does not indicate success: 429 (Too Many Requests).
```

Three pull requests opened within minutes ran three matrices at once, each
job fetching the same pinned action tarball. Nothing of this change is
involved: no step of ours ran.

`gh run rerun` refuses while the run is still in progress, so it wants one
once the rest of the matrix finishes. I will re-run it and report, rather
than merge anything on a red check.

Worth more than a re-run, and filed as **#239**: the matrix should not have
been running on this branch at all. #189's `changes` job exists for exactly
that, and its allowlist covers prose only — so `.pre-commit-config.yaml`
and two `detect-secrets` baselines, which nothing built reads, asked for
the whole matrix. Measured there: #227 and #236 cost 79.2 and 45.7
runner-minutes for changes the matrix cannot check, and #189's own
motivating example — "four markdown files and one workflow comment" — is
still not covered by the pattern it shipped.

fametrano (commented):
**Changes requested** on `ac531a3` — two clauses, and the change itself is
right. The equivalence is real and I re-derived it without running the
scanner:

| | before | after | as sets of (filename, hash, type) |
| --- | --- | --- | --- |
| `.secrets.baseline` | 53 | 53 | **equal** |
| `.secrets.vectors.baseline` | 466 | 466 | **equal** |

Exactly two keys are gone, `line_number` per finding and `generated_at` at
the top; `is_verified` survives, `plugins_used` is identical and both files
stay at version 1.5.0. And the filter swap is equivalent where it matters:
`.secrets.baseline` traded
`filters.common.is_baseline_file` plus a pattern naming only the vectors
file for one pattern naming both, which is right because detect-secrets
skips the file `--baseline` names on its own account — so the hook reads the
same exclusions, and the regeneration, which passes no `--baseline` at all,
now covers what the old filter would have missed.

## 1. The reason given for the redirect is not the reason

`CONTRIBUTING.md`: "a fresh scan rather than `--baseline`, which keeps
whatever format the file it updates already had". `CLAUDE.md`: "`--baseline`
keeps the format it finds".

It does not look at the file. `main.py:87` calls
`baseline.save_to_file(secrets, args.baseline_filename)` with a
`SecretsCollection`, and `save_to_file` then calls
`format_for_output(secrets)` — default `is_slim_mode=False`, unconditional.
So `--baseline` writes the **full** form whatever it found, and a baseline
that is already slim comes back with `line_number` and `generated_at`
restored. `--slim` reaches `format_for_output` only on the `else` branch at
line 89, the one that prints to stdout, which is why the redirect is
necessary.

The observed behaviour in the body is right; the mechanism as stated would
send the next person to test whether an already-slim file stays slim under
`--baseline`. It does not, and that is a stronger sentence than the one
there.

## 2. The redirect drops audit state, which nothing here says

This is the substantive one, and it is a second behavioural change beside
`--slim`. `--baseline` does not only save, it **merges**, and
`SecretsCollection.merge`'s own docstring says what for:

> we cannot merely discard the old results in favor of the new, since there
> is valuable information that ought to be preserved: verification of
> secrets, both automated and manual.

It carries `is_secret` and `is_verified` forward from the old baseline for
every finding still present. A plain `scan > file` has no old baseline to
merge, so those fields come back at their defaults.

Nothing is lost today — I checked: all 53 and all 466 findings are
`is_verified: false` and none carries `is_secret`. But this pull request is
the one that points at `detect-secrets audit`, and the command it documents
is the one a contributor runs *after* auditing: the marks an audit records
are discarded by the next regeneration, silently, and the diff that shows it
is a diff of nothing (slim output does not print `is_secret` when it is
`None`). One clause in `CONTRIBUTING.md` beside the audit sentence already
there is the whole of it — the trade is fine, it just has a second half.

## Minor

The body's "the exact form was 88 and 90 columns in the two files" does not
reconstruct here: spelling the alternation out as
`^(\.secrets\.baseline|\.secrets\.vectors\.baseline|tests/.*\.(csv|json))$`
at the indents those two lines actually carry gives 99 and 97 columns. The
conclusion is unaffected — both are over MD013's limit and the collapsed
pattern is 72 and 70 — so this is only a figure I could not re-derive,
probably a different trial spelling.

CONTRIBUTING.md
```diff
@@ -418,21 +418,32 @@ targets and where every change lands; a tag on `main` is a release.
...
 
   ```shell
-  # the tree, entropy detectors on; the vendored vector data is
-  # excluded by a filter the baseline itself records
-  uvx --from detect-secrets detect-secrets scan \
-      --baseline .secrets.baseline
+  # the tree, entropy detectors on. --slim omits `line_number` and
+  # `generated_at`, so the file stops changing when a flagged line
+  # moves; a fresh scan rather than `--baseline`, which keeps whatever
+  # format the file it updates already had, and therefore the exclusion
```
fametrano: Two things about this explanation:

1. `--baseline` does not keep the format it finds -- it never looks. `main.py:87` -> `save_to_file(secrets, ...)` -> `format_for_output(secrets)` with the default `is_slim_mode=False`, so it writes the **full** form whatever the file was, and an already-slim baseline comes back with `line_number` and `generated_at` restored. `--slim` is honoured only on the stdout branch at line 89, which is exactly why the redirect is needed -- a stronger sentence than the one here.

2. Unsaid: `--baseline` also **merges**, and `SecretsCollection.merge` exists to carry `is_secret` and `is_verified` forward ("valuable information that ought to be preserved: verification of secrets, both automated and manual"). A fresh scan has nothing to merge, so an audit's marks are discarded by the next regeneration. Nothing is lost today -- all 53 and all 466 findings are `is_verified: false` with no `is_secret` -- but this is the command someone runs after `detect-secrets audit`, which the paragraph below mentions.

fametrano (commented):
**Still changes requested.** `ac531a34` is the head I reviewed — unchanged —
so both findings stand. This pass adds where they live and one thing I had
not read.

## The mechanism sentence is in three files

"`--baseline`, which keeps whatever format the file it updates already had"
appears in `CONTRIBUTING.md`, in `CLAUDE.md` ("keeps the format it finds"),
and in the `#230` changelog bullet I had not read on the first pass. All
three say the same thing and none of it is what the code does:
`main.py:87` → `save_to_file(secrets, …)` → `format_for_output(secrets)`,
default `is_slim_mode=False`, unconditional. `--baseline` writes the **full**
form whatever it finds, so an already-slim baseline comes back with
`line_number` and `generated_at` restored — which is a better reason for the
redirect than the one written, and it is the same edit in three places.

## And the audit clause is missing from all three

`--baseline` merges as well as saves, and `SecretsCollection.merge` exists
for one purpose, in its own words: "we cannot merely discard the old results
in favor of the new, since there is valuable information that ought to be
preserved: verification of secrets, both automated and manual". It carries
`is_secret` and `is_verified` forward. A `scan > file` has nothing to merge,
so an audit's marks are gone at the next regeneration.

The changelog bullet is where this matters most, because it is the one that
says what the change costs and it stops one clause short: "what it costs is
one field of the diff … and that `detect-secrets audit` wants the full
form". The second cost is that the documented command discards what an audit
recorded, and the diff that shows it shows nothing. Nothing is lost today —
all 53 and all 466 findings are `is_verified: false` with no `is_secret`,
which is worth saying in the same clause, because it is why this is a
sentence and not a blocker.

## Re-verified on this head

53 → 53 and 466 → 466, equal as sets of filename, hash and type; exactly
`line_number` and `generated_at` gone; `plugins_used` identical; both files
still 1.5.0; and the filter swap equivalent, `is_baseline_file` plus the
vectors pattern giving way to one pattern that names both, which is right
because detect-secrets skips the file `--baseline` names on its own account.
CI on this head is 25 green with one job pending.

fametrano:
Both taken on `4ac4320`, and the minor answered.

## 1. The mechanism

You are right and my sentence was a guess dressed as a mechanism. Read
rather than inferred this time — `main.py`:

```python
        secrets.merge(args.baseline)
        baseline.save_to_file(secrets, args.baseline_filename)
    else:
        print(json.dumps(baseline.format_for_output(secrets, is_slim_mode=args.slim), indent=2))
```

`save_to_file` → `format_for_output(secrets)` with the default
`is_slim_mode=False`, unconditional. So `--baseline` writes the **full**
form whatever it read, and an already-slim file comes back with both fields
restored; `--slim` is honoured only on the `else` branch. That is what
`CONTRIBUTING.md`, `CLAUDE.md`, the changelog and the commit message say
now — and it is the stronger sentence, as you say, because it also answers
the question mine would have sent the next person to test.

## 2. The merge, which is the substantive one

Taken, and it is a second behavioural change that was going unsaid.
`SecretsCollection.merge`'s docstring is the reason in its own words —
"verification of secrets, both automated and manual" — and a
`scan > file` has nothing to merge, so `is_secret` and `is_verified` return
to their defaults. Checked, as you did: **0 of 53** and **0 of 466**
findings are verified and none carries `is_secret`, so nothing is lost
today.

The clause is beside the audit sentence in `CONTRIBUTING.md`, with the
part that makes it dangerous rather than merely true: slim output prints no
`is_secret` when it is unset, so the loss shows up as a diff of nothing.
The hook comment and the changelog carry it too.

## The minor

The figure is real but I did not say which spelling produced it. 88 and 90
are MD013's own report for the **factored** form,
`\.secrets\.(baseline|vectors\.baseline)`, which is what the first revision
carried; spelling both names out in full is longer still, which is your 99
and 97. The commit message now names the spelling beside the number, and
the collapsed one is 72 and 70.

640 passed, `uvx pre-commit run --all-files` exits 0. Rebased on
`752437d`, still one commit.

fametrano (commented):
**ACK `4ac43206`.** Both taken, in all four places, and the minor was mine
rather than yours.

## The 88 and 90 are right and my reconstruction was wrong

They reproduce exactly for the factored spelling you say the first revision
carried, `\.secrets\.(baseline|vectors\.baseline)`:

| spelling | `CLAUDE.md` | `CONTRIBUTING.md` |
| --- | --- | --- |
| factored | **88** | **90** |
| both names in full | 97 | 99 |
| collapsed, as shipped | 70 | 72 |

I measured the second row and reported the first as unreproducible. The
numbers were never wrong; the spelling behind them was not written down,
which is what naming it beside the number now fixes.

## Verified on this head

- **The mechanism** is right in all four places — `CONTRIBUTING.md`,
  `CLAUDE.md`, the hook comment and the bullet — and says what
  `main.py:87` does: `save_to_file` → `format_for_output(secrets)` with the
  default `is_slim_mode=False`, `--slim` honoured only on the branch that
  prints to stdout.
- **The merge clause** is beside the audit sentence with the half that makes
  it matter, which is that slim output prints no `is_secret` when unset, so
  the loss shows as a diff of nothing. And "nothing is lost today" carries
  the count, so a reader can tell a live hazard from a latent one.
- **The baselines are untouched by this revision**, so the equality I
  checked on `ac531a34` still stands: 53 → 53 and 466 → 466, equal as sets
  of filename, hash and type, `line_number` and `generated_at` the only keys
  gone, plugins and version identical, and the filter swap equivalent.

## The CI red is not yours either

`Build wheels on macos-26-intel` failed in **`Set up job`**: `429 Too Many
Requests` fetching `astral-sh/setup-uv@20cfd1bf` from codeload, after its
three internal retries, before a line of this repository ran. #237 lost a
wheel job the same way in the same minute, on a different runner, and
`githubstatus` reads **Partially Degraded Service** as of 14:20:58Z with the
REST API answering 503 on some endpoints. Re-run when it clears rather than
change anything.

Both of these want the gate green before the fast-forward, since "`uvx
pre-commit run --all-files` exits 0" in the body is a local run and
`Measure coverage, gated at 100%` is among the jobs still pending.

fametrano:
ACK read, thank you — and the minor was worth the round either way: naming
the spelling beside the number is what makes it reproducible, which is the
whole point of quoting one.

**Not landing it yet, for the reason your last paragraph gives.** The gate
has to be green on the head that lands, and `Build wheels on
macos-26-intel` is still the 429. `gh run rerun --failed` refuses while the
run is in progress, so it is waiting for the run to settle and will go in
as soon as it does; I will report the verdict rather than fast-forward on a
red check.

`githubstatus` agrees with your reading, and I have corrected my own
attribution in **#239** because of it: `api/v2/status.json` has been
`minor` / **Partially Degraded Service** since `2026-08-17T14:20:58Z`, and
three of the four 429s today fall inside that window. That issue's `429`
section had implied the concurrency caused it, which is not established —
what the concurrency does is make such a failure *possible* on a change the
matrix cannot check, and the runner-minutes there are the part that is
measured and unaffected.

Nothing about this branch changes for it: no revision, same `4ac43206`.

fametrano:
Landed at `f94540be`, which is not the sha I ACKed — `4ac43206` — so I checked what changed rather than assume a rebase. Nothing did: comparing only the added and removed lines of `git diff <parent> <head>` for the two, they are byte-identical, and the whole difference is hunk headers, context and blob hashes from the rebase onto `72be84f` (#236). So the ACK carries to what actually landed.

The rest of the landing: `f94540be` is `main`'s tip with parent `72be84f`, so unsquashed and fast-forwarded, signed `G Ferdinando M. Ametrano` with GitHub reporting `"reason": "valid"`. #230 closed.

Worth recording for the next reader of this pull request: the two red wheel jobs it and #237 collected were a GitHub incident and not this diff — a codeload `429` on `macos-26-intel` here, another on `macos-latest` there, and a windows-11-arm timeout after that, with the status page going from Partially Degraded Service to Partial System Outage across the same half hour.

Issue #240, opened 2026-08-17
The no-hook argument names the false positive and not the false negative


fametrano:
Raised in the [review of #237](https://github.com/btclib-org/btclib-secp256k1/pull/237#pullrequestreview-4957000000)
as optional and explicitly not worth a revision there, so it is filed rather
than folded in.

#233's bullet argues that a hook for mid-sentence short lines does not belong
in this gate, and the argument it makes is about **false positives**: the
thresholds are a heuristic, and a false positive has no `# noqa` to be marked
with in prose. That is true and it is the weaker half.

The stronger half is a false *negative*, and #237's own history is the worked
example. Correcting one sentence in `CLAUDE.md` left this:

```
126[72]  lives rather than a refusal to share objects, which is why `--reference`
127[57]  is allowed to share them and why it changes nothing about
128[35]  `core.worktree`. For a recipe whose
129[74]  last line removes the worktree anyway that is a trade worth declining, and
```

Line 128 is the defect — 35 columns, mid-sentence — and the detector that
had just counted thirteen of exactly that shape walked straight past it,
because **both** of its thresholds miss it at once:

| the filter asks | line 128 | line 127 |
| --- | --- | --- |
| this line ≤ 46 columns | 35 ✓ | 57 ✗ |
| the previous line ≥ 65 columns | 57 ✗ | 72 ✓ |

A *pair* of short lines is what hides a widow: the first break is above the
ceiling and the second has no long line above it. Neither threshold can be
loosened for free — at 55 columns the same scan reports 46 lines, most of
them a sentence deliberately begun on its own line, and dropping the
previous-line condition reports every short line before a long code span.

So what a hook would have missed here is the instance a human found by
reading, which is a better reason to decline it than the noise it would have
added. #233's bullet says only the second, and the clause belongs beside it
whenever `CHANGELOG.md`'s v0.8.0.3 section is next open — it is prose in a
released-notes file, so it wants a change that is already touching it rather
than one of its own.

Two smaller facts from the same review, worth keeping with it:

- the two short lines that remain in that paragraph after the rewrap are
  token-forced rather than left: 47 columns before
  `<primary>/.git/modules/secp256k1` (32 characters, so 80 joined) and 60
  before `…/.git/worktrees/<wt>/modules/secp256k1` (40, so 101). The same
  shape as the three link lines #237 leaves alone;
- "no consecutive short pair anywhere in it" is the property that made the
[...]

PR #241, opened 2026-08-17
Take the two file kinds nothing built reads off the matrix

### LINKED ISSUE #239: The prose allowlist does not cover the file #189 was measuring
[#189](https://github.com/btclib-org/btclib-secp256k1/pull/189) added the
`changes` job so a pull request that changes only prose stops running the
matrix, and its own motivating example is still not covered:

> Measured on the last full run before this, 48 jobs and 81.8
> runner-minutes, of which the change that added `windows.yml` to the table
> of sentinels — **four markdown files and one workflow comment** — asked
> for all of them.

The allowlist that shipped is prose alone:

```sh
prose='^(CHANGELOG|CLAUDE|CONTRIBUTING|HISTORY|RELEASING|REPOSITORY|SECURITY)\.md$|^\.github/[^/]*\.md$'
```

so a change that touches one file the matrix cannot read alongside prose
runs everything, and "one workflow comment" is exactly that file.

## Measured twice today

| pull request | files | jobs | runner-minutes |
| --- | --- | --- | --- |
| [#227](https://github.com/btclib-org/btclib-secp256k1/pull/227) | `.github/workflows/mutation.yml`, `CLAUDE.md`, `CHANGELOG.md` | 23 | 79.2 |
| [#236](https://github.com/btclib-org/btclib-secp256k1/pull/236) | `.github/mutation/bindings.toml`, `CLAUDE.md`, `CHANGELOG.md` | 23 | 45.7 |

Neither file is read by anything the matrix builds. `bindings.toml` is the
mutation sentinel's configuration — `cosmic-ray` takes its path as an
argument and no other workflow names it. `mutation.yml` is a workflow the
matrix does not contain.

## And it cost a red check, once

Three pull requests opened within minutes ran three matrices at once, and
`Build wheels on ubuntu-latest` on #238 failed in **`Set up job`**:

```
Failed to download archive 'https://codeload.github.com/astral-sh/setup-uv/tar.gz/20cfd1bf…' after 3 attempts.
Response status code does not indicate success: 429 (Too Many Requests).
```

A red check no pull request introduced, on a step that runs no code of
ours. That is a second-order cost of running the matrix on changes it has
nothing to check, and it is the shape the repository already dislikes
enough to have split the macOS and Windows sentinels out of the gate.

## What could be done

1. **Add what nothing built reads to the allowlist.** `^\.github/mutation/`
   is unambiguous: no workflow but `mutation.yml` names that directory, and
   `mutation.yml` is not in the matrix. `^\.secrets(\.[a-z]+)*\.baseline$`
   is the same kind of file — detect-secrets' own output, read by a hook
   and by nothing that builds.
2. **The workflow-comment case is harder, and should be said rather than
   guessed.** A change to `test.yml` must run the matrix it defines, and to
   `release.yml` likewise; a change to `mutation.yml`, `links.yml` or
   `macos.yml` cannot affect a wheel. An allowlist naming the sentinels by
   file would be right today and would go stale the first time a sentinel
   is folded back into the gate — which is a thing that has happened twice
   here in the other direction. Either name them and accept that, or leave
   every workflow change running everything and say why in the comment
   above the pattern.
3. **Nothing.** 45 to 79 runner-minutes on a prose-and-configuration change
   is real but not large, and the ceiling that matters is concurrent jobs
   rather than minutes. This is a legitimate answer; what it should not be
   is the answer by default, which is what it is now.

(1) is worth doing on its own. (2) is the decision, and the comment above
the pattern is where it belongs whichever way it goes.

fametrano:
#239's option (1), with its option (2) answered rather than left open. One
line of pattern, one comment, one changelog bullet; no job changes what it
does.

## What joins the allowlist

```diff
-prose='^(CHANGELOG|CLAUDE|…|SECURITY)\.md$|^\.github/[^/]*\.md$'
+prose='^(CHANGELOG|CLAUDE|…|SECURITY)\.md$|^\.github/[^/]*\.md$|^\.github/mutation/|^\.secrets\..*baseline$'
```

`.github/mutation/` is the session's configuration, taken as a path
argument by cosmic-ray from `mutation.yml` — a workflow the matrix does not
contain. The baselines are detect-secrets' own output, read by a hook.
Nothing built reads either, which is the list's own criterion.

## Measured against real file lists

| files changed | before | after |
| --- | --- | --- |
| #236: `bindings.toml` + 2 md | runs (23 jobs, 45.7 min) | **skips** |
| a baseline alone | runs | **skips** |
| `.github/mutation/` alone | runs | **skips** |
| #237: prose + `silentpayments.py` | runs | runs |
| #227: `mutation.yml` + 2 md | runs | runs |
| a bare `README.md` | runs | runs |
| a file under `docs/` | runs | runs |
| an empty list | runs | runs |

The last four are the cases the comment above the pattern already warns
this must not get wrong, and they are unchanged.

## The measurement needed the right `grep`, which inverted the first answer

Worth reading before trusting any local reproduction of this step: this
machine's `grep` is **ugrep 7.5.0**, and it takes the exit status of `-v`
from whether the *pattern* matched rather than from whether any line was
selected.

```
$ printf '%s\n' CHANGELOG.md README.md > n.txt
$ grep -cvE '^CHANGELOG\.md$' n.txt   # ugrep: one non-matching line
1
$ grep -qvE '^CHANGELOG\.md$' n.txt; echo $?   # ugrep
1
$ /usr/bin/grep -qvE '^CHANGELOG\.md$' n.txt; echo $?
0
```

So the first pass of the table above reported "skips" for a pull request
containing `silentpayments.py`. The step is unaffected — it runs on
ubuntu-latest — but a session reproducing it by hand on macOS gets the
opposite verdict silently. Filed as a separate issue rather than fixed
here.

## What is deliberately not on the list

A workflow. Which workflows the matrix reads is a judgement — this file,
what it calls, `release.yml` — and naming the rest would be right today and
stale the first time a sentinel is folded into the gate or one of these is
split out, both of which have happened here. That staleness skips a matrix
that *should* have run, which is the expensive direction. So every workflow
change keeps running everything, and the comment says why rather than
leaving the next reader to infer it — including that #227's own shape,
`mutation.yml` plus prose, still costs the full matrix.

## Verified

640 passed, `uvx pre-commit run --all-files` exits 0 — `actionlint` and
`zizmor` included, this being a workflow change. Branches from `main` at
`3b915d2`.

Closes #239

## Summary by Sourcery

Reduce unnecessary matrix runs for file changes that are not consumed by the matrix while preserving coverage for workflow and source changes.

Enhancements:
- Refine the workflow changed-file allowlist so mutation configuration and detect-secrets baseline files no longer trigger the full test matrix when only non-consumed files change.
- Document the rationale for excluding these files while deliberately continuing to run the matrix for workflow changes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -144,8 +144,39 @@ jobs:
           # the root and of .github, and neither README.md nor a file
           # under docs/, which the sdist carries. An empty list is not a
           # documentation change, it is a question this could not answer,
-          # so the count is checked before the pattern
-          prose='^(CHANGELOG|CLAUDE|CONTRIBUTING|HISTORY|RELEASING|REPOSITORY|SECURITY)\.md$|^\.github/[^/]*\.md$'
+          # so the count is checked before the pattern.
+          #
+          # The mutation configuration and the detect-secrets baselines are
+          # on it for the same reason as the prose and were measured off it:
+          # a change to bindings.toml alone cost 23 jobs and 45.7
+          # runner-minutes (#239), and cosmic-ray takes that path as an
+          # argument from mutation.yml, which is a workflow the matrix does
+          # not contain. A baseline is detect-secrets' own output, read by a
+          # hook -- and by lint.yml, which has no changed-file gate at all,
+          # so a malformed one is still caught by the job that catches it
+          # today.
+          #
+          # Anchored on the *file* and not on the directory, which is the
+          # same staleness this comment declines below in the direction that
+          # looks cheap: `^\.github/mutation/` would exempt in advance any
+          # file a later change drops there, and two tests already load a
+          # script by path one directory over (`.github/scripts/`), so such
+          # a script would skip the matrix and take its own test with it.
+          # `.toml` rather than `bindings\.toml` so a second configuration
+          # needs no edit here; a script belongs in .github/scripts/, which
+          # is where the three of them are.
+          #
+          # A workflow is *not* on it, deliberately, and that is the answer
+          # to the other half of #239 rather than an omission. Which
+          # workflows the matrix reads is a judgement -- this file and what
+          # it calls, plus release.yml -- and naming the rest would be right
+          # today and stale the first time a sentinel is folded into the
+          # gate or one of these is split out, both of which have happened
+          # here. A stale list of that kind skips a matrix that should have
+          # run, which is the expensive direction; leaving every workflow
+          # change to run everything is the cheap one, at the cost this
+          # comment now records
+          prose='^(CHANGELOG|CLAUDE|CONTRIBUTING|HISTORY|RELEASING|REPOSITORY|SECURITY)\.md$|^\.github/[^/]*\.md$|^\.github/mutation/[^/]*\.toml$|^\.secrets\..*baseline$'
           if [ ! -s files.txt ] || grep -qvE "$prose" files.txt; then
             echo "code=true" >> "$GITHUB_OUTPUT"
             echo "something the matrix reads changed: everything runs"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2076,6 +2076,34 @@ release-notes length in the first place, and are still in
   and the paragraph says what the ratchet does and does not catch
   instead: a line that stops running is caught, a line still run by
   another test and asserted by none is what neither question sees.
+- **The prose allowlist covers the two file kinds nothing built reads**
+  (#239). `changes` answers whether the matrix has anything to check,
+  and its pattern was the root and `.github` prose alone — so a pull
+  request touching one file the matrix cannot read alongside prose ran
+  everything. The mutation configuration is on it now, cosmic-ray taking
+  that path as an argument from a workflow the matrix does not contain,
+  and so are the detect-secrets baselines, which are that tool's own
+  output read by a hook and by a workflow that has no changed-file gate
+  at all. It is anchored on the file --
+  `^\.github/mutation/[^/]*\.toml$` -- and not on the directory, that
+  being the same staleness this declines for workflows in the direction
+  that looks cheap: a prefix would exempt in advance whatever a later
+  change drops there, and two tests already load a script by path one
+  directory over, so such a script would skip the matrix and take its
+  own test with it. `.toml` rather than the one filename so a second
+  configuration needs no edit; a script belongs in `.github/scripts/`,
+  where all three are. Measured before and after against the file lists
+  of four real pull requests, with `/usr/bin/grep` rather than this
+  machine's `grep`: #236's `bindings.toml` and two markdown files went
+  from 23 jobs and 45.7 runner-minutes to a skipped matrix, a baseline
+  alone likewise, while #237's `silentpayments.py`, a bare `README.md`,
+  a file under `docs/` and an empty list all still run everything. A
+  workflow is deliberately not on the list, which is the other half of
+  #239 answered rather than omitted: which workflows the matrix reads is
+  a judgement — this file, what it calls, and `release.yml` — and a list
+  of the rest would be right today and stale the first time a sentinel
+  is folded into the gate, which skips a matrix that should have run.
+  The comment above the pattern carries that reasoning and the figure.
 
 ## v0.8.0.2
 


.github/workflows/test.yml
```diff
@@ -144,8 +144,26 @@ jobs:
           # the root and of .github, and neither README.md nor a file
           # under docs/, which the sdist carries. An empty list is not a
           # documentation change, it is a question this could not answer,
-          # so the count is checked before the pattern
-          prose='^(CHANGELOG|CLAUDE|CONTRIBUTING|HISTORY|RELEASING|REPOSITORY|SECURITY)\.md$|^\.github/[^/]*\.md$'
+          # so the count is checked before the pattern.
+          #
```
fametrano: `^\.github/mutation/` is the one entry that names a **directory**; every other names a file or a filename shape. So it exempts in advance any file a later change puts there -- and this hunk's own argument against listing workflows is that a list goes stale and "skips a matrix that should have run, which is the expensive direction". A directory prefix is that same staleness in the direction that looks cheap: a script dropped under `.github/mutation/` and loaded by a test -- which is what `test_mutation_counts.py` and `test_submodule_pin.py` already do one directory over -- would skip the matrix and its own test with it.

Either anchor it (`^\.github/mutation/bindings\.toml$`, or `[^/]*\.toml$`) or say here that the directory is configuration by convention and a script belongs in `.github/scripts/`.

fametrano (commented):
**Changes requested** on `5e5567ae` — one clause, and the safety argument
itself holds. This is a mechanism that decides whether a change is tested at
all, so I went after the criterion rather than the pattern.

## The criterion holds, checked rather than reasoned about

"Nothing built reads either" is the claim, and nothing does:

- **`tests/test_mutation_counts.py` reads `.github/scripts/mutation_counts.py`,
  not `.github/mutation/`.** Its `_SCRIPT` is
  `parents[1] / ".github" / "scripts" / "mutation_counts.py"`; `bindings.toml`
  appears in its docstring as context and in `test_extension.py:201` in a
  comment. So no test in the matrix loads the allowlisted path. Same for
  `check_submodule_pin.py`, which is the other script the suite loads by
  path — also under `.github/scripts/`.
- **Nothing in `tests/` reads a baseline**, and the thing that does read one
  is not in the matrix: `lint.yml` carries no changed-file gate at all, so
  the detect-secrets hooks run on every pull request whatever the file list
  is, and `pre-commit.ci` runs beside them. A malformed baseline still gets
  caught, by the job that is supposed to catch it.
- **The pattern exists in exactly one place.** `grep -rn 'prose=' .github/workflows/`
  is one hit, so there is no second copy to drift out of step — which is the
  failure mode a shared allowlist usually has.
- `.github/workflows/mutation.yml` does not match `^\.github/mutation/`, so
  #227's shape still runs everything, as the table says.

## The ugrep finding reproduces, and its blast radius is one line

Worth confirming because it inverts a verdict silently. On this machine,
ugrep 7.5.0:

```
grep -qvE '^CHANGELOG\.md$' n.txt      -> exit 1
/usr/bin/grep -qvE '^CHANGELOG\.md$' n.txt -> exit 0
grep -cvE '^CHANGELOG\.md$' n.txt      -> 1
```

So `-q` with `-v` reports whether the *pattern* matched while `-c` correctly
counts selected lines. Filing it separately (#242) rather than working
around it is right, and one thing worth adding there: I grepped the tree for
every place it could reach, and `test.yml:149` is the **only** `grep -qv` in
the repository. So #242 is bounded to that one line and to sessions
reproducing it by hand, which is what it says — no script a contributor runs
is silently inverted.

## The clause: a directory prefix grants what does not exist yet

`^\.github/mutation/` is the one entry on the list that names a **directory**.
Every other one names a file, or a shape of filename. So it exempts, in
advance, any file a later change puts under that path — and the argument this
pull request makes against listing workflows is precisely that a list goes
stale:

> naming the rest would be right today and stale the first time a sentinel
> is folded into the gate or one of these is split out … A stale list of
> that kind skips a matrix that should have run, which is the expensive
> direction

A directory prefix is the same staleness in the direction that *looks*
cheap. A script dropped under `.github/mutation/` that the suite loads by
path — which is exactly what two tests already do, one directory over —
would skip the matrix and take its own test with it, silently. Today the
convention is clean: `.github/mutation/` holds `bindings.toml` and nothing
else, and the three scripts live in `.github/scripts/`.

Either anchor it — `^\.github/mutation/bindings\.toml$`, or
`^\.github/mutation/[^/]*\.toml$` if a second config is foreseeable — or say
in the comment that the directory is configuration by convention and that a
script belongs in `.github/scripts/`, so the next reader knows the exemption
is about *what kind of file* rather than about the path. The comment argues
its way carefully everywhere else in this hunk; this is the one place it
leaves a judgement implicit.

Non-blocking beside it: the changelog bullet says the figures were measured
"with `/usr/bin/grep` rather than this machine's `grep`", which is the right
disclosure and reads oddly without #242 to point at. A parenthetical
reference would carry it.

CI is 25 green on this head.

fametrano:
Taken on `f2fce31`, and you are right that it is the same staleness wearing
the cheap face. The entry is anchored:

```diff
-|^\.github/mutation/|
+|^\.github/mutation/[^/]*\.toml$|
```

`.toml` rather than `bindings\.toml` so a second configuration needs no edit
here, and the comment now says the rest: a script belongs in
`.github/scripts/`, which is where all three are.

**The case that decides it, measured with the anchor in place:**

| files changed | before the anchor | after |
| --- | --- | --- |
| `.github/mutation/bindings.toml` | skips | skips |
| `.github/mutation/numeric.toml` | skips | skips |
| `.github/mutation/report.py` | **skips** | **runs** |
| `.github/mutation/lib/helper.py` | **skips** | **runs** |

The third row is your example, and the fourth is the subdirectory it implies.

## Your four checks, re-derived here

- `.github/mutation/` holds `bindings.toml` and nothing else.
- `grep -rn "prose=" .github/workflows/` → **1**. No second copy to drift.
- The two path-loading tests are `_SCRIPT` in `tests/test_mutation_counts.py`
  and the same shape in `tests/test_submodule_pin.py`, both under
  `.github/scripts/`. Nothing in `tests/` reads `.github/mutation/` or a
  baseline.
- `lint.yml` carrying no changed-file gate is the part I had not said and the
  comment now does: a malformed baseline is still caught by the job that
  catches it today, whatever the file list.

## On #242, your bound

Confirmed and added there: `grep -qv` occurs **once** in the whole tree, now
at `test.yml:167` after this comment grew. So the ugrep inversion is bounded
to that one line and to a session reproducing it by hand — no script a
contributor runs is silently inverted.

640 passed, `uvx pre-commit run --all-files` exits 0, `actionlint` and
`zizmor` included. Rebased on `main` at `3b915d2`, still one commit.

fametrano (commented):
**ACK `f2fce31c`.** Anchored on the file kind, and the comment now carries
both the anchor's reasoning and the `lint.yml` half of the safety argument,
which was the part the criterion rested on and did not say.

`[^/]*\.toml$` rather than `bindings\.toml$` is the better of the two I
offered — a second configuration needs no edit, and a script still cannot
sneak in.

I ran the whole table through the step's own logic rather than reading the
pattern, with `/usr/bin/grep` for #242's reason:

```
SKIPS  .github/mutation/bindings.toml CHANGELOG.md CLAUDE.md
SKIPS  .secrets.baseline
SKIPS  .secrets.vectors.baseline
SKIPS  .github/mutation/bindings.toml
SKIPS  .github/mutation/second.toml
SKIPS  CHANGELOG.md .github/CONTRIBUTING.md
RUNS   .github/mutation/helper.py          <- the fix
RUNS   .github/mutation/nested/x.toml      <- what [^/]* buys
RUNS   .github/workflows/mutation.yml CHANGELOG.md
RUNS   btclib_secp256k1/silentpayments.py CHANGELOG.md
RUNS   README.md
RUNS   docs/index.md
RUNS   .secrets.baseline btclib_secp256k1/dsa.py
RUNS   (empty list)
```

Fourteen cases, every one the way it should be, including the two the anchor
was for: a script under `.github/mutation/` runs the matrix, and so does a
`.toml` one directory deeper, which `[^/]*` refuses on purpose. The four the
old comment already warned about are unchanged, and a mixed list still runs —
one file the matrix reads is enough, which is the property that makes an
allowlist safe.

The rest I verified on the first pass and it is untouched: nothing in `tests/`
loads `.github/mutation/` or a baseline, the pattern lives in exactly one
place so there is no copy to drift, and `.github/workflows/mutation.yml` does
not match.

CI is green on this head.

fametrano:
Correcting one line of that: "CI is green on this head" was written before I read it and it was `success=4 pending=15` at the time — the run had just started. The ACK is on the change, which is what I checked; the gate is its own statement and had not made it yet. Nothing else in the review depends on it.

fametrano:
Rebased onto `ed88674` — #243 landed — and pushed as `d55f3c9`. No content
change from `f2fce31`: the rebase moved the base, and `CHANGELOG.md`'s
`merge=union` did its job. Checked rather than assumed, that file being the
one whose union merge duplicates in silence:

```
bullets 267 | duplicated 0 | conflict markers 0
#240's clause present once, #239's bullet present once
```

Gate exits 0 and 640 pass on the rebased tree.

This is the only pull request still open. #243 is on `main` at `ed88674`,
`verified: true`, with #240 closed as completed — and its own matrix skipped
on one `CHANGELOG.md`, which is this branch's mechanism working one file kind
over, twice in a row now.

fametrano (commented):
**ACK `d55f3c92`** — the ACK on `f2fce31c` carries, and the union-merge check
you added is the right one to have run.

Both verified independently:

- **Pure rebase.** Comparing only the added and removed lines of
  `git diff <parent> <head>` for the two heads, the difference is **empty**,
  and the new parent is `ed88674`, which is `main`. So nothing about the
  change moved.
- **`merge=union` did not duplicate.** `.gitattributes` does set it for
  `CHANGELOG.md` and `HISTORY.md`, with the comment naming exactly this
  hazard, and on this head: **267 bullets, 267 distinct, 0 duplicated, 0
  conflict markers**, with `#239`'s bullet and `#240`'s clause each appearing
  once. Your figure and mine agree, and this is a check worth keeping in the
  habit — a union merge that duplicates leaves a file that lints, tests and
  reads almost right.

CI is 9 green and 10 pending as I write, so this head has not finished its
run. `f2fce31c` finished 25 of 25, and the diff between them is nothing, so
what is pending is a re-run of an answered question rather than an open one —
worth saying only because I got that distinction the wrong way round on this
same pull request an hour ago and would rather be exact about it than brief.

This is the last of the eight. Nothing else from me.

PR #243, opened 2026-08-17
Say which half of the no-hook argument is the strong one

### LINKED ISSUE #233: Thirteen more lines stopped short where a qualifier went in

### LINKED ISSUE #240: The no-hook argument names the false positive and not the false negative

fametrano:
#240, in the bullet it belongs to. Prose only, one file.

## What the bullet said, and what it left out

#233's bullet declines a hook for mid-sentence short lines and gives one
reason: the thresholds are a heuristic, and a false positive has no
`# noqa` to be marked with in prose. That is the **weaker** half.

The stronger half is a false negative, and the change that closed #233
produced one of its own. Correcting a sentence in `CLAUDE.md` left this:

```
126[72]  lives rather than a refusal to share objects, which is why `--reference`
127[57]  is allowed to share them and why it changes nothing about
128[35]  `core.worktree`. For a recipe whose
129[74]  last line removes the worktree anyway that is a trade worth declining, and
```

Line 128 is the defect, and the detector that had just counted thirteen of
exactly that shape walked past it, because **both** thresholds miss it at
once:

| the filter asks | line 128 | line 127 |
| --- | --- | --- |
| this line ≤ 46 columns | 35 ✓ | 57 ✗ |
| the previous line ≥ 65 columns | 57 ✗ | 72 ✓ |

A *pair* of short lines hides a widow: the first break is too long to be
flagged, and the second has no long line above it. So what a hook would
have missed here is the instance a reader found — a better reason to
decline one than the noise it would have added.

## The two smaller facts, because together they are what "complete" means

The two short lines left in that paragraph after the rewrap are
token-forced rather than left: 47 columns before a 32-character path and 60
before a 40-character one, the same shape as the three links #233 leaves
alone. And *no consecutive short pair anywhere in the paragraph* is the
property that says the rewrap is done, which is cheaper to check by eye
than either threshold.

## Verified

The bullet is rewrapped whole rather than patched, so the extension leaves
no widow of its own — #233's detector reports none in `CHANGELOG.md` on
this head, and the one hit a looser scan shows is line 9, the
`markdownlint-configure-file` comment, which is pre-existing and not prose.
640 passed, `uvx pre-commit run --all-files` exits 0.

Rebased on `main` at `3b915d2`. It touches only `CHANGELOG.md`, so
`changes` skips the matrix for it — which is what #241 is about, one file
kind over.

Closes #240

## Summary by Sourcery

Clarify both the false-positive and false-negative reasons for declining a prose widow-detection hook.

Enhancements:
- Clarify the rationale for rejecting a prose widow-detection hook by documenting false negatives, including hidden widows formed by consecutive short lines, alongside false positives.
- Clarify that the remaining short lines are token-forced and that completion is assessed by ensuring no consecutive short-line pairs remain.

Documentation:
- Expand the CHANGELOG discussion of prose widow detection to explain its limitations and why a hook is not appropriate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1764,10 +1764,30 @@ release-notes length in the first place, and are still in
   not defects at all and stay: a line before a markdown link of 63, 84
   and 125 columns is short because nothing can follow it, and the filter
   measured the link's first word rather than the link. No hook does
-  this, which #233 asked about and answers no: the thresholds are a
+  this, which #233 asked about and answers no. The thresholds are a
   heuristic, a false positive has no `# noqa` to be marked with in
   prose, and a gate that is exact everywhere else is the wrong place for
-  one that is nearly right.
+  one that is nearly right — but the stronger half of that is the false
+  *negative*, and this change produced one of its own (#240). Correcting
+  a sentence in `CLAUDE.md` left a 35-column line mid-sentence under a
+  57-column one, and the detector that had just counted ten of exactly
+  that shape -- thirteen reported, ten of them defects -- walked past
+  it: 35 is under the 46-column ceiling but 57 is under the 65 the line
+  above has to reach, and the line at 57 is itself the same shape and
+  over the ceiling. A *pair* of short lines hides a widow, because the
+  first break is too long to be flagged and the second has no long line
+  above it. So what a hook would have missed here is the instance a
+  reader found, which is a better reason to decline it than the noise it
+  would have added — and the two short lines left in that paragraph
+  after the rewrap are token-forced, 47 columns before a 32-character
+  path and 58 before a 40-character one, the same shape as the three
+  links. Those are columns and not bytes, which is worth saying where
+  the figures are: the second line carries an em dash, three bytes and
+  one column, so `awk` in a byte locale and `wc -c` answer 60 where the
+  detector's own `len` answers 58, and every figure in this bullet is
+  the second kind. No consecutive short pair anywhere in it is what says
+  the rewrap is complete, and it is cheaper to check by eye than either
+  threshold.
 - **The worktree recipe costs a submodule clone, and the sentence under
   it counts it now** (#234). It said the venv and the C build "are the
   whole of the cost" and after the submodule line they are not: a linked


CHANGELOG.md
```diff
@@ -1764,10 +1764,25 @@ release-notes length in the first place, and are still in
...
+  57-column one, and the detector that had just counted thirteen of
+  exactly that shape walked past it: 35 is under the 46-column ceiling
+  but 57 is under the 65 the line above has to reach, and the line at 57
+  is itself the same shape and over the ceiling. A *pair* of short lines
+  hides a widow, because the first break is too long to be flagged and
+  the second has no long line above it. So what a hook would have missed
+  here is the instance a reader found, which is a better reason to
+  decline it than the noise it would have added — and the two short
+  lines left in that paragraph after the rewrap are token-forced, 47
+  columns before a 32-character path and 60 before a 40-character one,
```
fametrano: `60` is bytes; in columns it is **58**. Line 112 of `CLAUDE.md` carries an em dash -- three bytes, one column -- so `awk 'length($0)'` and `wc -c` say 60 where Python's `len` says 58, and Python's `len` is what #233's detector uses and what every other figure in this bullet is.

The 47 and both token lengths (32 and 40) are right. And the 60 came from my own `awk` run in the #237 review, so the digit is mine; the clause worth adding is that these are columns rather than bytes, since this file is dense with em dashes.

CHANGELOG.md
```diff
@@ -1764,10 +1764,25 @@ release-notes length in the first place, and are still in
...
   measured the link's first word rather than the link. No hook does
-  this, which #233 asked about and answers no: the thresholds are a
+  this, which #233 asked about and answers no. The thresholds are a
   heuristic, a false positive has no `# noqa` to be marked with in
   prose, and a gate that is exact everywhere else is the wrong place for
-  one that is nearly right.
+  one that is nearly right — but the stronger half of that is the false
+  *negative*, and this change produced one of its own (#240). Correcting
+  a sentence in `CLAUDE.md` left a 35-column line mid-sentence under a
+  57-column one, and the detector that had just counted thirteen of
```
fametrano: "thirteen of exactly that shape" against this same bullet ten lines up: "Three of the thirteen the issue counted are not defects at all". Thirteen were reported, **ten** were of that shape -- which is the number in the bullet's own opening. "Reported thirteen", or "counted ten of exactly that shape".

fametrano (commented):
**Changes requested** on `f9c0d3e`-or-whatever this head is — two figures,
both small, and the argument is right and is mine to have asked for so I have
checked it harder than usual.

## Verified

- **The instance is exactly as described.** `CLAUDE.md` at the head that
  landed had 128 at 35 columns mid-sentence under 127 at 57, and the
  threshold table is right: 35 is under the 46 ceiling, 57 is under the 65
  floor, and 57 is itself over the ceiling. Both halves of the filter miss it
  at once, which is the finding.
- **The bullet is rewrapped whole**, 66 to 73 columns throughout, so the
  extension leaves no widow of its own — `#233`'s detector reports three on
  this head and they are the three link lines, none in `CHANGELOG.md`.
- **Line 9 is what a looser scan finds** and it is inside the
  `markdownlint-configure-file` JSON block, pre-existing and not prose. ✓

## 1. "Thirteen of exactly that shape" contradicts the same bullet

Three lines above the new text, this bullet says:

> Three of the thirteen the issue counted are not defects at all and stay

and the new text says the detector "had just counted thirteen of exactly that
shape". Thirteen were *reported*; **ten** were of that shape, which is the
number in this bullet's own opening ("Ten lines that stopped short
mid-sentence"). "Reported thirteen" or "counted ten of exactly that shape"
— either, but not thirteen of that shape.

## 2. "60 before a 40-character one" is 58 columns, and the reason is worth a clause

The two token-forced lines, measured on `main` at `3b915d2`:

| line | columns | next token | join needs |
| --- | --- | --- | --- |
| `CLAUDE.md:116` | 47 | `<primary>/.git/modules/secp256k1`, 32 | 80 |
| `CLAUDE.md:112` | **58** | `` …/.git/worktrees/<wt>/modules/secp256k1` ``, 40 | 99 |

The 47 and both token lengths check out. The 60 does not, and the reason is
the interesting part: line 112 carries an **em dash**, which is three bytes
and one column, so `awk 'length($0)'` in a byte locale and `wc -c` say 60
where Python's `len` — which is what `#233`'s detector uses, and therefore
what every other figure in this bullet is — says 58.

I made the same mistake measuring it for you: my review of `#237` quoted this
line at 60 from an `awk` run. So the number came from me and the correction
is mine too. Worth more than a digit, though, because this file is dense with
em dashes and line lengths get measured here often: a clause saying the
figures are columns rather than bytes would keep the next measurement
honest, and would explain why two tools disagree by exactly two on this line.

Fix those two and this is an ACK. The matrix skipped on this head, correctly
— `changes` reading one `CHANGELOG.md` and nothing else, which is #241's
mechanism working one file kind over.

fametrano:
Both taken on `ed88674`, and the second one is better than a digit.

## 1. Ten, not thirteen

Right, and it contradicted this bullet's own opening. It now reads "the
detector that had just counted **ten** of exactly that shape — thirteen
reported, ten of them defects — walked past it", so the two numbers appear
together with what separates them, which is what the bullet three lines up
says about the other three.

## 2. 58, and the reason is in the clause now

Measured both ways here, on `main` at `3b915d2`:

| | columns (`len`) | bytes | em dashes | next token | join needs |
| --- | --- | --- | --- | --- | --- |
| `CLAUDE.md:116` | 47 | 47 | 0 | 32 | 80 |
| `CLAUDE.md:112` | **58** | 60 | 1 | 40 | 99 |

`awk '{print length($0)}'` says 60 on that line and `wc -c` 61 with the
newline — a byte locale counting the em dash three times.

So the bullet now carries the distinction rather than just the corrected
number: *columns and not bytes, and every figure in that bullet is the
second kind*. Your reason for wanting it is the right one — this file is
dense with em dashes and its line lengths get measured often, so two tools
disagreeing by exactly two on one line is a trap worth naming once.

The commit message carries both corrections, so `git log` and the file agree.

The matrix skipped again on this head, one `CHANGELOG.md` and nothing else.
640 passed, `uvx pre-commit run --all-files` exits 0, and #233's detector
reports no widow in `CHANGELOG.md` — the extension is rewrapped whole, as
before.

fametrano (commented):
**ACK `ed886742`.** Both figures corrected, and the second one is better than
what I asked for: naming both tools and both answers — `awk` in a byte locale
and `wc -c` say 60 where the detector's own `len` says 58 — makes the clause
re-derivable instead of just right, and "every figure in this bullet is the
second kind" settles it for the whole bullet rather than for the one line.

"Counted ten of exactly that shape — thirteen reported, ten of them defects"
is the precise version and it now agrees with the bullet's own opening.

Verified on this head:

- **The bullet is rewrapped whole**, 66 to 73 columns with the 12-column
  `threshold.` ending it, and no consecutive short pair anywhere in it — the
  property this bullet itself now proposes as the check, applied to itself.
- **`#233`'s detector reports three** and they are the three link lines, none
  in `CHANGELOG.md`.
- The word-level diff against the previous head is confined to the edit.

One thing I went to check and did not find: I was going to raise the `--` in
"that shape -- thirteen reported" as inconsistent with the em dashes around
it, and it is not — this file carries 318 em dashes and 39 lines using `--`,
so both are established here and the new text sits inside the existing usage.

The matrix skipped on this head, correctly: `changes` reading one
`CHANGELOG.md` and nothing else, 4 green and 9 skipped, which is #241's
mechanism doing what its table says.

PR #244, opened 2026-08-17
Say which grep reproduces the changes job, and why it matters

### LINKED ISSUE #242: Reproducing the changes job locally answers the opposite, where grep is ugrep
Reproducing the `changes` job's decision by hand answers the opposite of what
the runner answers, on a machine where `grep` is
[ugrep](https://ugrep.com). It is silent, and the wrong direction: a pull
request full of code reads as "prose only".

## What the step does

`.github/workflows/test.yml`, the `changes` job:

```sh
prose='^(CHANGELOG|CLAUDE|…|SECURITY)\.md$|^\.github/[^/]*\.md$|…'
if [ ! -s files.txt ] || grep -qvE "$prose" files.txt; then
  echo "code=true" >> "$GITHUB_OUTPUT"
```

The verdict is `grep -qvE`'s exit status: zero where some line does *not*
match, which is a file the matrix has to check.

## What ugrep answers

```console
$ grep --version | head -1
ugrep 7.5.0 aarch64-apple-macosx …

$ printf '%s\n' CHANGELOG.md README.md > n.txt

$ grep -cvE '^CHANGELOG\.md$' n.txt      # non-matching lines: right
1
$ grep -vE '^CHANGELOG\.md$' n.txt       # and it prints the line
README.md
$ grep -qvE '^CHANGELOG\.md$' n.txt ; echo $?
1

$ /usr/bin/grep -qvE '^CHANGELOG\.md$' n.txt ; echo $?
0
```

It counts and prints the selected lines correctly and then exits **1**,
taking the status from whether the pattern matched rather than from whether
`-v` selected anything. `/usr/bin/grep` — BSD grep, and GNU grep on the
runner — exits 0.

Where that bites: the first pass of #241's own before-and-after table
reported a *skipped* matrix for the file list of #237, which contains
`btclib_secp256k1/silentpayments.py`. The step on ubuntu-latest is not
affected; what is affected is anybody checking it locally, and
CONTRIBUTING.md's whole "how to reproduce each CI job" section is an
invitation to do exactly that.

## What could be done

1. **Nothing in the tree, one line in CONTRIBUTING.md**: reproduce this step
   with `/usr/bin/grep`, and why. Cheapest, and it puts the knowledge where
   the invitation is.
2. **Make the step's verdict not an exit status.** It reads naturally and is
   portable:

   ```sh
   if [ ! -s files.txt ] || [ -n "$(grep -vE "$prose" files.txt)" ]; then
   ```

   The output *is* the answer here — the list of files the matrix has to
   check — so reading it is not the habit CLAUDE.md warns about under "read
   exit codes, not filtered output"; that rule is about a filter hiding a
   failure, and there is no failure to hide in a file list. Against it: a
   `run:` block is `bash -e`, and a command substitution whose command exits
   non-zero aborts the step, so this wants a `|| true` that itself needs
   explaining — and it changes a working step to help a machine the step
   never runs on.
3. **`grep -cvE` and a numeric test**, which both greps agree on. Same
   `bash -e` caveat as (2), and it counts where it only needs to know
   whether the count is zero.

(1) is enough and (2) is tempting: the argument for (2) is that a step whose
meaning depends on which `grep` is installed is a step nobody can check, and
the argument against is that the runner has one `grep` and it is the right
one. Either closes this; what should not stand is a documented local
reproduction that lies.

fametrano:
Closes #242, by option (1) — the option the issue's own comment argues is
proportionate to a blast radius of one line.

## What changed

One entry in CONTRIBUTING.md's "Running what CI runs" list, for
`Ask which files the pull request touches`. That job was missing from a
list whose first sentence promises *each* job of `lint`, `docs` and
`test`, and it is the job that decides whether the rest of them run — so
the gap and the trap are the same entry.

Nothing in the tree changes. `test.yml` keeps its `grep -qvE`, which is
right on the runner, and options (2) and (3) stay available if a later
change wants the verdict to stop being an exit status.

## The mechanism, narrower than the issue states

The issue says ugrep "takes the status from whether the pattern matched".
Measured on all three shapes of file list, with the real `prose=` pattern:

| file list | `/usr/bin/grep` | `grep` (ugrep 7.5.0) |
| --- | --- | --- |
| only prose | prose only | prose only |
| only code | everything runs | everything runs |
| mixed (#237's own) | everything runs | **prose only** |

So the two agree except where the list mixes the two kinds, which is where
`-v` has both a selected and an unselected line to distinguish. ugrep
inverts the *file's* match status rather than asking whether `-v` selected
anything, and the divergent direction is the expensive one: a matrix
skipped where it should have run. `-cv` agrees with both greps in all three
rows, which is why the wrong answer is silent — the output looks right and
only the status is wrong.

The last row is #237's actual file list (`CHANGELOG.md`, `CLAUDE.md`,
`CONTRIBUTING.md`, `HISTORY.md`, `README.md`, `RELEASING.md`,
`btclib_secp256k1/silentpayments.py`), fetched with the documented `gh api`
command and run through both greps: the documented recipe answers
"everything runs" and a bare `grep` answers "prose only", reproducing the
misreading that raised this.

## Checks

- `uvx pre-commit run --all-files` — exit 0, 37 hooks, no failures
- `https://ugrep.com` answers 200, so the new link does not cost `links` a
  `.lycheeignore` entry
- `grep -qv` still occurs exactly once outside `secp256k1/`, which is what
  lets the entry claim the reproduction is all ugrep touches here

## Summary by Sourcery

Document the CI gate that decides which test jobs run and clarify how to reproduce its result reliably.

Enhancements:
- Document how to reproduce the CI job that determines whether the remaining test workflow jobs run, including the correct grep implementation and workflow pattern.
- Document the required `test: every job passed` check alongside the other CI jobs.

Documentation:
- Update CONTRIBUTING.md and CHANGELOG.md with the missing CI reproduction guidance and the ugrep-specific caveat that can produce an incorrect result for mixed prose and code file lists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2104,6 +2104,27 @@ release-notes length in the first place, and are still in
   of the rest would be right today and stale the first time a sentinel
   is folded into the gate, which skips a matrix that should have run.
   The comment above the pattern carries that reasoning and the figure.
+- **and the list that promises to reproduce each gating job was missing
+  the job that decides whether the others run** (#242), which is the same
+  step again from the other side. Reproducing it by hand answers the
+  opposite where the shell's `grep` is ugrep: `-qv` there takes its
+  status from whether the pattern matched anywhere and inverts that,
+  rather than from whether `-v` selected a line, so a list mixing prose
+  with code — the case the step exists to judge — exits 1 and reads as
+  "prose only", the direction that skips a matrix that should have run,
+  and nothing says it has. Measured on each shape of file list with the
+  pattern taken out of the workflow, and against the runner rather than
+  against itself: #237's own list answers "everything runs" as its run
+  did, and the pull request carrying this entry answers "prose only" with
+  its matrix skipped. Nothing in the tree moves, `test.yml`'s `grep`
+  being the runner's, which is GNU's. The entry spells the path out,
+  lifts the pattern out of `test.yml` so the two cannot drift, and names
+  `-q` as the mechanism rather than `-v`: ugrep takes the same
+  file-level path when the output is discarded, which is where dropping
+  `-q` lands next, and `-cvE` with a numeric test is what agrees with
+  both greps. `test: every job passed` gets an entry in the same pass,
+  that being the neighbouring gap and the check a contributor sees red:
+  it reproduces as nothing, reading the conclusions of the jobs above it.
 
 ## v0.8.0.2
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -236,6 +236,35 @@ command at all, for the reason above, and no longer gates.
   uvx pre-commit run --all-files
   ```
 
+- `Ask which files the pull request touches`, whose answer decides whether
+  the rest of `test.yml` runs at all
+
+  ```shell
+  gh api "repos/btclib-org/btclib-secp256k1/pulls/<number>/files" \
+      --paginate --jq '.[].filename' > files.txt
+  eval "$(/usr/bin/grep -m1 '^ *prose=' .github/workflows/test.yml)"
+  if [ ! -s files.txt ] || /usr/bin/grep -qvE "$prose" files.txt; then
+      echo "everything runs"
+  else
+      echo "prose only"
+  fi
+  ```
+
+  The second line lifts the pattern out of the workflow rather than
+  restating it here, so the two cannot drift. `/usr/bin/grep` rather than
+  `grep`, because where the shell's is [ugrep](https://ugrep.com) the
+  third line answers the opposite question: ugrep takes the status from
+  whether the *pattern* matched anywhere and inverts that, rather than
+  from whether `-v` selected a line. A list of only prose and a list of
+  only code come out right either way; one mixing the two exits 1 and
+  reads as "prose only", which is the direction that skips a matrix that
+  should have run, and nothing says it has. `-q` is what asks the
+  question of the file rather than of the lines, and ugrep takes the same
+  path when the output is discarded, so `-cvE` and a numeric test agree
+  with both greps. The runner's `grep` is GNU's, so what ugrep can get
+  wrong here is the reproduction and not the workflow
+  (<https://github.com/btclib-org/btclib-secp256k1/issues/242>)
+
 - `Measure coverage, gated at 100%`
 
   ```shell
@@ -308,6 +337,11 @@ command at all, for the reason above, and no longer gates.
 
 - `Build on Linux for Windows` needs `mingw-w64`, and a Linux host to be
   faithful: the cross-compilation CI does is from ubuntu, not from macOS
+
+- `test: every job passed` reproduces as nothing, and is here because it
+  is the required check and therefore the one a contributor sees red: it
+  reads the conclusions of the jobs above it and runs no command of its
+  own. What turned it red is one of them, in the run
 
 - `Build the documentation`, the same command `.readthedocs.yaml` runs
   and `docs/README.rst` documents. Needs the submodule checked out,


fametrano:
## Review

Everything the description asserts was re-measured here rather than read,
on ugrep 7.5.0 `aarch64-apple-macosx` and GNU/BSD `/usr/bin/grep`, with the
`prose=` line taken out of `test.yml` rather than retyped.

### What reproduces

- the three-row table, with the real pattern: only prose and only code
  agree between the two greps; `#237`'s own list
  (`CHANGELOG.md CLAUDE.md CONTRIBUTING.md HISTORY.md README.md
  RELEASING.md btclib_secp256k1/silentpayments.py`) answers
  `everything runs` under `/usr/bin/grep` and `prose only` under `grep`
- the recipe as documented, run verbatim against two live pull requests:
  `244` → `prose only`, and this pull request's own run has the whole
  matrix `SKIPPED` with `Decide whether the matrix has anything to check`
  green; `237` → `everything runs`, which is what its run did. So the
  entry is checked against the runner and not only against itself
- `grep -qv` outside `secp256k1/`: `test.yml:180`, and nothing else
- `https://ugrep.com` → 200, so `links` needs no `.lycheeignore` entry
- the commit is signed, `"reason": "valid"`, Ferdinando M. Ametrano

### Please change

1. **"whose answer decides whether any job below it runs at all" is not
   true of the jobs below it in this list.** `needs.changes.outputs.code ==
   'true'` occurs on `test.yml`'s jobs only; `docs.yml`'s `docs` job — the
   `Build the documentation` entry, three entries further down — carries
   `if: ${{ !github.event.pull_request.draft }}` and nothing else, and
   `lint.yml`'s job the same. The list mixes three workflows, so "below it"
   reads as "below it here", where it is wrong in the direction that
   surprises: a contributor concludes a prose-only pull request runs no
   docs build. `whose answer decides whether the rest of test.yml runs at
   all` says what the job does and cannot be misread.

2. **"this is the tree's one `grep -qv`" states a count that nothing
   checks**, which CLAUDE.md rules out in those words. It is also the
   load-bearing half of the sentence: the moment a second one appears the
   file keeps saying ugrep touches nothing else here, and that is the
   expensive direction. The reassurance survives without the number —
   *The runner's `grep` is GNU's, so what ugrep can get wrong here is the
   reproduction, not the workflow* — and the tally stays in
   https://github.com/btclib-org/btclib-secp256k1/issues/242, with the
   command that re-derives it.

3. **No CHANGELOG.md bullet.** The unreleased section has a `Documentation`
   group, and the last ten commits touching CONTRIBUTING.md on `main` all
   carry an entry — `4747a9a` being the closest precedent, CONTRIBUTING.md
   alone plus its bullet. The two older exceptions (`5a1b39a`, `93d672b`)
   predate that run. What the bullet has to say is what a contributor
   reading only the changelog would want: the list promised each job and
   the gating one was missing, and reproducing it by hand under ugrep
   answers the opposite.

4. **The mechanism is `-q`'s, and the entry does not say so.** "The
   selected lines are still printed and counted correctly, so only the
   status diverges" is true, and reads as *ugrep's `-v` exit status is
   wrong*, which it is not. Measured on `#237`'s list, `-v` without `-q`:

   | where the output goes | `/usr/bin/grep` | `grep` (ugrep 7.5.0) |
   | --- | --- | --- |
   | a terminal, or a pipe | 0 | 0 |
   | `> out.txt` | 0 | 0 |
   | `> /dev/null` | 0 | **1** |

   So it is the quiet path that answers the file-level question, and ugrep
   takes that path when `-q` asks for it, when `-l` does, and on its own
   when the output goes to `/dev/null` — which is where a reader who
   "just drops `-q`" lands next, since discarding the output is the
   obvious next edit. One clause covers it: *`-q` is what asks the
   question of the file rather than of the lines, and ugrep takes the same
   path when the output is discarded, so `-cvE` and a numeric test agree
   with both greps.* That last is also why the count in the issue's option
   (3) is portable, which the entry currently leaves the reader to guess.

### Suggestions, take or leave

5. `prose=…` is the one line of the section a reader cannot paste. This
   pastes, and cannot go stale against `test.yml`:

   ```shell
   eval "$(/usr/bin/grep -m1 '^ *prose=' .github/workflows/test.yml)"
   ```

   Checked: it sets `prose` to the pattern, leading indentation and all.
   Against it, `eval` is a construct the rest of the file does not use.

6. `/usr/bin/grep` is right on both platforms CI and a contributor use, and
   absent on a machine that keeps no `/usr/bin` (NixOS, Guix). Not worth
   changing: the entry's value is that it is the workflow's line verbatim
   with one path spelled out, and 4's `-cvE` clause is the escape for
   anyone whose `grep` is neither.

7. `(#242)` is the only issue reference in CONTRIBUTING.md, and a bare
   `#N` is not autolinked in a rendered blob — CHANGELOG.md's own
   `Documentation` group spells the sibling case out in full
   (`<https://github.com/btclib-org/btclib-secp256k1/issues/161>`). Same
   here would cost one line of width.

8. **The list still does not keep its first sentence.** Census of the three
   gating workflows: `lint.yml` one job, `docs.yml` one, `test.yml` eleven
   — `changes`, `coverage`, `build-cibuildwheel`, `build-dynamic`,
   `build-sdist`, `suite-sdist`, `build-windows`, `check-dist`,
   `suite-dynamic`, `suite-static`, `test-passed`. With this entry in, all
   are named but `test-passed` / `test: every job passed`, which is a
   required check and the one a contributor sees red. It reproduces as
   nothing — it reads the results of the jobs above — and that is worth an
   entry saying so, or the intro sentence should promise each job *that
   runs something*. Closing the gap this pull request is about and leaving
   the neighbouring one is the shape that invites a second issue.

### Verdict

ACK once 1–4 are in — none of them touches the entry's substance, which
holds up: the divergence is real, the direction is the expensive one, and
the reproduction now answers what the runner answered on two live pull
requests. 5–8 are yours to weigh.

fametrano:
## Review addressed — 1–4 in, and 5, 7, 8 taken

`7086d97`. Each of the four was re-measured here before being written,
rather than taken on the review's word; all four reproduce.

### The four asked for

1. **"any job below it" → "the rest of `test.yml`".** Confirmed: `docs.yml`
   and `lint.yml` carry `if: ${{ !github.event.pull_request.draft }}` and
   nothing else, and `needs.changes.outputs.code` occurs in `test.yml`
   alone. The old wording put `Build the documentation` under a gate it
   does not have.

2. **The count is gone**, replaced by the reassurance that does not need
   it — *The runner's `grep` is GNU's, so what ugrep can get wrong here is
   the reproduction and not the workflow*. The tally stays in #242.

3. **CHANGELOG.md bullet**, appended to the unreleased `Documentation`
   group, immediately after the `#239` entry — which is the same step from
   the other side, and already says "with `/usr/bin/grep` rather than this
   machine's `grep`". `4747a9a`'s shape followed.

4. **`-q` named as the mechanism.** Re-measured on #237's list, and the
   third row is the one that matters:

   | where the output goes | `/usr/bin/grep` | `grep` (ugrep 7.5.0) |
   | --- | --- | --- |
   | a pipe | 0 | 0 |
   | `> out.txt` | 0 | 0 |
   | `> /dev/null` | 0 | **1** |
   | `-lv`, output discarded | 0 | **1** |

   So the quiet path is the file-level one and `-q` is what asks for it.
   The clause now says that, and names `-cvE` with a numeric test as the
   spelling both greps agree on — which is also why the issue's option (3)
   is portable.

### Of the suggestions

- **5, taken.** `eval "$(/usr/bin/grep -m1 '^ *prose=' .github/workflows/test.yml)"`
  sets `prose` to the pattern, indentation and all. It buys more than a
  pasteable block: the recipe now cannot drift from `test.yml`, which the
  elided `prose=…` could not promise. `eval` being unlike the rest of the
  file is the cost, and it is one line.
- **6**, nothing to change, agreed.
- **7, taken.** `<https://github.com/btclib-org/btclib-secp256k1/issues/242>`,
  matching the `Documentation` group's own sibling.
- **8, taken.** `test: every job passed` has an entry saying it reproduces
  as nothing and why it is worth naming anyway. With it the list keeps its
  first sentence: `lint` 1 + `docs` 1 + `test` 11 = 13 jobs, all now named.

### Re-run

The recipe pasted verbatim out of the file: `244` → `prose only`,
`237` → `everything runs`, which is what each run did. Gate exit 0, 37
hooks. Both commits `G`, Ferdinando M. Ametrano.

Two commits now, to be squashed locally before the fast-forward.

PR #245, opened 2026-08-17
Take the public key the caller already has, and say which way a check failed

fametrano:
`dsa.sign` derives the public key on every checked signature and drops
it. That derivation is a point multiplication and it is most of what
ECDSA's check costs over BIP340's, so a caller who already holds the key
-- a psbt signer, a wallet that has the address, anything signing twice
under one key -- pays for it anyway.

`dsa.sign` and `dsa._sign_` now take a keyword-only `pubkey`, and the
check verifies under it. Measured against a signature of 11.76
microseconds, an Apple M5, macOS 26.6, arm64, CPython 3.13.14, the five
shapes alternated in one process, 9 rounds of 3 000 calls with the
minimum of each kept, and the unchecked signature run again at the end
so the noise has a figure of its own:

| `dsa.sign` | per call | the check |
| --- | --- | --- |
| `verify=False` | 11.72 | |
| the key derived, as before | 31.92 | 20.20 |
| a compressed key handed in | 26.80 | 15.08 |
| an uncompressed key handed in | 24.78 | 13.06 |
| the point already parsed, through `_sign_` | 24.52 | 12.80 |
| `verify=False` again (the noise) | 11.76 | ±0.04 |

The 7.40 between the second row and the last is the derivation, which
is `secp256k1_ec_pubkey_create` through `keys._pubkey_from_prvkey_`;
timed alone in a session of the same shape it is 7.31, against a noise
of 0.14. The 7.55 the README carries and the 7.16 measured for
`secp256k1_keypair_create` are that other call, and land within half a
microsecond of it because the two do the same work. Within the
handed-in rows, 2.02 is what a compressed key costs to parse over an
uncompressed one -- the field square root that recovers y, which is why
the longer encoding is the cheaper argument -- and 0.26 is the
uncompressed parse itself: what the parse adds back above the fully
parsed row, not slices of the derivation. Under all of it is a bare
verification, which is what BIP340's check has always been.

## Taken on trust, and the failing branch pays for it

Checking the key against the private key on the way in would cost the
multiplication the argument exists to avoid. So it is not checked -- and
that changes what a failed check *means*, a key that is not this private
key's failing verification exactly as a faulted computation does.
Telling a caller their hardware is wrong because they mistyped an
argument is worse than the microseconds are worth.

`_checked` is where that is resolved and it is the whole of the new
logic: where a supplied key fails, and only there, it derives and asks
again. `RuntimeError` if the signature does not verify under the key the
private key actually has, which is the fault this check has always been
for; `ValueError` if it does and the supplied one is simply not it. The
rare branch pays back the derivation the common one saved.

## What the trust cannot do

It cannot let a bad signature through, and that is the property the
argument rests on. The keys a signature verifies under are a property of
that signature -- `recovery.recover` walks them -- so a key fixed before
the signature exists is not one of them, and the trust can cost a wrong
diagnosis but never a wrong success.
`test_a_key_fixed_in_advance_cannot_pass_a_signature_of_another_key`
holds that over forty keys and messages rather than leaving it in a
docstring.

## Refused rather than resolved

A key given beside `verify=False` raises: the check it is for is the one
that flag declines, and this package answers a caller contradicting
themselves rather than picking for them, as it already does for
`aux_rand32` beside `grind`. Octets that are not a key are parsed and
refused on the way in, so a mistyped argument does not arrive as a
failed check on a signature the caller is now holding.

## Not on `ssa`

The check there costs 13.2 microseconds whether the key is held or not,
so the argument would buy nothing and sell one thing: a second reason a
check can fail. Its absence is a decision.

## Verification

`uv run pytest` -- 647 passed, coverage back at 100. `uv run pre-commit
run --all-files` is clean, every hook, no skips.

`tests/test_verified_signing.py` gains seven cases: the signature
unchanged in both serializations and while grinding, the wrong key told
apart from a wrong computation, bad octets refused early, a key refused
beside `verify=False` and the same refusal reached through `_sign_`, the
property above, and the other half of the discrimination -- which needs
a substituted verification, `raise RuntimeError` being outside the
coverage ratchet, and which fails if the second verification is stubbed
to succeed.

The private half keeps its own copy of the contradiction rather than
leaning on `sign`'s: `_sign_` is the row the table recommends to a
caller holding a parsed point, so it is the one place a `pubkey` beside
`verify=False` must not be silently ignored.

`recovery` has the same derivation available to save and a third way for
the comparison to fail, so what it would take is
btclib-org/btclib-secp256k1#246 rather than a paragraph here.

The README's section on the check carries the design, the table and the
reasoning; the CHANGELOG carries the record behind it.

Part of btclib-org/btclib#982, whose comment carries the design this
implements. The `dsa.Signer` that issue also asks for is not here: it
needs the private key to live in memory this package owns, and `scalar`
coerces every path to an immutable `bytes`, so a `Signer` offering
`wipe` would promise what it cannot keep. What settling that would take
is btclib-org/btclib-secp256k1#247.

## Summary by Sourcery

Let checked ECDSA signing reuse a caller-provided public key and clearly classify verification failures.

New Features:
- Allow callers to provide an existing public key to `dsa.sign` and `dsa._sign_` for signature verification.

Bug Fixes:
- Distinguish an incorrect supplied public key from a signature verification failure using separate exceptions.
- Reject invalid public keys before signing and reject public keys when verification is disabled.

Enhancements:
- Avoid repeated public-key derivation during checked ECDSA signing while preserving verification safety guarantees.

Documentation:
- Document the supplied-public-key verification behavior, failure semantics, performance impact, and its deliberate scope.

Tests:
- Add coverage for supplied-key signing, validation, contradictory arguments, failure classification, and prevention of false verification success.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,86 @@ release-notes length in the first place, and are still in
 
 ## v0.8.0.3 (work in progress, not released yet)
 
+### The check takes a public key the caller already has
+
+- **`dsa.sign` and `dsa._sign_` take a keyword-only `pubkey`**, the key
+  the check verifies under, so that a caller who already holds it does
+  not pay for deriving it again. It is taken on trust and never checked
+  against the private key on the way in: checking it would cost the
+  point multiplication the argument exists to avoid.
+- **What that saves is most of what ECDSA's check costs over BIP340's.**
+  Measured on an Apple M5, macOS 26.6, arm64, CPython 3.13.14, the five
+  shapes alternated in one process, 9 rounds of 3 000 calls with the
+  minimum of each kept, and the unchecked signature run again at the end
+  so the noise has a figure of its own -- ±0.04 against an unchecked
+  11.72. The check is **20.20** deriving the key per call, **15.08** with
+  a compressed key handed in, **13.06** with an uncompressed one, and
+  **12.80** with the point already parsed and `_sign_` called directly.
+  The 7.40 between the first and the last is the derivation, which is
+  `secp256k1_ec_pubkey_create` through `keys._pubkey_from_prvkey_`;
+  timed alone in a session of the same shape it is 7.31, against a
+  noise of ±0.14. The 7.55 the README carries and the 7.16 measured for
+  `secp256k1_keypair_create` are that other call rather than this one,
+  and land within half a microsecond of it because the two do the same
+  work. Within the handed-in rows, 2.02 is what a compressed key costs
+  to parse over an uncompressed one -- the field square root -- and
+  0.26 is the uncompressed parse: what the parse adds back above the
+  fully parsed row, not slices of the derivation. Under all of it is a
+  bare verification, which is what BIP340's check has always been.
+- **A failed check now says which of two things failed**, and that is
+  what makes the trust affordable rather than merely cheap. A key that
+  is not this private key's makes the verification fail exactly as a
+  faulted computation does, and reporting one as the other would tell a
+  caller their hardware is wrong because they passed the wrong argument.
+  So the failing branch, and only the failing branch, derives the key
+  and asks again: `RuntimeError` where the signature does not verify
+  under the key the private key actually has, `ValueError` where it does
+  and the one handed in is simply not it. The rare branch pays the
+  derivation the common one saved.
+- **The trust cannot let a bad signature through**, which is the
+  property the whole argument rests on and is now a test rather than a
+  paragraph. The keys a signature verifies under are a property of that
+  signature -- `recovery.recover` walks them -- so a key fixed before
+  the signature exists is not one of them.
+  `test_a_key_fixed_in_advance_cannot_pass_a_signature_of_another_key`
+  holds that over forty keys and messages.
+- **And it catches a fault the derived check cannot.** A private key
+  corrupted before it was signed with passes the derived check in
+  silence: the signature and the key it is verified against come out of
+  the same corrupted octets and agree. A key handed in came from
+  somewhere no fault in that call could reach, so it does not agree --
+  which means the `ValueError` names the likelier of two causes rather
+  than the only one, and the docstring says so.
+- **Both halves of the discrimination are held to.** The `ValueError`
+  side has a case of its own; the `RuntimeError` side has one too, and
+  it needs a substituted verification because no input reaches it --
+  `raise RuntimeError` is outside the coverage ratchet, so nothing else
+  would have said that stubbing the second verification to succeed
+  turns every genuine fault under a handed-in key into a report that
+  the caller mistyped an argument. That mutation now fails the suite.
+- **Refused rather than resolved**, as `aux_rand32` beside `grind`
+  already is: a key given with `verify=False` raises, being a caller
+  contradicting themselves, and octets that are not a key are parsed and
+  refused on the way in rather than arriving as a failed check on a
+  signature the caller now holds.
+- **Not on `ssa`, and deliberately.** The check there is 13.2
+  microseconds whether the key is held or not, the keypair already
+  holding the point, so the argument would buy nothing and sell one
+  thing: a second reason a check can fail, and the discrimination step
+  that reason costs. btclib-org/btclib#982 has the measurement.
+- **Not on `recovery` yet, and that is a decision to take rather than
+  one taken.** `_abort_unless_recovered` derives the same key --
+  `keys._pubkey_cmp_(recovered, keys._pubkey_from_prvkey_(...))` -- so
+  the saving available there is exactly the one taken here: the recovery
+  stays and the derivation goes. What differs is the discrimination, a
+  recovered key that is not the one handed in having a third possible
+  cause the two schemes above do not, which is the recovery id. It wants
+  its own measurement and its own cases, so it is left to #246 rather
+  than folded in here.
+- **The README carries all of this**, in the section on the check, which
+  said `verify=False` was the only thing a caller could do about the
+  cost and is where CLAUDE.md puts the design.
+
 ### A signer checks its own signature
 
 - **`dsa.sign` and `ssa.sign` verify what they made before answering
--- a/README.md
+++ b/README.md
@@ -636,6 +636,67 @@ already, which is the same 7.55 the [signer](#two-outposts-past-the-boundary)
 hoists. So the step the specification prescribes is also the cheaper of
 the two, and a `Signer` pays it at the same price as a bare `ssa.sign`.
 
+**`verify=False` is not the only thing a caller can do about that gap.**
+`dsa.sign` takes a `pubkey`, the key the check verifies under, so that a
+caller already holding it does not have the multiplication done again.
+That is most of what ECDSA's check costs above BIP340's, and handing the
+key in brings the two to the same operation — a bare verification. In a
+session of its own, an Apple M5, macOS 26.6, arm64, CPython 3.13.14, nine
+rounds of 3 000 calls with the minimum of each kept, and the unchecked
+signature run a second time so the noise has a figure of its own — its
+`dsa.sign` rows sitting a few tenths under the two sessions above, which
+is between sessions and not within any of them:
+
+| `dsa.sign` | per call | the check |
+| --- | --- | --- |
+| `verify=False` | 11.72 | |
+| the key derived, as before | 31.92 | 20.20 |
+| a compressed key handed in | 26.80 | 15.08 |
+| an uncompressed key handed in | 24.78 | 13.06 |
+| the point already parsed, through `_sign_` | 24.52 | 12.80 |
+| `verify=False` again (the noise) | 11.76 | ±0.04 |
+
+The 7.40 between the second row and the last is the derivation, which is
+`secp256k1_ec_pubkey_create` through `keys._pubkey_from_prvkey_`. Timed
+alone in a session of the same shape it is 7.31, against a noise of
+±0.14. The 7.55 above is a different call — the keypair's own
+multiplication — and lands within half a microsecond of it because the
+two do the same work.
+
+Within the rows where the key is handed in, 2.02 is what a *compressed*
+key costs to parse over an uncompressed one — the field square root that
+recovers y, which is why the longer encoding is the cheaper argument
+here — and 0.26 is the uncompressed parse itself, which only the private
+half avoids. Those are what the parse adds back above the fully parsed
+row, not slices of the derivation above.
+
+The key is taken on trust: checking it against the private key would
+cost the multiplication the argument exists to save. What that would
+otherwise confuse is a wrong argument with a wrong computation, since a
+key that is not this private key's fails verification exactly as a fault
+does, so the failing branch — and only that one — derives the key and
+asks again: `RuntimeError` where the signature does not verify under the
+key the private key actually has, `ValueError` where it does. A key
+handed in beside `verify=False` is refused rather than ignored.
+
+What the trust cannot do is pass a bad signature. The keys a signature
+verifies under are a property of that signature, so a key fixed before
+the signature exists is not one of them, and the argument can cost a
+wrong diagnosis but never a wrong success. It can also catch what the
+derived check cannot: a private key corrupted before it was signed with
+agrees with a public key derived from the same corrupted octets, and
+does not agree with one that came from anywhere else.
+
+`ssa.sign` takes no such argument, and that is a decision rather than an
+omission: its check is a bare verification already, the keypair holding
+the point, so there would be nothing to save and one more way for a
+check to fail. `recovery.sign` is the one where the same saving is
+available and not yet taken — its check derives the very same key before
+comparing what it recovered against it — and what it would cost to take
+is
+[issue 246](https://github.com/btclib-org/btclib-secp256k1/issues/246):
+a third way for the comparison to fail, which is the recovery id.
+
 **`recovery.sign` recovers instead of verifying**, and the difference is
 the recovery id. A verification does not look at it: a recoverable
 signature carrying the wrong one verifies perfectly and then recovers a
--- a/btclib_secp256k1/dsa.py
+++ b/btclib_secp256k1/dsa.py
@@ -253,13 +253,85 @@ def _grind(signature: CData, msg_bytes: bytes, prvkey_bytes: bytes) -> None:
         _signed(signature, msg_bytes, prvkey_bytes, entropy)
 
 
-def _sign_(
+def _checked(
+    signature: CData, msg_bytes: bytes, prvkey_bytes: bytes, pubkey: CData | None
+) -> None:
+    """Verify what was just signed, and say which of two things failed.
+
+    The check itself is one call, and everything around it is about a
+    public key a caller may have handed in. It is taken on trust: a
+    signer that derived it to check it would have paid the point
+    multiplication the argument exists to avoid, so what is skipped on
+    the way in is paid for on the way out, and only where it is owed.
+
+    A key that is not this private key's makes the verification fail, and
+    a verification that fails means something quite different -- the
+    computation having gone wrong, memory gone bad or a fault induced,
+    which is what this check is here to catch at all. Reporting one as
+    the other would tell a caller their hardware is faulty because they
+    passed the wrong argument. So the failing branch, and only the
+    failing branch, derives the key and asks again: it is the rare one by
+    construction, and it is where the microseconds saved everywhere else
+    are worth giving back.
+
+    What the trust cannot do is let a bad signature through. The keys
+    under which a signature verifies are a property of that signature --
+    for ECDSA they are what `recovery.recover` walks -- so a key fixed
+    before the signature exists cannot be one of them except by knowing
+    it in advance.
+    `test_a_key_fixed_in_advance_cannot_pass_a_signature_of_another_key`
+    is that property rather than this paragraph.
+
+    What it catches that the derived path cannot is worth knowing too,
+    because it is more than a mistyped argument. A private key corrupted
+    before it was signed with passes the derived check silently: the
+    signature and the key it is checked against both come out of the
+    same corrupted octets, so they agree. A key handed in came from
+    somewhere no fault in this call could have reached, and does not
+    agree -- so the `ValueError` below is raised over a fault the other
+    branch has no way of seeing, and names only the likelier of its two
+    causes.
+
+    Args:
+        signature: the signature just produced, parsed.
+        msg_bytes: the 32-byte hash it was made over, already checked.
+        prvkey_bytes: the 32-byte private key that made it, already
+            checked, and what the failing branch derives from.
+        pubkey: the public key to check against, parsed, or None to
+            derive it -- which is what a caller holding nothing gets.
+
+    Raises:
+        RuntimeError: if the signature does not verify under the key this
+            private key actually has, which is the fault this is for.
+        ValueError: if it verifies under that key but not under the one
+            handed in -- the wrong key given, most likely, or else a
+            private key that was already corrupt when it was signed
+            with, as the paragraph above says.
+    """
+    if pubkey is not None:
+        if _verify_(msg_bytes, pubkey, signature):
+            return
+        # the given key did not verify it, and the two reasons for that
+        # are told apart here rather than guessed at
+        if _verify_(msg_bytes, keys._pubkey_from_prvkey_(prvkey_bytes), signature):
+            raise ValueError("the public key given is not this private key's")
+        raise RuntimeError("signing produced a signature that does not verify")
+
+    if not _verify_(msg_bytes, keys._pubkey_from_prvkey_(prvkey_bytes), signature):
+        raise RuntimeError("signing produced a signature that does not verify")
+
+
+# six arguments here for the reason `sign` below carries at greater
+# length: the four ECDSA questions plus the two this package adds, and
+# `pubkey` is `verify`'s argument rather than a group with any of them
+def _sign_(  # noqa: PLR0913
     msg_bytes: BytesLike,
     prvkey: BytesLike | int,
     aux_rand32: BytesLike | None = None,
     grind: bool = False,
     *,
     verify: bool = True,
+    pubkey: CData | None = None,
 ) -> CData:
     """Create an ECDSA signature, as the parsed signature.
 
@@ -284,6 +356,11 @@ def _sign_(
             signature the grinding settled on: the attempts it discarded
             are never answered with, so a fault in one of them is a
             wasted attempt rather than a published signature.
+        pubkey: the public key to check under, parsed, where the caller
+            already holds it and would rather not have it derived again.
+            Taken on trust; `_checked` says what that does and does not
+            risk. Refused beside `verify=False`, which declines the
+            check it is for.
 
     Returns:
         The libsecp256k1 signature object, in the lower-s form
@@ -292,8 +369,12 @@ def _sign_(
     Raises:
         ValueError: if the message hash is not 32 bytes, if aux_rand32 is
             given and is not 32 bytes, if aux_rand32 and grind are given
-            together, or if the private key is not 32 bytes, does not fit
-            in them, or is not in [1, n-1].
+            together, if the private key is not 32 bytes, does not fit
+            in them, or is not in [1, n-1], if a `pubkey` is given beside
+            `verify=False`, or if the signature verifies under the key
+            this private key has and not under the one given -- the
+            wrong key handed in, most likely, and `_checked` says what
+            else it can be.
         RuntimeError: if libsecp256k1 fails to serialize an attempt while
             grinding, which a signature it has just made cannot do, or if
             `verify` asks and the signature does not verify.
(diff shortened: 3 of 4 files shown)


fametrano:
## Review

The design is right and the argument for it is made where it belongs: the key
is taken on trust, the trust is paid back only on the branch that fails, and
the property that makes it safe is a test rather than a paragraph. What follows
is four things to change before this lands and four worth considering. No ACK
yet.

### Changes requested

**1. The three "steps" do not reconcile with the table they decompose.**
The table's own differences are 6.01, 2.01 and 0.68, spanning 8.70 from 21.07
to 12.37. The prose calls them "7.7 the derivation, 2.0 the field square root
a compressed key costs to parse, 0.9 the parse of key and signature" — which
sums to 10.6, and only the middle figure matches a gap in the table. 7.7 is in
the neighbourhood of the 7.55 and 6.8 the README already carries for that same
multiplication, so these look like figures from another session presented as
this table's arithmetic. Either give the differences this table actually has,
or say the three were measured separately and against what.

Related, and the reason the smallest of them is the one at risk: the table has
no noise row. Every other benchmark in the README carries one (`ssa.sign` again
(the noise), ±0.04; the `recovery` session, ±0.02) precisely so that a
difference of a few tenths can be told from the measurement. The claim at issue
here is 0.68 microseconds.

**2. Two counts that nothing checks.**
`test_a_key_fixed_in_advance_cannot_pass_a_signature_of_another_key` loops
`range(PRVKEY + 1, PRVKEY + 40)` — 39 keys, where the CHANGELOG and the
description both say "over forty keys and messages". And `_checked`'s docstring
says "Measured over five hundred signatures under other keys, a wrong key fixed
in advance verified none of them", which no test in the suite does. The
description's own sentence — held in a test "rather than leaving it in a
docstring" — is the argument for making the docstring cite the test's number.

**3. The `RuntimeError` half of the discrimination is untested, and coverage
cannot say so.** `raise RuntimeError` is excluded from the ratchet by design, so
the 100% here is silent about the branch that reports a genuine fault under a
handed-in key. `test_the_wrong_key_is_told_apart_from_a_wrong_computation` is
named for the discrimination and asserts one side of it: nothing in the file
passes `pubkey=` with a verification substituted.

What that leaves open is the inversion of the failure this change exists to
prevent. Replace

```python
if _verify_(msg_bytes, keys._pubkey_from_prvkey_(prvkey_bytes), signature):
```

with `if True:` and the suite stays green, while every real fault met under a
handed-in key is reported as "the public key given is not this private key's" —
a caller told they mistyped an argument because their hardware is wrong, which
is the mirror image of the misdiagnosis the docstring refuses. The file already
has what it takes: `_refuse_every_check(patch)` and
`dsa.sign(MSG, PRVKEY, pubkey=<the correct key>)`, expecting `RuntimeError` and
"does not verify". `mutation` would report this as a survivor on Sunday.

**4. The README section on the check is now incomplete.** "Checking the
signature before answering with it" states the cost as "what a verification
costs, and in ECDSA a point multiplication besides", with `verify=False` as the
only thing a caller can do about it — and that is no longer true. Per CLAUDE.md
the design lives in the README, and this is the section where a reader learns
what the check costs and what their options are; the CHANGELOG is the record
behind it, not a substitute for it. The last paragraph of that section ("Every
other microsecond figure in this file was measured before this argument
existed") is the shape the addition wants.

### Worth considering

**5. `_checked`'s control flow.** The `elif`/`else` where the `else` always
raises, followed by a trailing `if not _verify_(...)` reachable only from the
first branch, costs a reader a re-read to establish that. The handed-in key is
a whole case and reads as one:

```python
if pubkey is not None:
    if _verify_(msg_bytes, pubkey, signature):
        return
    # the given key did not verify it, and the two reasons for that are
    # told apart here rather than guessed at
    if _verify_(msg_bytes, keys._pubkey_from_prvkey_(prvkey_bytes), signature):
        raise ValueError("the public key given is not this private key's")
    raise RuntimeError("signing produced a signature that does not verify")

if not _verify_(msg_bytes, keys._pubkey_from_prvkey_(prvkey_bytes), signature):
    raise RuntimeError("signing produced a signature that does not verify")
```

Same two raises as now, with the derived path left as the tail it was.

**6. `recovery` gets no sentence either way.** The CHANGELOG argues `ssa`'s
absence and is silent about `recovery.sign`, whose `_abort_unless_recovered`
derives the very same key —
`keys._pubkey_cmp_(recovered, keys._pubkey_from_prvkey_(prvkey_bytes))` — and
whose check the README puts at 22.4 microseconds against ECDSA's 19.5. The
saving there is exactly the one claimed here: the recovery stays, the
derivation goes. Whichever way it is decided, a section that names `ssa` and
skips `recovery` reads as an oversight rather than a decision.

**7. The `ValueError` means more than a wrong argument, and the docstring could
say so.** With the key derived, a private key corrupted *before* signing passes
the check silently: the signature and the key both come from the same corrupted
bytes, and they agree. With the key handed in, they do not agree, because the
caller's key is material no fault in `prvkey_bytes` could have touched — so
this path catches a fault the old one never could, and calls it "the public key
given is not this private key's". True as written, and an understatement of
what it found.

**8. Nit on the order of two refusals.** In `sign`, `keys.parse(pubkey)` is
evaluated in the argument list, so it runs before `_sign_` sees the
`verify=False` contradiction: `sign(..., verify=False, pubkey=b"junk")` reports
"invalid public key". Both are `ValueError` and neither is wrong; it is only
worth deciding which one a caller contradicting themselves should be told about
first.

fametrano:
All eight taken, at `68461bd`. The first was a real error and the third
found a defect the suite could not have.

### 1. The numbers were from two sessions, and now they are from one

You are right, and worse than you put it: 7.7 came from an earlier probe
whose rows were composed differently, so it was not this table's arithmetic
at all. Re-measured, alternated in one process, with the noise row the claim
needed — 9 rounds of 3 000, minimum kept:

| `dsa.sign` | per call | the check |
| --- | --- | --- |
| `verify=False` | 11.72 | |
| the key derived, as before | 31.92 | 20.20 |
| a compressed key handed in | 26.80 | 15.08 |
| an uncompressed key handed in | 24.78 | 13.06 |
| the point already parsed, through `_sign_` | 24.52 | 12.80 |
| `verify=False` again (the noise) | 11.76 | **±0.04** |

The decomposition now comes out of the table rather than beside it. The whole
span, 7.40, is the derivation — and it agrees with the 7.55 the README
already carries and with `secp256k1_keypair_create` timed alone at 7.16, so
three routes reach it. Within the handed-in rows, 2.02 is what a compressed
key costs to parse over an uncompressed one, and 0.26 is the uncompressed
parse itself. Against ±0.04 the smallest of the three is six times the noise,
which is what your point about the missing noise row was for.

The 0.26 also replaces the old 0.9 for a reason worth naming: the last row
now goes through `_sign_` and `serialize_der`, which is the shape a caller
holding a parsed point actually writes, rather than through `_verify_`
directly as the earlier probe did.

### 2. Two counts that nothing checked

`range(PRVKEY + 1, PRVKEY + 41)` — forty keys, as everything about it said.
And `_checked`'s docstring no longer cites five hundred signatures from a
throwaway probe: it names the test, which is the argument the description was
already making about docstrings.

### 3. The `RuntimeError` half, and your mutation

This was the one worth the review. `test_a_fault_under_a_handed_in_key_is_not_reported_as_a_wrong_key`
uses `_refuse_every_check(patch)` with the correct key handed in, so the only
thing that can fail either verification is the substitution and what has to
arrive is the fault.

Run against your mutation:

```text
if True:   ->  1 failed, 58 passed
restored   ->  59 passed
```

The inversion you describe is exactly what it catches, and it is the mirror
of the misdiagnosis the design refuses: a caller told they mistyped an
argument because their hardware went wrong.

### 4. The README

The section on the check now carries the argument, the table and the
decomposition, and no longer says `verify=False` is the only thing a caller
can do about the cost. It sits after the paragraph you point at, in the same
shape.

### 5. `_checked`'s control flow

Taken verbatim. The handed-in key is a whole case that returns or raises, and
the derived path is the tail it was.

### 6. `recovery`

Decided rather than skipped, and the decision is "not here". The saving is
the same one — `_abort_unless_recovered` derives the very key it compares
against — but the discrimination is not: a recovered key that is not the one
handed in has a third possible cause the other two schemes do not have, which
is the recovery id. That wants its own measurement and its own cases rather
than being folded into a change about `dsa`. The CHANGELOG says that, in
those terms.

### 7. The `ValueError` means more than a wrong argument

Taken, and it is the sharpest thing in the review. With the key derived, a
private key corrupted before signing passes the check in silence — signature
and key come out of the same corrupted octets and agree. With the key handed
in, they do not, because the caller's key came from somewhere no fault in
that call could reach. So this branch sees a fault the other one cannot, and
`_checked` now says the `ValueError` names the likelier of two causes rather
than the only one.

### 8. The order of the two refusals

Moved: the `verify=False` contradiction is raised in `sign` before
`keys.parse` runs. A caller who asked for no check and handed in a key to
check with should hear about the two arguments, not about the octets of one
of them — which is what parsing first told them.

---

`uv run pytest` — 646 passed. `uv run pre-commit run --all-files` clean, with
`submodule-pin` still skipped: it reports README naming v0.8.0 as `bea100b`
against a submodule pinned to `6e2c8bc`, which is `main`'s state and untouched
here. Worth an issue of its own.

fametrano:
## Correction: there is no `submodule-pin` problem, and there never was

I said twice — in the commit message and in the reply above — that
`submodule-pin` fails on `main` and that I had skipped it for that reason.
That was wrong, and it was wrong because I checked the wrong thing.

What I checked was that the gitlink is `6e2c8bc` in both my worktree and the
main checkout, and concluded from the hook's message that the failure was
therefore pre-existing. What I never checked was whether the hook fails on
`main` at all. It does not:

```console
$ git submodule status secp256k1
 6e2c8bc4ecdc6e71dbe7a368f360d8d453ce435d secp256k1 (v0.8.0)
```

and upstream agrees — `v0.8.0` resolves to `6e2c8bc`. `bea100b` appears
nowhere in the tree; it was what the hook reported having nothing to resolve
`v0.8.0` against.

The cause was mine. `git worktree add` does not populate submodules, so
`secp256k1/` was empty; I then copied the directory from the main checkout to
get the build working, and deleted the `.git` file inside it because
`detect-secrets` was choking on a gitlink pointing at a module dir that does
not exist in a worktree. That left the hook with a checkout and no refs.

`git submodule update --init secp256k1` is what it wanted, and the objects
were already local, so it needed no network. With that:

```text
uv run pre-commit run --all-files   # every hook, no SKIP
```

is clean. The commit message no longer mentions a skip, and `7c8113a` is the
same change with that sentence corrected.

Nothing about the code changed. What changed is that a claim about this
repository's `main` which I had no evidence for is withdrawn — and the
convenience of it being someone else's problem is exactly why it deserved
the check it did not get.

fametrano:
## Second pass, at `68461bd`

Six of the eight are settled and two of them well: the fault case is the test
this needed, and `_checked` now reads in one pass. The re-measurement answers
the first point properly — the decomposition comes out of the table, the noise
row is there, and 12.80 landing beside the 12.7 this section already gives
BIP340's check is the sentence "brings the two to the same operation" earning
itself. What follows is four things still to change, three of them in the prose
of the re-measurement, and four worth considering. Still no ACK.

### Changes requested

**1. Coverage is red, and the line is `_sign_`'s own refusal.**
[The job](https://github.com/btclib-org/btclib-secp256k1/actions/runs/32065257815/job/95495738697)
says it exactly:

```text
btclib_secp256k1/dsa.py     124      1     34      1  98.73%   382
ERROR: Coverage failure: total of 99.80 is less than fail-under=100.00
```

382 is `raise ValueError("pubkey is for the check that verify=False declines")`
inside `_sign_`. Moving the contradiction up into `sign` is right, and it made
the private half's copy unreachable from every test in the suite, all of which
go through `sign`.

Keep the copy: `_sign_` is not an internal detail here but the row the new
table recommends — "the point already parsed, through `_sign_`" is a caller who
has done the work to save 0.26 microseconds and is exactly the one who should
not have a `pubkey` silently ignored beside `verify=False`. What it wants is
the case that reaches it, which is one line and has a precedent in this file:
`test_every_signer_defaults_to_checking` exists because a private half left at
a weaker policy is the hole the public ones closed.

```python
with pytest.raises(ValueError, match="verify=False declines"):
    dsa._sign_(MSG, PRVKEY, verify=False, pubkey=keys._pubkey_from_prvkey_(PRVKEY_BYTES))
```

**2. The corroboration is of a different libsecp256k1 call.** Both the CHANGELOG
and the README say the 7.40 "agrees with the 7.55 the README already carries for
that multiplication and with `secp256k1_keypair_create` measured alone at 7.16".
But the derivation this table spans is `keys._pubkey_from_prvkey_`, which calls
`secp256k1_ec_pubkey_create`; `secp256k1_keypair_create` is what `_secret.py`
builds, and the 7.55 above *is* that keypair — "the difference between them is
`secp256k1_keypair_create` on any machine … of which the keypair is 7.55". So
the sentence cites the same call twice and neither time the one being decomposed.

The agreement still means something, the two being one point multiplication
each and within half a microsecond of one another — but that is the sentence to
write. Naming `secp256k1_ec_pubkey_create` as what the 7.40 is, and the other
two as the keypair's multiplication of the same shape, costs a clause. Or time
`ec_pubkey_create` alone, which would make the third route the one claimed.

**3. "Of it" attributes to the derivation two figures that are not in it.**
Same sentence, both files: "The 7.40 … is the derivation … Of it, 2.02 is what a
compressed key costs to parse over an uncompressed one … and 0.26 is the
uncompressed parse." Those two are not slices of the 7.40 — they are what the
parse adds back above the fully-parsed row, on the other side of the ledger from
the multiplication. Your comment above gets it right, and the phrase is already
written there: *within the handed-in rows*.

**4. `_checked`'s `Raises` now contradicts `_checked`'s body.** The paragraph
added for point 7 says the `ValueError` "is raised over a fault the other branch
has no way of seeing, and names only the likelier of its two causes". Eight
lines below, unchanged:

```text
ValueError: if it verifies under that key but not under the one
    handed in, which is a wrong argument and not a fault.
```

The second half of that clause is the thing the new paragraph exists to deny.

### Worth considering

**5. The test module's docstring is the file's map, and no longer maps it.** It
enumerates what the file holds — "what is tested is the four things around it",
and `recovery` "adds a fifth" — and there are now six cases about a key handed
to the check, one of which is the strongest thing in the file. A sentence in
that opening would carry them.

**6. Two tables in one README section, disagreeing by more than either noise
row.** `dsa.sign` is 12.15 and 31.67 above, 11.72 and 31.92 below: 0.43 and 0.25
apart, against ±0.04 and ±0.02. Both tables declare their own session, so
nothing is wrong — but this section already has the habit of saying so out loud,
the `recovery` table being introduced by "re-measured `dsa.sign` beside it so
that the two are comparable … agreeing with the row above". A clause pointing at
the earlier rows would keep a reader from checking the arithmetic and finding it
does not close.

**7. The comment above the new `if` in `sign`.** The composition comment
("composed, where `verify` below is spelled out … two python frames do not show
against it") now runs straight into the contradiction comment, and describes a
call four lines further down. A blank line between them, or the refusal above
both, puts each back beside what it is about.

**8. An issue for `recovery`.** "Left to the change that gives it them" reads
better pointing at something — the same treatment you are giving the
`submodule-pin` skip.

fametrano:
## Third pass, at `7c8113a` — the push crossed the review

`7c8113a` is `68461bd`'s tree exactly (`git diff --quiet` between the two says
so); what changed is the commit message and the claim withdrawn with it. So the
second pass stands as written, all four of its requested changes included, and
there is nothing new to review yet. No ACK.

The withdrawal is worth more than the sentence it removed. `submodule-pin`
passing on `main` means the gate ran in full behind "clean", where a skip had
been standing in for it — and CLAUDE.md's worktree recipe names the cause in
its own words: a linked worktree gets no submodule, and
`git submodule update --init secp256k1` before `uv sync` is the line that makes
the rest of the recipe run. Checking whether the hook fails on `main`, rather
than inferring it from a gitlink that matched, is the difference between the
two, and it is the same discipline as reading a log instead of predicting one.

The state to clear, unchanged from the second pass:

- **coverage is still red**, same line, on
  [the run for this head](https://github.com/btclib-org/btclib-secp256k1/actions/runs/32065838316/job/95497748245):
  `btclib_secp256k1/dsa.py 124 1 34 1 98.73% 382`, which is `_sign_`'s own
  refusal of a `pubkey` beside `verify=False`, unreachable from every test
  since the contradiction moved up into `sign`
- **`secp256k1_keypair_create` is not the call the 7.40 decomposes**, that
  being `secp256k1_ec_pubkey_create` through `keys._pubkey_from_prvkey_`
- **"Of it, 2.02 … and 0.26 …"** puts on the derivation's side of the ledger
  two figures the parse adds back on the other
- **`_checked`'s `Raises` still says "a wrong argument and not a fault"**, which
  is what the paragraph above it now denies

fametrano:
## Fourth pass, at `c9ad984` — all eight of the second pass taken

The four requested and the four worth considering. Coverage is back at
100, and the second point turned out to be worth measuring rather than
rephrasing.

### 1. Coverage: the case that reaches `_sign_`'s own refusal

Taken as written, and the argument for keeping the copy is the one now in
the test's docstring: `_sign_` is the row the table recommends to a caller
holding a parsed point, so it is the one place a `pubkey` beside
`verify=False` must not be silently ignored.

```text
btclib_secp256k1/dsa.py    100%
TOTAL   826   0   182   0   100.00%
```

647 passed.

### 2. I timed `ec_pubkey_create` instead of rewriting the clause

You offered the clause or the measurement; the measurement is cheap and
makes the third route the one actually claimed. Same shape as the table —
alternated in one process, 9 rounds of 3 000, minimum kept, the anchor
run twice:

| the row | per call |
| --- | --- |
| `dsa.sign`, `verify=False` (the anchor) | 11.68 |
| `keys._pubkey_from_prvkey_` alone | **7.31** |
| `dsa.sign`, the key derived | 31.35 |
| `dsa._sign_`, the point handed in | 23.40 |
| the anchor again (the noise) | 11.82, **±0.14** |

So the derivation is `secp256k1_ec_pubkey_create`, and 7.31 is what it
costs on its own. Both files now name it as that, and say the 7.55 and
the 7.16 are the keypair's own multiplication — a point multiplication of
the same shape, within half a microsecond, rather than the same call.

Worth reporting rather than hiding: that session's own span between its
third and fourth rows is 7.96, not the table's 7.40. It is a different
session and its rows are not the table's rows — the fourth here is
`_sign_` returning the parsed signature, where the table's is serialized
— so the 7.31 is offered as an independent figure for the call, not as
this table's arithmetic re-derived.

### 3. "Of it" → "within the handed-in rows"

Taken verbatim, both files, with the ledger said out loud: those two are
what the parse adds back above the fully parsed row, not slices of the
derivation.

### 4. `_checked`'s `Raises`

You are right that the clause denied its own paragraph. It now names both
causes in the order the paragraph puts them:

```text
ValueError: if it verifies under that key but not under the one
    handed in -- the wrong key given, most likely, or else a
    private key that was already corrupt when it was signed
    with, as the paragraph above says.
```

### 5. The module docstring maps the new group

A paragraph after `recovery`'s, saying what the six turn on — the key
taken on trust, what that costs, and what it cannot cost — and naming the
property as the one case here that is a property over many keys rather
than a single case.

### 6. The two tables

Said out loud, in the section's own habit: the new table is its own
session, and its `dsa.sign` rows sit a few tenths under the two above,
which is between sessions and not within any of them.

### 7. The two comments in `sign`

The refusal moved above both, which is your second suggestion and the
better one — a blank line would have left the composition comment
describing a call four lines away. Each comment is now beside what it is
about.

### 8. #246

[Filed](https://github.com/btclib-org/btclib-secp256k1/issues/246), and
both files now point at it instead of at a future change. It carries the
saving (`_abort_unless_recovered` derives the very key it compares
against), the reason it is not this change (a third cause — the recovery
id — so the two-way branch from `dsa` does not transfer), and what it
needs: its own table in the shape of the one here, cases for a right key
with a wrong id, and the same two refusals. It also says the answer is
not assumed: `recovery`'s check is the dearest of the three, so the
saving is largest and so is the number of ways a diagnosis can be wrong.

---

`uv run pytest` — 647 passed, coverage 100. `uv run pre-commit run
--all-files` — every hook, no skips.

fametrano:
## Fourth pass, at `c9ad984`

All eight taken, and two of them better than asked: timing
`ec_pubkey_create` instead of rewriting the clause makes the third route the
one actually claimed — 7.31 against the table's 7.40, inside a ±0.14 — and
reporting that session's own 7.96 span rather than quietly keeping the figure
that agreed is the half of a measurement that usually goes missing. #246 is the
shape that decision wanted: the saving, the reason it does not transfer, and
what would have to be measured. Coverage is back at 100 with the private half's
refusal reached by a case that says why the copy is there.

One thing left, and it is small.

### Change requested

**`_sign_`'s `Raises` does not list the two `ValueError`s this change gave it.**
Its `Args` documents `pubkey` and its refusal, and `sign`'s `Raises` was updated
for both — but the private half still says only:

```text
ValueError: if the message hash is not 32 bytes, if aux_rand32 is
    given and is not 32 bytes, if aux_rand32 and grind are given
    together, or if the private key is not 32 bytes, does not fit
    in them, or is not in [1, n-1].
```

`_sign_` raises two more now: its own refusal of a `pubkey` beside
`verify=False`, four lines into the body, and `_checked`'s "the public key given
is not this private key's", which reaches a caller of `_sign_` exactly as it
reaches a caller of `sign`. The argument for adding them is the one
`test_the_private_half_refuses_that_contradiction_too` makes in its own
docstring: `_sign_` is the row the table recommends to a caller holding a parsed
point, not an internal detail, and the caller who reads that table is the one
who will read this docstring.

### Nothing else, and one thing to wait for

No further suggestions: the prose, the tests and the two files that carry the
design all say the same thing now, and the failure modes are documented where
they are raised.

Five cells of the matrix are still running on this head — the two
`Build wheels` on Linux, both dynamic-wheel suites, and the Windows arm sdist
install. Everything else is green, coverage and lint included. Once the
docstring is settled and those five have reported per cell rather than by
aggregate, this has my ACK.

fametrano:
## Fifth pass, at `52f913e` — the docstring, and a second issue

### `_sign_`'s `Raises`

Taken. Both new paths are listed, and the second is worded as `_checked`
is rather than as "a wrong argument":

```text
ValueError: if the message hash is not 32 bytes, if aux_rand32 is
    given and is not 32 bytes, if aux_rand32 and grind are given
    together, if the private key is not 32 bytes, does not fit
    in them, or is not in [1, n-1], if a `pubkey` is given beside
    `verify=False`, or if the signature verifies under the key
    this private key has and not under the one given -- the
    wrong key handed in, most likely, and `_checked` says what
    else it can be.
```

Your argument for it is the one that settles it: the caller who reads the
table is the caller who reads this docstring, and a public entry point
that hides two of its failures is not one.

### #247, for the `dsa.Signer` this PR leaves out

The paragraph declining it named a blocker and pointed at nothing, which
is the same defect #246 fixed. [It now
points here](https://github.com/btclib-org/btclib-secp256k1/issues/247):
`secp256k1_ecdsa_sign` takes a bare `seckey32` and the only thing that
reaches it is `scalar()`'s `bytes`, so a signer holding an
`ffi.new("unsigned char[32]")` would rebuild an unwipeable copy per
signature — worse than today, not better.

Writing it corrected my own framing. The README already measures what a
`dsa.Signer` could hoist — the argument check alone, 0.117 microseconds
of 12.93, against `ssa.Signer`'s 7.55 of 15.82 — and concludes the trade
"is worth making for half a signature and not for a hundredth of one".
So the issue does not ask for a measurement that exists. It says the
speed argument is closed against the signer, that what `scalar` blocks is
the *other* argument — where the key lives — and that the answer may
still be no, in which case what closes it is the reasoning and not code.

The issue also names what makes this more than a patch: `CData` is `Any`,
so widening `scalar`'s parameter to admit one widens it to `Any` and every
argument stops being checked — by the function whose whole job is
checking them.

---

`uv run pytest` — 647 passed, coverage 100. `uv run pre-commit run
--all-files` — every hook, no skips. Nothing but the docstring changed
since `c9ad984`, so the matrix is re-running on a tree that differs by
eight lines of prose.

fametrano:
## ACK — `52f913e`

Everything requested across the four passes is in, nothing is outstanding, and
the tree is the one the gates ran on. This is my formal ACK for `52f913e`.

**What the ACK rests on, read rather than assumed:**

- **28 checks, all pass; 27 jobs across the four workflows, none non-success** —
  no cell reporting green while dying in *Set up job*, which is what reading the
  aggregate alone would have hidden. `test: every job passed` is a conclusion
  here and not a substitute for the cells.
- **The three required contexts** — `Lint and type-check`, `Build the
  documentation`, `test: every job passed` — all success on this head, with
  coverage back at 100% and the private half's refusal reached by a case of its
  own.
- **`strict` is on and satisfied**: the branch sits on `93cdc12`, `main`'s tip,
  so what was tested is what would land.
- **One commit, no version bump**, and the CHANGELOG entry is under the open
  `v0.8.0.3` where a work-in-progress entry belongs.

**What the change is, in one paragraph.** `dsa.sign` and `dsa._sign_` take the
public key the caller already holds, so the check verifies under it instead of
deriving one — 20.20 microseconds down to 13.06 with an uncompressed key handed
in, of which 7.40 is the `secp256k1_ec_pubkey_create` no longer done. The key is
taken on trust, and the trust is paid back only where it fails: `ValueError`
where the signature verifies under the private key's own key and not under the
one given, `RuntimeError` where it verifies under neither, which is the fault
this check has always been for. Both directions of that discrimination are held
by a test, the second needing a substituted verification because
`raise RuntimeError` sits outside the coverage ratchet — the one hole a green
100% could not have reported. What the trust cannot do is accept a bad
signature, and that is a property over forty keys rather than a paragraph. What
it can do, and the derived check never could, is catch a private key that was
already corrupt when it was signed with.

**How it got here** is worth as much as what it says. The first pass found
numbers that did not reconcile with their own table, and the answer was a
re-measurement with the noise row the claim needed. The second found the
corroboration was of a different libsecp256k1 call, and the answer was to time
the right one and report the figure that *disagreed* alongside the one that
agreed. The claim about `submodule-pin` failing on `main` was withdrawn by the
person who made it, after checking the thing he had inferred. Three of those
were the difference between a measured repository and one that merely sounds
like one.

**Merging is the documented route**, not a button: one commit, so
`git push origin <sha>:refs/heads/main` after a last look at the head — the
`BLOCKED` state is the required approving review, which GitHub does not let an
author give themselves, and REPOSITORY.md carries the bypass that answers it.

`#246` and `#247` carry what this deliberately leaves out, each with the
measurement it would need rather than a sentence declining it.

Issue #251, opened 2026-08-18
Nothing asserts that dsa's check skips the derivation when a key is handed in


fametrano:
Noticed while writing the same argument for `recovery` (#246), and not
addressed there: the saving `dsa.sign(pubkey=...)` exists for is asserted
by nothing.

`btclib_secp256k1/dsa.py`'s `_checked` verifies under the key handed in
and returns, deriving only where that fails. Delete the early return and
the tests do catch it — a correct key then reaches the second
verification and `test_a_key_handed_in_is_the_key_that_would_have_been_derived`
gets a `ValueError`. What nothing catches is the saving being given up
while the behaviour stays identical:

```python
# still passes every test in tests/test_verified_signing.py
derived = keys._pubkey_from_prvkey_(prvkey_bytes)
if pubkey is not None:
    if _verify_(msg_bytes, pubkey, signature):
        return
    ...
```

That spelling answers the same signature for every input and pays the
`secp256k1_ec_pubkey_create` the argument exists to avoid — 7.4
microseconds of the 12.8 the check costs, per the table in
`README.md`. A performance contract nothing holds is one a refactor can
drop silently, and the figures in the README would then be false with no
test red.

`recovery`'s side of this is closed by
`test_the_derivation_the_argument_saves_is_not_made_at_all` in that same
file, which substitutes `keys._pubkey_from_prvkey_` for one that raises
and asserts the signature still comes back. The `dsa` case is two lines
of the same shape, and it belongs to #245's diff rather than to #246's,
which is why it is here rather than in that pull request.

PR #252, opened 2026-08-18
Compare the recovered key with the one the caller already holds

### LINKED ISSUE #246: recovery.sign derives the key it compares against, and takes no key from the caller
`dsa.sign` now takes the public key the caller already holds and checks
under it instead of deriving one (#245). `recovery.sign` has the same
derivation and does not take the same argument, and this is what it would
take to give it one.

## The saving is the same

`_abort_unless_recovered` compares the recovered key against a derived
one:

```python
keys._pubkey_cmp_(recovered, keys._pubkey_from_prvkey_(prvkey_bytes))
```

That right-hand side is `secp256k1_ec_pubkey_create`, the same call
`dsa`'s check was making, and a caller who holds the key already has no
reason to pay for it twice. The recovery stays — it is what this scheme
checks and a plain verification does not — and only the derivation goes.

## The discrimination is not the same, which is why it was not folded in

In `dsa` a failed check has two causes and the failing branch tells them
apart by deriving: the key handed in is not this private key's, or the
computation faulted. Here there is a third. A recovered key that does not
match the one handed in can also mean the recovery id is wrong, which is
a property of the signature rather than of either key, and `recovery` is
the one module in the package where that failure is reachable from an
input rather than needing a stand-in — `test_the_recovery_id_is_what_the_check_catches`
already exercises it.

So the branch cannot simply be the two-way one from `dsa`. What it should
say for each of the three cases is the design question, and it is not
answered by the change that landed in `dsa`.

## What the change needs

- a measurement of its own, in the shape of the table in #245: the
  unchecked signature as the anchor, the key derived, a compressed key
  handed in, an uncompressed one, and the anchor again for the noise —
  `recovery.sign`'s check is the dearest of the three, so what the
  argument buys there is a figure nothing yet has
- cases of its own for the three-way failure, including the one nothing
  else in the package needs: a right key, a wrong id
- the same refusals `dsa` grew — a key beside `verify=False`, and octets
  parsed at the boundary rather than inside the check

## Not obviously the same answer

It is worth deciding rather than assuming that `recovery` should take the
argument at all. It is the scheme whose check is doing the most work, so
the saving is largest; it is also the one where a wrong diagnosis has the
most ways to be wrong. Either outcome is a decision this repository can
record — what should not stand is the CHANGELOG's "left to the change
that gives it them" pointing at nothing.

Follows btclib-org/btclib-secp256k1#245 and the design in
btclib-org/btclib#982.

fametrano:
`recovery.sign` and `recovery._sign_` take a keyword-only `pubkey`, the key
`_abort_unless_recovered` compares the recovered one with, so the check no
longer derives a key the caller is holding. Everything around it is what
[ISS 245](https://github.com/btclib-org/btclib-secp256k1/issues/245)
settled for `dsa` and is not restated in a second shape: taken on trust,
refused beside `verify=False` in both halves, octets parsed at the
boundary.

**The decision [ISS 246](https://github.com/btclib-org/btclib-secp256k1/issues/246)
left open** — whether `recovery` should take the argument at all, the
saving being the largest and the diagnosis having the most ways to go
wrong — is taken, and the reason is that the third cause needs no third
message. A mismatch means the key given is wrong, the recovery id is not
the signature's, or the computation faulted; the derivation the matching
path skipped separates the first from the other two, and those two share
the `RuntimeError` because a caller of `sign` cannot pass an id at all —
it comes back from `secp256k1_ecdsa_sign_recoverable` beside the
signature, so a wrong one is a fault by the time the check runs. The cost
of the largest saving in the package is therefore one comparison, on a
path that was already failing.

**The measurement**, one session with `dsa.sign` beside it so the two are
comparable rather than merely printed together — Apple M5, macOS 26.6,
arm64, CPython 3.14.6, nine rounds of 3 000 calls, minimum kept, private
key handed in as octets in every row:

| `recovery.sign` | per call | the check |
| --- | --- | --- |
| `verify=False` | 12.07 | |
| the key derived, as before | 35.00 | 22.93 |
| a compressed key handed in | 29.67 | 17.60 |
| an uncompressed key handed in | 27.49 | 15.42 |
| the point already parsed, through `_sign_` | 27.25 | 15.18 |
| `verify=False` again (the noise) | 12.09 | ±0.02 |

`dsa.sign` there: 12.14 unchecked, 31.97 derived, 24.55 uncompressed in —
19.83 becoming 12.41, within a few tenths of ISS 245's own session. The
7.51 removed is `secp256k1_ec_pubkey_create` and nothing besides (7.53
timed alone in the same session); 2.18 is the field square root and 0.24
the uncompressed parse, agreeing with the 2.02 and 0.26 measured for
`dsa`. What stays is the ~3 µs a recovery and a comparison cost over a
verification, the same 3 before and after — which is what says the
argument removed the derivation and left the rest alone.

**Cases.** `dsa`'s group repeated for this module, plus two of its own:
`test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key` puts
all three causes in front of the real libsecp256k1 with no stand-in
anywhere (right key + right id passes, right key + wrong id is the
`RuntimeError`, wrong key + right id is the `ValueError`), and
`test_the_derivation_the_argument_saves_is_not_made_at_all` pins the
saving rather than the answer — `keys._pubkey_from_prvkey_` substituted
for a raise, the signature still returned. Verified by deleting the early
return, which fails both.

`README.md` carries the table and the three-cause paragraph; the sentence
that called the saving "available and not yet taken" for `recovery` is now
the one that says what taking it cost. `HISTORY.md` gains the release note
for `pubkey`, which covers `dsa` too — the notes describe the release
rather than the commit, and ISS 245 left that paragraph unwritten.

Evidence, local, in a worktree on this commit:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, 654 passed, 100% coverage
- `uvx pre-commit run --all-files` — exit 0, 37 hooks
- the saving deleted, `pytest -k` on the three cases — 2 failed, as intended; restored, full suite green again (`PYTHONDONTWRITEBYTECODE=1` throughout)

Noticed and filed, not addressed here: [ISS 251 — dsa's own saving is asserted by nothing](https://github.com/btclib-org/btclib-secp256k1/issues/251)

Closes #246

## Summary by Sourcery

Let recoverable signing compare against a caller-held public key to avoid redundant key derivation while preserving clear verification failures.

New Features:
- Allow `recovery.sign` and `recovery._sign_` callers to provide an existing public key for recoverable-signature verification.

Bug Fixes:
- Distinguish a mismatched caller-provided public key from signatures that recover the wrong key or otherwise indicate a signing failure.
- Reject invalid public-key inputs before signing and reject `pubkey` when verification is disabled.

Enhancements:
- Avoid deriving a public key during the successful verification path when the caller already provides one, while retaining fallback diagnosis for mismatches.

Documentation:
- Document the new recovery public-key parameter, its performance impact, error behavior, and rationale in the README, changelog, and history.

Tests:
- Add coverage for matching and mismatched keys, invalid inputs, disabled verification, wrong recovery IDs, no-derivation behavior, key trust handling, and recovery properties.

### DIFF
--- a/btclib_secp256k1/recovery.py
+++ b/btclib_secp256k1/recovery.py
@@ -34,7 +34,7 @@ _COMPACT_BUFFER_TYPE = ffi.typeof(f"char[{_COMPACT_SIZE}]")
 
 
 def _abort_unless_recovered(
-    signature: CData, msg_bytes: bytes, prvkey_bytes: bytes
+    signature: CData, msg_bytes: bytes, prvkey_bytes: bytes, pubkey: CData | None
 ) -> None:
     """Recover from a signature just made, and refuse another key.
 
@@ -65,13 +65,47 @@ def _abort_unless_recovered(
     it raises where a `_foo_` half would answer, and the verb is the one
     BIP340 uses of the step.
 
+    **A key handed in is compared with instead of derived**, which is
+    `dsa._checked`'s arrangement and buys the same
+    `secp256k1_ec_pubkey_create` here, at the same price it costs there:
+    what is dearest in this package is the check, not the saving. What it
+    costs is that a mismatch now has three causes where `dsa`'s has two:
+    the key given is not this private key's, the recovery id is not the
+    signature's, or
+    the computation faulted. The first is an argument and the other two
+    are not, and one comparison separates them -- the recovered key
+    against the derived one, paid only where something already went
+    wrong. Where they agree, the signature is the signer's and it is the
+    argument that is wrong; where they do not, the signature does not
+    recover its own signer, which is what a wrong id and a fault both
+    look like from here and neither is anything the caller passed.
+
+    That those two share an answer is the decision rather than an
+    oversight. A caller of `sign` cannot pass an id at all -- it comes
+    back from `secp256k1_ecdsa_sign_recoverable` beside the signature --
+    so an id that is not the signature's is a fault by the time this
+    runs, and `test_the_recovery_id_is_what_the_check_catches` reaches
+    that state the only way anything can: through the parsed signature
+    this private half takes.
+
     **Neither call below needs a `context.check` behind it, and both are
     the kind that usually would.** `keys._pubkey_cmp_` answers an
     ordering that means nothing where libsecp256k1 could not read an
-    object, and that answer is a security decision here -- but both
-    objects come from calls that returned success a line earlier, so the
-    only way its `0` could lie is both keys being unreadable at once,
-    which is not a state either call can leave behind. And
+    object, and that answer is a security decision here. One of the two
+    objects is no longer proved by a preceding success: `recovered` is,
+    but `pubkey` is what this half takes on trust, and a caller handing
+    in a `secp256k1_pubkey` nothing has written to reaches
+    `secp256k1_ec_pubkey_cmp` with it -- so an illegal argument *is*
+    recorded on the thread here, which the comparison against a derived
+    key could not do. What holds instead is the comparison itself:
+    libsecp256k1 serializes a key it cannot load as 33 zero octets,
+    documented there as "less than any valid public key", and the
+    recovered key is readable and serializes with an `0x02` or `0x03`
+    prefix. So the `0` still cannot lie -- an unreadable key compares
+    *unequal*, the fall-through derives, and what the caller is told is
+    that the key they gave is not this private key's, which is the truth
+    about it. `dsa._verify_` documents the same trust from the other
+    side, and `dsa._checked` passes such a key straight into it. And
     `keys._pubkey_from_prvkey_` raises a `ValueError` for a key outside
     [1, n-1], which this one is not: libsecp256k1 accepted it for signing
     immediately above. That is why `_recover_`'s `ValueError` is
@@ -84,13 +118,20 @@ def _abort_unless_recovered(
         signature: the recoverable signature just made.
         msg_bytes: the 32-byte hash it was made over, already checked.
         prvkey_bytes: the 32-byte private key that made it, already
-            checked.
+            checked, and what the failing branch derives from.
+        pubkey: the public key the recovered one is compared with,
+            parsed, or None to derive it -- which is what a caller
+            holding nothing gets.
 
     Raises:
         RuntimeError: if no key recovers from the signature, or if the
             one that does is not the signer's. Neither is reachable by
-            any input: what they report is the computation itself having
-            gone wrong.
+            any input to `sign`: what they report is the computation
+            itself having gone wrong, the recovery id included.
+        ValueError: if the recovered key is the signer's and not the one
+            handed in, which is the wrong key given -- or a private key
+            that was already corrupt when it was signed with, as
+            `dsa._checked` says of the same branch.
     """
     try:
         recovered = _recover_(msg_bytes, signature)
@@ -102,8 +143,14 @@ def _abort_unless_recovered(
         raise RuntimeError("signing produced a signature no key recovers from") from (
             failure
         )
+    if pubkey is not None and not keys._pubkey_cmp_(recovered, pubkey):
+        return
+    # either nothing was handed in, or what was did not match: the
+    # derivation the common path skipped is what says which
     if keys._pubkey_cmp_(recovered, keys._pubkey_from_prvkey_(prvkey_bytes)):
         raise RuntimeError("signing produced a signature that recovers another key")
+    if pubkey is not None:
+        raise ValueError("the public key given is not this private key's")
 
 
 def _sign_(
@@ -112,6 +159,7 @@ def _sign_(
     aux_rand32: BytesLike | None = None,
     *,
     verify: bool = True,
+    pubkey: CData | None = None,
 ) -> CData:
     """Create a recoverable ECDSA signature, as the parsed signature.
 
@@ -130,19 +178,32 @@ def _sign_(
         verify: whether to recover the key from the signature and refuse
             one that is not the signer's, as `sign` documents and
             `_abort_unless_recovered` reasons about.
+        pubkey: the public key the recovered one is compared with,
+            parsed, where the caller already holds it and would rather
+            not have it derived again. Taken on trust;
+            `_abort_unless_recovered` says what that does and does not
+            risk. Refused beside `verify=False`, which declines the
+            check it is for.
 
     Returns:
         The libsecp256k1 recoverable signature object.
 
     Raises:
         ValueError: if the message hash is not 32 bytes, if aux_rand32 is
-            given and is not 32 bytes, or if the private key is not 32
-            bytes, does not fit in them, or is not in [1, n-1].
+            given and is not 32 bytes, if the private key is not 32
+            bytes, does not fit in them, or is not in [1, n-1], if a
+            `pubkey` is given beside `verify=False`, or if the recovered
+            key is this private key's and not the one given -- the wrong
+            key handed in, most likely, and `_abort_unless_recovered`
+            says what else it can be.
         RuntimeError: if `verify` asks and the signature does not recover
             the key that made it.
     """
     prvkey_bytes = scalar(prvkey, "private key")
     msg_bytes = octets(msg_bytes, "message hash", 32)
+
+    if pubkey is not None and not verify:
+        raise ValueError("pubkey is for the check that verify=False declines")
 
     signature = ffi.new("secp256k1_ecdsa_recoverable_signature *")
 
@@ -159,7 +220,7 @@ def _sign_(
     ):
         raise ValueError("invalid private key: not in [1, n-1]")
     if verify:
-        _abort_unless_recovered(signature, msg_bytes, prvkey_bytes)
+        _abort_unless_recovered(signature, msg_bytes, prvkey_bytes, pubkey)
     return signature
 
 
@@ -169,6 +230,7 @@ def sign(
     aux_rand32: BytesLike | None = None,
     *,
     verify: bool = True,
+    pubkey: BytesLike | None = None,
 ) -> tuple[bytes, int]:
     """Create a recoverable ECDSA signature.
 
@@ -194,6 +256,19 @@ def sign(
     verification's work, and the comparison and the derivation are the
     rest. CHANGELOG.md names the session.
 
+    `pubkey` is how a caller already holding the key stops the comparison
+    deriving it, and what it removes is the same call `dsa.sign` removes,
+    at the same price: the check here is the dearest of the three, the
+    saving is not. In a session of its own: 35.00 microseconds with the
+    key derived, 29.67 with a compressed one handed in, 27.49 with an
+    uncompressed one, against 12.07 unchecked and a noise of
+    ±0.02. The 7.51 that removes is `secp256k1_ec_pubkey_create`, which
+    timed alone there is 7.53, and what stays is the recovery and the
+    comparison -- about 3 microseconds over what `dsa.sign` pays for a
+    verification, whether the key is handed in or derived. The README
+    carries both tables, and `_abort_unless_recovered` what the trust
+    costs in diagnosis.
+
     Args:
         msg_bytes: the 32-byte hash of the message.
         prvkey: the private key, 32 bytes or an int below 2**256.
@@ -203,6 +278,14 @@ def sign(
             does not give the signer's back. It costs a recovery, a point
             multiplication and a comparison, and False is what a caller
             that has measured those against its own threat model passes.
+        pubkey: the public key of this private key, where the caller
+            already holds it, so that the comparison does not derive it
+            again. That derivation is the multiplication named above, and
+            this is how not to pay it twice. It is taken on trust rather
+            than checked against the private key, which would cost the
+            multiplication it saves; where it is wrong, the check says so
+            and says which of the three things went wrong. Refused
+            beside `verify=False`, which declines the check it is for.
 
     Returns:
         The 64-byte compact signature and its recovery id. The id is 0
@@ -212,8 +295,11 @@ def sign(
 
     Raises:
         ValueError: if the message hash is not 32 bytes, if aux_rand32 is
-            given and is not 32 bytes, or if the private key is not 32
-            bytes, does not fit in them, or is not in [1, n-1].
+            given and is not 32 bytes, if the private key is not 32
+            bytes, does not fit in them, or is not in [1, n-1], if pubkey
+            is not a public key or is given beside `verify=False`, or if
+            pubkey is not this private key's -- which is told apart from
+            the failure below rather than reported as it.
         RuntimeError: if libsecp256k1 fails to serialize the signature,
             which no input can make it do, or if `verify` asks and the
             signature does not recover the key that made it.
@@ -227,7 +313,26 @@ def sign(
         >>> pubkey == keys.pubkey_from_prvkey(1)
         True
     """
-    return serialize_compact(_sign_(msg_bytes, prvkey, aux_rand32, verify=verify))
+    # the contradiction before the value, as `dsa.sign` answers it and
+    # for the same reason: a caller who asked for no check and handed in
+    # a key to check with should hear about the two arguments rather than
+    # about the octets of one of them
+    if pubkey is not None and not verify:
+        raise ValueError("pubkey is for the check that verify=False declines")
+
+    return serialize_compact(
+        _sign_(
+            msg_bytes,
+            prvkey,
+            aux_rand32,
+            verify=verify,
+            # parsed here rather than inside the check, so that octets
+            # which are not a public key are refused before anything is
+            # signed: a caller who mistyped an argument should not be
+            # told about it by a check on a signature they now hold
+            pubkey=None if pubkey is None else keys.parse(pubkey),
+        )
+    )
 
 
 def _recover_(msg_bytes: BytesLike, signature: CData) -> CData:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,101 @@ release-notes length in the first place, and are still in
 
 ## v0.8.0.3 (work in progress, not released yet)
 
+### The recoverable check takes that key too
+
+- **`recovery.sign` and `recovery._sign_` take a keyword-only `pubkey`**
+  (#246), the key the recovered one is compared with, so that
+  `_abort_unless_recovered` no longer derives a key the caller is
+  holding. Everything around it is what #245 settled for `dsa` and is
+  not restated in a second shape: the key is taken on trust, it is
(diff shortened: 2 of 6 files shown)


fametrano:
READY 8bd7c081ccc97cb9eabce35677b666c54c12513c

`recovery.sign` and `recovery._sign_` take the public key the caller
already holds, so the check compares the recovered key with it instead of
deriving one — and the question
[ISS 246](https://github.com/btclib-org/btclib-secp256k1/issues/246) left
open is answered rather than assumed: `recovery` does take the argument,
because the third cause a mismatch gains (the recovery id) needs no third
message — it is unreachable from any argument of `sign`, so by the time
the check runs it is the fault the `RuntimeError` already names.

Evidence, local:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, 654 passed, 100% coverage
- the early return in `_abort_unless_recovered` deleted, then `pytest -k` on the three cases — 2 failed (the wrong-id case and the saving case), restored, full suite green again; `PYTHONDONTWRITEBYTECODE=1` throughout, per CLAUDE.md
- the measurement in the body: one session, `dsa.sign` beside it, nine rounds of 3 000 calls, minimum kept, noise ±0.02

[ISS 251 — dsa's own saving is asserted by nothing](https://github.com/btclib-org/btclib-secp256k1/issues/251).

btclib_secp256k1/recovery.py
```diff
@@ -102,8 +131,14 @@ def _abort_unless_recovered(
         raise RuntimeError("signing produced a signature no key recovers from") from (
             failure
         )
+    if pubkey is not None and not keys._pubkey_cmp_(recovered, pubkey):
```
fametrano: **blocking** — the paragraph at `recovery.py:90-97` is not true of this call, and it is the paragraph that justifies leaving `context.check` out.

It reads: *"both objects come from calls that returned success a line earlier, so the only way its `0` could lie is both keys being unreadable at once, which is not a state either call can leave behind."* That held while the only comparison was `recovered` against a key `keys._pubkey_from_prvkey_` had just returned. Here the second object is `pubkey`, which `_sign_` takes on trust: `sign` parses it, but a caller of the private half hands in a `CData` nothing in this package validated — which is the whole point of `_sign_` taking a parsed point.

How I know, on 8bd7c081, in a worktree at that sha:

```python
unreadable = ffi.new("secp256k1_pubkey *")      # zeroed, which is what a caller can hand in
recovery._sign_(msg, prv, pubkey=unreadable)
# ValueError: the public key given is not this private key's
context.check()
# ValueError: libsecp256k1 illegal argument: !secp256k1_fe_is_zero(&ge->x)
```

So the state the paragraph says neither call can leave behind is left behind by this one, reached through the private half's own documented contract rather than through a stand-in.

**The safety conclusion survives, and that is what the sentence should be saying instead.** `recovered` is readable, `secp256k1_ec_pubkey_cmp` substitutes 33 zero octets for a key it cannot load, and a readable key serializes with an `0x02`/`0x03` prefix — so the `0` still cannot lie, and the fall-through reaches the derivation and answers `ValueError`, which the probe above confirms. What has gone is the *reason given*: one of the two objects is no longer proved by a preceding success, so "both objects come from calls that returned success" has to become the argument that a readable key never compares equal to zeros.

Worth noting which way the package leans elsewhere: `dsa._verify_` documents the opposite for its own trusted argument — "False where libsecp256k1 could not read one of the two objects ... `context.check` immediately after the call is what says the False is not a verdict" — and `dsa._checked` passes a trusted `pubkey` straight into it. `dsa._sign_(msg, prv, pubkey=unreadable)` gives the same pair of answers as the probe above, so the behaviour is consistent between the two modules and only this docstring claims the case cannot arise.

CLAUDE.md: "a change that makes a comment untrue updates the comment, in the workflows and the build scripts as much as in the package".

btclib_secp256k1/recovery.py
```diff
@@ -65,6 +65,28 @@ def _abort_unless_recovered(
     it raises where a `_foo_` half would answer, and the verb is the one
     BIP340 uses of the step.
 
+    **A key handed in is compared with instead of derived**, which is
+    `dsa._checked`'s arrangement and buys the same
+    `secp256k1_ec_pubkey_create` here -- the larger saving of the two,
```
fametrano: **non-blocking** — "the larger saving of the two" is a distinction this session's own numbers do not draw. Same claim at `CHANGELOG.md:36` ("the largest of the three"), `CHANGELOG.md:86` ("the largest saving in the package"), `README.md:694` and `HISTORY.md:123`.

Both savings are `secp256k1_ec_pubkey_create`, as the diff itself says two sentences later. From the tables in the body and in CHANGELOG.md, all in the one session:

- `recovery`: 35.00 − 27.49 = **7.51**
- `dsa`: 31.97 − 24.55 = **7.42**
- the call timed alone there: **7.53**

0.09 apart, where the same paragraph calls 2.18 and 2.02 "agreeing", and warns in the next clause that a comparison of this size is "between sessions and not within either". Both land on the 7.53 the call costs by itself, which is the reading that matches "and nothing besides".

What the measurement does carry is that the *check* here is the dearest — 22.93 against `dsa`'s 19.83 — not that the saving is larger. ISS 246 guessed the other way before anything was measured ("It is the scheme whose check is doing the most work, so the saving is largest"); the numbers this pull request produced are what settles that guess, and they say the derivation removed is the same call, at the same price, in both modules.

Not blocking, and the fix is a word: the decomposition immediately below already states the true fact.

fametrano:
Reviewed **8bd7c081** in a read-only worktree of my own at that sha — one commit on `5a8248b`, signed `G Ferdinando M. Ametrano`, `origin/main` an ancestor.

**Gates, run locally on that sha, exit codes read rather than output filtered:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 654 passed, 100.00% (834 statements, 190 branches, none missed)
- `uvx pre-commit run --all-files` — **exit 0**, 37 hooks Passed, nothing modified

`PYTHONDONTWRITEBYTECODE=1` throughout, and a fresh venv and submodule of its own, so nothing here read the author's tree.

**What I checked beyond the gates**

- the `pubkey=None` path is the old behaviour statement for statement: the new first `if` is false, the derived comparison runs as before, and the trailing `if pubkey is not None` is false. Nothing an existing caller can reach has moved
- the three-way discrimination, by hand rather than by reading: a right key with a wrong id gives the `RuntimeError`, a stranger's key with the right id gives the `ValueError`, and `test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key` is that pair in both directions with no stand-in — which is the case ISS 246 said no other module could write, and it is written
- every figure in the three tables reconciles against the anchor and against each other: 22.93/17.60/15.42/15.18 are 35.00/29.67/27.49/27.25 less 12.07; 7.51 is 35.00 − 27.49 and lands on the 7.53 of the call timed alone; 2.18 is 29.67 − 27.49 and 0.24 is 27.49 − 27.25; `dsa`'s 19.83 and 12.41 are 31.97 and 24.55 less 12.14, and the ~3 µs "the same before and after" is 22.93 − 19.83 = 3.10 against 15.42 − 12.41 = 3.01. No number in the prose is without a row
- the scope: recovery.py, its tests, and the three documents. The HISTORY.md paragraph covers `dsa` as well, which reads as in scope rather than beside it — HISTORY describes the release, and this is the change that finishes the argument it describes
- the `verify=False` refusal exists in both halves and in the same position `dsa` puts it, and the boundary parse happens before anything is signed because it is an argument to the `_sign_` call

**Blocking (1)**

1. `btclib_secp256k1/recovery.py:134` — the docstring paragraph that justifies leaving `context.check` out of these calls is made untrue by the comparison this diff adds. Its premise is that both objects come from calls that returned success a line earlier; the handed-in key does not, and one probe through the private half leaves libsecp256k1's `!secp256k1_fe_is_zero(&ge->x)` recorded on the thread — the state the paragraph says neither call can leave behind. The safety conclusion survives on a different argument, which is what the sentence should be saying; the comment has the details and the probe.

**Non-blocking (1), left to you**

2. `btclib_secp256k1/recovery.py:70`, and the same claim at `CHANGELOG.md:36`, `CHANGELOG.md:86`, `README.md:694`, `HISTORY.md:123` — "the largest of the three savings" is 7.51 against 7.42 in one session, both being `secp256k1_ec_pubkey_create` and both landing on the 7.53 it costs alone. What this session measured as largest is the *check*, not the saving.

**Nit, once**

`HISTORY.md:124-126` sets `dsa.sign`'s 31.92/24.78 beside `recovery.sign`'s 35.00/27.49 in one comparative sentence. Those are two different sessions; 31.97/24.55 exist in the new one precisely so the pair is comparable, and the surrounding prose is otherwise careful about exactly this.

No issues filed: everything I found is inside this diff. ISS 251, which the handoff names, is where it belongs.

CHANGES REQUESTED 8bd7c081ccc97cb9eabce35677b666c54c12513c

btclib_secp256k1/recovery.py
fametrano: Accepted, in `a35c281`. Reproduced your probe first rather than reading it: `recovery._sign_(msg, 7, pubkey=ffi.new("secp256k1_pubkey *"))` raises the `ValueError` about the key, and `context.check()` immediately after reports `libsecp256k1 illegal argument: !secp256k1_fe_is_zero(&ge->x)`. And the upstream reason is in the vendored tree, `secp256k1/src/secp256k1.c:319-332`: on a serialization failure `secp256k1_ec_pubkey_cmp` memsets that key's 33 octets to zero, under a comment saying it is "less than any valid public key" so that comparisons stay consistent.

So the paragraph now says what holds. The premise that both objects come from a preceding success is gone, and it is replaced by the comparison's own property: an unreadable key is 33 zeros, a readable one carries an `0x02`/`0x03` prefix, so the `0` cannot lie — an unreadable key compares *unequal*, the fall-through derives, and the caller is told about the key they gave rather than about their hardware. It also names the record as something this call *does* leave, which is the fact the old sentence denied, and points at `dsa._verify_` documenting the same trust from the other side.

Pinned rather than asserted, since the reasoning is now the only thing standing between this and a wrong diagnosis: `test_the_key_the_recoverable_check_trusts_is_the_one_nobody_validated` in `tests/test_callbacks.py` drives your probe and holds both halves — the `ValueError` from the wrapper and the illegal argument from `context.check` after it. That module's docstring gains the sentence that makes it a listed shape, this being the one whose untrusted object is a wrapper's own argument rather than something a caller handed `lib`.

btclib_secp256k1/recovery.py
fametrano: Accepted, in `a35c281` — you are right, and the arithmetic is the argument: 7.51 against 7.42 is 0.09, in prose that calls 2.18 and 2.02 "agreeing" and warns in the next clause that a difference of that size is between sessions. Both land on the 7.53 the call costs timed alone, which is what "and nothing besides" already said.

Corrected in all five places, and the claim now made is the one the numbers carry — the *check* is the dearest of the three, 22.93 against 19.83, and the saving is one `secp256k1_ec_pubkey_create` at one price in both modules:

- `recovery.py:70` — "buys the same `secp256k1_ec_pubkey_create` here, at the same price it costs there: what is dearest in this package is the check, not the saving"
- `recovery.py`, `sign`'s docstring — "what it removes is the same call `dsa.sign` removes, at the same price: the check here is the dearest of the three, the saving is not"
- `CHANGELOG.md` — the bullet heading is now "The dearest check of the three, and the same saving as `dsa`'s", and the closing sentence drops "largest"
- `README.md:694` — the same, with the pointer to the table
- `HISTORY.md:123` — "one call at one price in both modules — what differs is the check around it"

And since ISS 246 guessed the other way in writing, the CHANGELOG now records that this session settled it: the guess was that the largest check implied the largest saving, and the measurement says the derivation removed is the same call in both.

fametrano:
READY a35c281ef929b9df0aaca0f8b4f056ceeddff258

Round 2, on `8bd7c08` — a second commit rather than an amend, so `8bd7c08..a35c281` is exactly what changed; the two get squashed at landing, per CLAUDE.md.

All three taken, and each answered in its thread:

1. **blocking, `recovery.py:134`** — accepted. Your probe reproduced here, and the upstream reason read out of the vendored tree (`secp256k1/src/secp256k1.c:319-332`: an unserializable key is memset to 33 zeros, "less than any valid public key"). The paragraph now says the record *is* left here and rests the safety on the comparison instead — an unreadable key compares unequal, so the fall-through derives and the caller hears about their argument. Pinned by a new test in `tests/test_callbacks.py`, that module's docstring gaining the sentence that makes it a listed shape.
2. **non-blocking, "largest of the three"** — accepted, corrected in all five places. The check is the dearest, 22.93 against 19.83; the saving is one `secp256k1_ec_pubkey_create` at one price, 7.51 against 7.42 against 7.53 alone. The CHANGELOG now also records that this settles the guess ISS 246 made in writing.
3. **nit, `HISTORY.md`** — accepted. Both pairs are now the new session's, 31.97/24.55 beside 35.00/27.49, so the comparative sentence compares within one session as the surrounding prose does.

Nothing declined.

Evidence, local, on this sha:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, **655 passed**, 100.00% (834 statements, 190 branches)
- `uvx pre-commit run --all-files` — exit 0, 37 hooks, nothing modified (it caught one 109-column line in `HISTORY.md` on the way, rewrapped before the commit)
- your probe, by hand: `ValueError: the public key given is not this private key's`, then `context.check()` → `ValueError: libsecp256k1 illegal argument: !secp256k1_fe_is_zero(&ge->x)`

`PYTHONDONTWRITEBYTECODE=1` throughout.

fametrano:
Round 2 reviewed: the delta `8bd7c081..a35c281e`, one commit on the sha I reviewed, signed `G Ferdinando M. Ametrano`. Both threads **resolved** — both findings addressed, neither declined, and the nit taken as well.

**Gates, in a fresh worktree of my own at a35c281:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 655 passed, 100.00% (834 statements, 190 branches, none missed)
- `uvx pre-commit run --all-files` — **exit 0**, 37 hooks Passed, `git status` clean afterwards

**The blocking one, checked rather than taken on your word**

- the paragraph now names the record as something this call *does* leave, which is the fact the old sentence denied, and rests the safety on the comparison instead. That reason is the right one and it is upstream's: `secp256k1/src/secp256k1.c:319-332` in the vendored tree says "we will serialize the key as all zeros which is less than any valid public key", so the quotation is verbatim and the property is libsecp256k1's own rather than an inference from a probe
- `test_the_key_the_recoverable_check_trusts_is_the_one_nobody_validated` pins both halves, and it is the right file for it: every other shape in `test_callbacks.py` is an object a caller passed through `lib`, and the module docstring now says which one this is instead. `match="secp256k1_fe_is_zero"` is a libsecp256k1 assertion string, which is what that file already holds itself to elsewhere (`match="pubkey != NULL"`), so it is consistent rather than newly fragile
- the stale record left on the thread needs nothing further: `context.check` already documents that a message may be a wrapper's rather than the caller's, which is exactly this

**The non-blocking one**: corrected in all five places, and the CHANGELOG now records that the measurement settled ISS 246's guess rather than quietly dropping it. 0.16 is indeed what that paragraph calls "agreeing", so the argument is stated in the document's own terms. HISTORY.md now carries the same-session pair, 31.97 and 24.55, which was the nit.

**Non-blocking, left to you** — reflow, in text this round added. Four places where a line stops early or a clause did not survive editing:

- `recovery.py:71-72` — "What / it costs is that a / mismatch now has three causes"
- `recovery.py:262` — "the saving is not. In a session of its own: 35.00 microseconds"
- `HISTORY.md:127-128` — "the very multiplication the argument saves — so a / key that is not this private key's"
- `README.md:695-696` — "though the saving is not larger for it, being the same call at the same price, which the table below is where to read". That last clause is the one I would not leave: it is not a sentence, where the other three are only ragged

None of that holds the ack. If another commit goes on top before landing it is worth folding in; if not, it is a line in a follow-up.

Nothing else in the delta: the `pubkey=None` path is still untouched, no figure moved, and the three-way discrimination is the same code I checked at 8bd7c081.

ACK a35c281ef929b9df0aaca0f8b4f056ceeddff258

fametrano:
READY 31c4eb1637908d2387de6b3c42e3e7a341939438

Your non-blocking reflows, folded in rather than left to a follow-up: the README one was not a sentence, and landing it to fix it next would put the broken clause in `main` for the sake of saving a round. So this asks for the ack again on the head that will actually land — `a35c281`'s ack is not this head's, and the tree that CI passes has to be the tree that lands.

`a35c281..31c4eb1` is prose only, and no claim or figure moves:

- `recovery.py:70-73` and `:260-263` — two paragraphs rewrapped where a replacement kept the old breaks
- `HISTORY.md:127-128` — the same
- `README.md:693-698` — "which the table below is where to read" becomes two sentences: the check is the dearest of the three, the saving is not larger for that being the same call at the same price, and the table is where both are read

Three commits on the branch now, squashed into one at landing.

Evidence, local, on this sha:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, 655 passed, 100.00%

fametrano:
Round 3 reviewed: `a35c281e..31c4eb16`, one commit, signed `G Ferdinando M. Ametrano`. Taking the reflow rather than landing it and fixing it after was the right call — the README clause would have been in `main` to save a round, and a round is cheaper than that.

**That the delta moves no claim and no figure, checked rather than read.** `git diff -w` cannot show this, a rewrap moving words across lines rather than changing whitespace within one, so I compared the word streams of the three files at the two shas with every run of whitespace collapsed:

```text
HISTORY.md                    7231 -> 7231 words;  0 word-level changes
btclib_secp256k1/recovery.py  3331 -> 3331 words;  0 word-level changes
README.md                    10415 -> 10414 words; 13 word-level changes
```

So two of the three files are pure rewrap, to the word, and every changed token in the third is inside the one sentence: "three — though ... for it, ... which the table below is where to read" becomes "three. The saving is not larger for that, ... and the table below is where both are read". It is a sentence now, and "both" is the pair the paragraph is contrasting. `recovery.py`'s two hunks are inside docstrings, and their text is identical to the word — the doctests in `sign` among them.

**Gates, in a fresh worktree of my own at 31c4eb16:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 655 passed, 100.00% (834 statements, 190 branches, none missed)
- `uvx pre-commit run --all-files` — **exit 0**, 37 hooks Passed, `git status` clean afterwards

Nothing open: both threads were resolved at a35c281 and this head reopens none of them. No findings of any severity on this delta.

ACK 31c4eb1637908d2387de6b3c42e3e7a341939438

PR #253, opened 2026-08-18
Let a scalar be 32 octets the caller can overwrite

### LINKED ISSUE #247: _scalar.scalar() coerces every private key to bytes, so a dsa.Signer cannot wipe one
`ssa.Signer` holds a `secp256k1_keypair` — memory this package allocated,
owns, and can overwrite — so `wipe()` and `__exit__` mean what they say.
btclib-org/btclib#982 asks for a `dsa.Signer` beside it, and it cannot
make the same promise, because `_scalar.scalar()` coerces every path to
an immutable `bytes` before the key reaches libsecp256k1.

## Why the two schemes differ

`secp256k1_schnorrsig_sign32` takes a keypair, which is a struct this
package builds through `_secret.keypair` and hands to `_secret.wipe`.
`secp256k1_ecdsa_sign` takes a bare `const unsigned char *seckey32`, and
the only thing that reaches it is the return value of

```python
prvkey_bytes = scalar(prvkey, "private key")
```

whose annotation is `-> bytes` and whose fast path is `type(num) is
bytes`. So a `dsa.Signer` holding an `ffi.new("unsigned char[32]")` would
have to turn it back into a `bytes` on every call: a fresh immutable copy
of the secret per signature, none of them wipeable, which is worse than
the state today rather than better. Offering `wipe()` on top of that
would promise what it cannot keep.

## What is being asked for

That `scalar()` accept a cffi `CData` of exactly 32 octets and pass it
through unconverted, so that a caller who owns the buffer keeps owning
it. Everything else about the function stays: an `int` is still
serialized, `bytes`/`bytearray`/`memoryview` still go to `octets` and are
still *copied* there for the reason that docstring gives, and the length
is still checked before a bare pointer reads it.

The precedent is in the tree. `keys.py` already spells the round trip out
by hand where it needs owned memory:

```python
prvkey_buffer = ffi.new("char[32]", scalar(prvkey, "private key"))
```

and `silentpayments.py` does the same. Those allocate *because* `scalar`
would not give them a buffer — a `CData` arriving already 32 octets long
is the case none of them can express.

## What has to be decided rather than assumed

- **The type.** `BytesLike` is `bytes | bytearray | memoryview` and
  `CData` is `Any`, so widening `scalar`'s parameter to include `CData`
  widens it to `Any` and every argument stops being checked at all. That
  is the real cost of this change and the reason it is an issue rather
  than a patch: what a caller passes is checked by `scalar` precisely
  because the annotation cannot do it.
- **How a `CData` is recognised at run time.** `ffi.typeof` answers, and
  the fast path is currently one `type(...) is bytes`. Whatever is added
  must not slow the common path, which every entry point pays at least
  once and several pay three times — the docstring records 0.034 against
  0.080 microseconds for exactly that trade.
- **The length.** `ffi.sizeof` on an `unsigned char[32]` is 32, but on a
  pointer it is 8 — the same trap `_secret.wipe` documents, where
  `ffi.sizeof` would have wiped a quarter of a private key and reported
  success. A `CData` whose length cannot be trusted must be refused, not
  read.
- **Whether the copy is skipped or taken.** `octets` copies deliberately,
  so that a caller overwriting their own buffer cannot change what
  libsecp256k1 is about to read. Passing a `CData` straight through
  gives that up on purpose: the buffer belongs to this package in the
  case this is for. That distinction wants to be stated where the
  copying one is.
- **Whether `dsa.Signer` is worth it at all**, and the README has
  already answered the performance half of that: what such a signer
  could hoist is the argument check alone — 0.117 microseconds of the
  12.93 a signature costs, against the 7.55 of 15.82 that `ssa.Signer`
  hoists — because ECDSA derives no point to keep. Its verdict is that
  the trade "is worth making for half a signature and not for a
  hundredth of one".

  So a `dsa.Signer` cannot be argued for on speed, and this change is
  what would let it be argued for on the other ground: a private key
  that lives in memory this package can overwrite, rather than in a
  `bytes` that stays until the collector gets to it. That is a different
  claim from `ssa.Signer`'s and should be made in its own words —
  including the possibility that the answer is still no, in which case
  what closes this issue is the reasoning rather than the code.

## What this does not ask for

Not an opaque handle: #87 declined one and the reasoning holds. This is
about a buffer the caller already has, crossing one function that
currently refuses it.

Blocks the `dsa.Signer` half of btclib-org/btclib#982, which
btclib-org/btclib-secp256k1#245 left out for this reason.

fametrano:
Wherever a private key or a tweak is accepted, a cffi array of exactly 32
octets is now accepted too and handed to libsecp256k1 unconverted — the
only argument of these bindings that is not copied on the way in. The
reason is what the copy is: an immutable `bytes` of a secret, which
nothing can overwrite and which stays until the collector gets to it.

**[ISS 247](https://github.com/btclib-org/btclib-secp256k1/issues/247)'s
premise turned out to be wrong, and the finding narrowed the change.** It
reasoned that a `dsa.Signer` holding a buffer "would have to turn it back
into a `bytes` on every call". It would not: `dsa._signed` and
`dsa._checked` hand libsecp256k1 the pointer, exactly as `ssa.Signer`
bypasses `keypair` per signature. A prototype signer holding
`ffi.new("unsigned char[32]")` produced byte-identical signatures
*before* this change and hoisted 0.37 µs of 23.42 — Apple M5, macOS 26.6,
arm64, CPython 3.14.6, nine rounds of 3 000 calls, minimum kept, the
checked call run twice for a noise of 0.02. So the README's verdict on
speed stands untouched.

**What does bind is the derivation, and the sharpest case is a
diagnosis.** `keys._pubkey_from_prvkey_` asks for a scalar, so the failing
branch of `dsa._checked` — which derives precisely in order to tell a
wrong argument from a fault — answered `TypeError: the private key must be
bytes or an int, not __CDataOwn` under a buffer-held key. A caller whose
hardware had faulted was told they had mistyped an argument they passed
correctly, which is the misdiagnosis
[ISS 245](https://github.com/btclib-org/btclib-secp256k1/issues/245) built
that branch to avoid. That is why the door is open — for the diagnosis,
not for the microsecond — and
`test_the_discrimination_holds_for_a_key_held_in_a_buffer` holds both
sides of it with the key in a buffer.

**Three refusals**, because what follows is a bare pointer libsecp256k1
reads 32 octets from: a *pointer* (`ffi.sizeof` 8, the trap `_secret.wipe`
records from the other side), an array of wider items (`uint32_t[8]` is 32
octets of this machine's byte order, refused as `octets` refuses a
`memoryview` of wider items), and the wrong length.

**A `str` is refused before `ffi.typeof` is asked, and that is a bug the
suite caught rather than a case I thought of.** `ffi.typeof` reads a str as
a *cdecl*: `"x" * 32` raises cffi's own `error: undefined type name` —
neither this function's `TypeError` nor about the argument — and
`"char[32]"` is worse, resolving to a 32-octet type, so the str would have
been handed to libsecp256k1. `test_type_checks_refuse_what_merely_has_a_length`
failed on the first spelling the moment the branch existed; the second was
found looking for why. Both are cases now.

**The public annotations do not widen**, which was the cost ISS 247 named:
`CData` is `Any`, so `BytesLike | int | CData` on `dsa.sign` would stop
mypy checking every caller's private key. Only `scalar`'s own parameter
widens — the entry points still declare `BytesLike | int`, a `float` is
still refused statically at the call the caller made, and a cdata still
arrives because `ffi.new` answers `Any`. What is given up is mypy checking
this package's own calls of `scalar`, whose run-time check does the work.

The cffi stub gains the three ctype fields the question reads (`kind`,
`cname`, `item`) and `typeof` now takes a cdata as `sizeof` already does;
`SECURITY.md`'s claim that the entry side "cannot be improved from inside
this package" is no longer true and says what it now is.

Whether a `dsa.Signer` is worth having stays btclib-org/btclib#982's
question. This answers it only by removing the blocker: the wiping ground
it would be argued on no longer needs a signer at all.

Evidence, local, on this sha:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, 650 passed, 100% coverage
- `uvx pre-commit run --all-files` — exit 0, 37 hooks
- by hand: the buffer passes through as the same object, signs and derives identically, wipes to zero; the pointer/struct/wider-item/short shapes and both str spellings all refused with their own messages; the fault path answers `RuntimeError` and the wrong key `ValueError`

Closes #247

## Summary by Sourcery

Allow callers to pass wipeable cffi-backed 32-byte scalars while preserving copying for operations that modify or erase scalar memory.

New Features:
- Accept caller-owned cffi arrays of exactly 32 octets for private keys and tweaks without copying them at read-only scalar boundaries.

Bug Fixes:
- Preserve correct key-validation and hardware-fault diagnostics when private keys are held in cffi buffers.
- Reject invalid cffi scalar shapes and string inputs with clear type and length errors.
- Ensure operations that mutate or wipe scalar data continue copying into package-owned buffers.

Enhancements:
- Add safe scalar buffer handling and broaden cffi stubs for inspecting and re-viewing cdata.
- Document caller responsibilities, memory-lifetime considerations, and wiping guidance for caller-owned scalar buffers.

Documentation:
- Update the README, security guidance, and changelog to describe caller-owned scalar buffers and their security trade-offs.

Tests:
- Add coverage for cffi-backed scalars across scalar entry points, validation failures, parity-dependent paths, diagnostics, and wiping behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,6 +570,48 @@ the paragraph above describes, a second copy of the secret held for as
 long as the signer is. That trade is worth making for half a signature
 and not for a hundredth of one.
 
+**The other ground such a signer could be argued on is the memory, and
+it needs no signer.** Where a scalar is accepted — a private key, a tweak
+— a cffi array of 32 octets is accepted too, and where the call only
+*reads* it libsecp256k1 is handed the caller's own memory. So a caller
+signing again and again under one key can hold it in
+`ffi.new("unsigned char[32]")`, wipe that when done, and have no `bytes`
+of the key made per signature. It is the only argument of these bindings
+that is not copied on the way in, and the trade is stated where the
+copying one is: the copy `bytes` and `bytearray` get is also what stops
+the caller changing the octets libsecp256k1 is reading, and a caller who
+hands in memory instead of a value has taken that on.
+[SECURITY.md](https://github.com/btclib-org/btclib-secp256k1/blob/main/SECURITY.md)
+carries it with the wipe that goes with it.
+
+Which item type the array was declared as does not matter — `char`,
+`unsigned char`, `uint8_t`, `signed char` — because what crosses is a
+re-view of those octets rather than a conversion of them. What is refused
+is a pointer, whose length is not the pointer's own to know, an array of
+wider items, which is this machine's byte order rather than a scalar, and
+any length but 32.
+
+**Four calls copy it, and are meant to.** `keys.prvkey_negate`,
+`keys.prvkey_tweak_add`, `keys.prvkey_tweak_mul` and the sender side of
+`silentpayments` each own the buffer libsecp256k1 works in: the first
+three because it writes the answer through that pointer, and the last
+because this package wipes it afterwards. Handing those the caller's
+memory would negate or zero the key they passed, so a copy is owed and
+`_secret.scalar_buffer` takes it. Each of them *answers* a new secret,
+which is the other facility's question rather than this one's — `into`
+is how that comes back into a buffer instead of a `bytes`.
+
+Where that binds is not the signing, which never asked: the private
+halves hand libsecp256k1 the pointer, so a key in a buffer reached
+`_signed` before this and reaches it now. It binds on the *derivation* —
+`keys._pubkey_from_prvkey_` asks for a scalar — and the sharpest case is
+the failing branch of the check above, which derives in order to tell a
+wrong argument from a fault: with the key in a buffer it used to answer
+`TypeError: the private key must be bytes or an int`, telling a caller
+they had mistyped the argument they had passed correctly. Which is the
+reason the door is open, rather than the microsecond a signer would have
+hoisted.
+
 What does cost in `dsa.sign` is the DER serialization, 0.757
 microseconds, and a caller who wanted the other form never has to pay it:
 `compact=True` answers the 64 octets of `r ‖ s` directly, where reaching
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -103,13 +103,9 @@ These are known and inherent, not vulnerabilities:
     zeroed, which is the limitation above and not this narrowing of it.
     What the caller then does with the buffer is theirs: this does not
     wipe it for them, and a buffer they never overwrite is exactly the
-    un-zeroizable copy `into` was reached for to avoid. Nor does it
-    touch the *entry* side, which is the half that cannot be improved
-    from inside this package: the `bytes` or `int` handed in already
-    existed before the call, and the arithmetic that produced it
-    happened where this package cannot see. The obligation is stated
-    rather than enforced, for the reason `ssa.Signer` has no finalizer:
-    a guarantee nothing keeps is worse than none
+    un-zeroizable copy `into` was reached for to avoid. The obligation is
+    stated rather than enforced, for the reason `ssa.Signer` has no
+    finalizer: a guarantee nothing keeps is worse than none
 
     ```python
     prvkey = bytearray(32)
@@ -149,6 +145,53 @@ These are known and inherent, not vulnerabilities:
     type, a length or a magnitude and never on the content of a secret.
     Everywhere an `int` is accepted `bytes` is too, and for a secret that
     is the form to use
+- **the entry side takes one form that is not a copy**: where a scalar is
+    accepted, so is a cffi array of 32 octets —
+    `ffi.new("unsigned char[32]", ...)`, or any other item type an octet
+    wide — and where the call only reads it, it reaches libsecp256k1 as
+    it stands. That is the whole of the difference: a `bytes` or a
+    `bytearray` is copied at the boundary, deliberately, so that nothing
+    can change what libsecp256k1 is about to read, and the copy of a
+    secret is an immutable object nothing can overwrite. A caller signing
+    again and again under one key can therefore hold it in memory it
+    owns, wipe that memory when done, and have no `bytes` of the key made
+    per call. What this does not change is what came before: the value
+    the buffer was filled from existed already, and the arithmetic that
+    produced it happened where this package cannot see — which is why
+    filling one from an `int` buys nothing the bullet above does not
+    already refuse. And the wipe stays the caller's, as it is for `into`:
+    a buffer never overwritten is the same un-zeroizable copy under
+    another name.
+
+    **What the caller takes on is three things, and `_scalar` names them
+    where it refuses the shapes it cannot take.** The octets must stay
+    put for the whole call, which is more than one read — libsecp256k1
+    loads the scalar and then derives the nonce from the same pointer, and
+    grinding and the check read it again — so a write in between yields a
+    nonce and a signature under two different keys, reported as the fault
+    it is indistinguishable from. The memory must outlive the call, which
+    no python argument has had to promise: a cffi *view*, a slice or a
+    cast, does not keep its owner alive, and a dangling one reads freed
+    memory as a private key. And the length is the declaration's word:
+    `ffi.cast("unsigned char[32]", ...)` over 8 octets is accepted, cffi
+    having no way to report what was really allocated.
+
+    Four calls copy the octets into a buffer of their own instead, and
+    have to: `keys.prvkey_negate`, `keys.prvkey_tweak_add`,
+    `keys.prvkey_tweak_mul` have libsecp256k1 write the answer through
+    that pointer, and the sender side of `silentpayments` wipes it on the
+    way out — so passing the caller's memory would negate or zero the key
+    they handed in. Each of those answers a *new* secret, which is what
+    `into` above is for
+
+    ```python
+    prvkey = ffi.new("unsigned char[32]", secret_octets)
+    try:
+        sig = dsa.sign(msg, prvkey)   # no bytes of the key is made
+    finally:
+        ffi.buffer(prvkey)[:] = bytes(32)   # the caller's wipe
+    ```
+
 - randomness comes from `secrets.token_bytes`, i.e. from the operating
     system, both for the context randomization and for the auxiliary
     randomness of BIP340 signing and of ElligatorSwift encoding
--- a/btclib_secp256k1/_scalar.py
+++ b/btclib_secp256k1/_scalar.py
@@ -11,6 +11,12 @@ import secrets
 
 from . import BytesLike, CData, ffi
 
+# the one item type cffi will pass where libsecp256k1 declares
+# `const unsigned char *`, resolved once at import as the buffer types of
+# the wrappers are: `_owned_octets` re-views a caller's 32 octets as this,
+# whatever they declared them
+_OCTETS_TYPE = ffi.typeof("unsigned char[32]")
+
 
 def octets(value: BytesLike, name: str, size: int | None = None) -> bytes:
     """Normalize an argument to the bytes a bare pointer needs, of a length.
@@ -90,7 +96,120 @@ def octets(value: BytesLike, name: str, size: int | None = None) -> bytes:
     return value_bytes
 
 
-def scalar(num: BytesLike | int, name: str) -> bytes:
+def _owned_octets(num: CData, name: str) -> CData:
+    """Accept 32 octets of memory the caller owns, unconverted.
+
+    The one argument of these bindings that is not copied on the way in,
+    and the reason is what a copy would be: an immutable `bytes` of a
+    secret, which nothing can overwrite and which stays until the
+    collector gets to it. A caller holding the key in memory it can zero
+    -- `ffi.new("unsigned char[32]")`, wiped with `_secret.wipe` -- would
+    otherwise have one made per call, so what `octets` is right to do for
+    a value is wrong for this one.
+
+    What that gives up is stated where `octets` states the opposite: the
+    copy taken there is what stops a caller overwriting their own buffer
+    while libsecp256k1 reads it. **Three obligations pass to the caller
+    with the copy**, and none of them exists for a `bytes`:
+
+    - the octets stay put for the whole call, and the whole call is more
+      than one read. `secp256k1_ecdsa_sign` loads the scalar
+      (`secp256k1/src/secp256k1.c:555`) and then hands the same pointer
+      to the nonce derivation (:563), and above it `dsa._sign_` reads it
+      again for every grinding attempt and once more in `_checked`'s
+      failing branch. So a write in between is not "a different key
+      signed": it is a nonce and a signature derived from two different
+      keys, which arrives as the fault `RuntimeError` and looks like
+      hardware;
+    - the memory outlives the call, which no python argument has ever had
+      to promise. A cdata view does not keep its owner alive:
+      `ffi.new("unsigned char[64]")[0:32]` and `ffi.cast(...)` over a
+      temporary read freed memory, measured -- zeros here by the
+      allocator's grace, and a plausible scalar on a busier heap. What
+      this function returns is safe in that respect, `from_buffer`
+      keeping its view alive, but it cannot rescue a dangling argument;
+    - the declaration is the truth about the length, because `ffi.sizeof`
+      answers what the *type* says rather than what was allocated:
+      `ffi.cast("unsigned char[32]", <8 octets>)` clears every refusal
+      below and has libsecp256k1 read 24 octets of whatever follows.
+      cffi has nothing to check that against, so neither has this.
+
+    The alternative on offer is the copy that cannot be wiped, which is
+    why the trade is a caller's to make and not a default.
+
+    Any 1-octet array is taken, whatever the caller declared it as, and
+    what makes that safe is a re-view rather than a conversion:
+    `ffi.from_buffer` over `ffi.buffer` answers an `unsigned char[32]`
+    pointing at the caller's own memory, which is the one item type cffi
+    will pass to `const unsigned char *`. Without it the acceptance would
+    be `char` and `unsigned char` alone -- cffi holds `uint8_t` and
+    `signed char` to be primitives of their own, so `uint8_t[32]`, which
+    is what a C programmer writes for 32 octets, would clear every check
+    here and die at the boundary in cffi's words.
+
+    **Not `ffi.cast`, and this is the trap in the neighbourhood.** It
+    would answer the same pointer for nothing, and it does not keep the
+    memory alive: a cast whose owner is dropped reads freed memory, which
+    was measured rather than read off the documentation -- 32 octets that
+    no longer hold the key, with no error anywhere. `from_buffer` keeps a
+    reference to what it views, and the buffer it views keeps the cdata,
+    so the chain holds for as long as the value is in use.
+
+    Which shapes are refused is the whole of the care here, because what
+    follows is a bare pointer libsecp256k1 reads 32 octets from:
+
+    - a *pointer* rather than an array, whose `ffi.sizeof` is 8 on this
+      machine and says nothing about what it points at. That is the trap
+      `_secret.wipe` records from the other side, where `ffi.sizeof`
+      would have wiped a quarter of a private key and reported success
+    - an array whose items are wider than an octet. `uint32_t[8]` is 32
+      octets of whatever this machine's byte order made of them, which
+      is exactly what `octets` refuses a `memoryview` of wider items
+      for, and refused rather than reinterpreted for the same reason
+    - an array of the wrong length, which is the check every other
+      scalar gets and the one a bare pointer cannot be given later
+
+    Args:
+        num: the cdata, as the caller passed it.
+        name: what the scalar is, as the exception should call it.
+
+    Returns:
+        The caller's own 32 octets as an `unsigned char[32]`, which is a
+        view of that memory and not a copy of it: writing through the
+        original changes what libsecp256k1 reads.
+
+    Raises:
+        TypeError: if it is not a cdata at all, or is not an array of
+            octets. A cdata's own `cname` is in the message, that being
+            what a caller declared.
+        ValueError: if it is an array of octets and not 32 of them.
+    """
+    # what reaches here is whatever was neither bytes nor an int, and the
+    # message for it is `scalar`'s own, stated once for the two ways of
+    # arriving at it
+    not_a_scalar = f"the {name} must be bytes or an int, not {type(num).__name__}"
+    # a `str` is refused before the question rather than by it, because
+    # `ffi.typeof` reads a str as a *cdecl*: `"x" * 32` comes back as
+    # cffi's own `error: undefined type name`, which is not this
+    # function's TypeError and not about the argument -- and `"char[32]"`
+    # is worse, being a cdecl that resolves, 32 octets wide, so the str
+    # would have been accepted and handed to libsecp256k1 as a str.
+    # `tests/test_core.py` holds both
+    if isinstance(num, str):
+        raise TypeError(not_a_scalar)
+    try:
+        ctype = ffi.typeof(num)
+    except TypeError:
+        raise TypeError(not_a_scalar) from None
+    if ctype.kind != "array" or ffi.sizeof(ctype.item) != 1:
+        msg = f"the {name} must be a cffi array of octets, not {ctype.cname}"
+        raise TypeError(msg)
+    if ffi.sizeof(num) != 32:
+        raise ValueError(f"the {name} must be 32 bytes")
+    return ffi.from_buffer(_OCTETS_TYPE, ffi.buffer(num))
+
+
+def scalar(num: BytesLike | int | CData, name: str) -> bytes | CData:
     """Normalize a scalar argument to 32 bytes.
 
     An int is serialized big endian, as libsecp256k1 expects; anything
@@ -102,6 +221,23 @@ def scalar(num: BytesLike | int, name: str) -> bytes:
     disbelieved, whereas an int states only a value and the width is the
     curve's.
 
+    A cffi array of 32 octets is the one thing not copied. It is memory
+    the caller owns and can overwrite, so converting it would make the
+    copy that cannot be -- `_owned_octets` is that decision, what it
+    refuses, and the re-view that lets any 1-octet spelling through.
+    Where libsecp256k1 writes through the pointer instead of reading it,
+    or where this package wipes the buffer afterwards, a copy is owed and
+    `_secret.scalar_buffer` is the one that takes it.
+
+    Where it matters is not the per-signature path, the private halves
+    handing libsecp256k1 the pointer without coming through here at all,
+    but the *derivation*:
+    `keys._pubkey_from_prvkey_` asks for a scalar, so before this a key
+    held in a buffer could not have its public key derived -- and
+    `dsa._checked`'s failing branch, which derives in order to tell a
+    wrong argument from a fault, answered a `TypeError` about the
+    argument instead of the diagnosis it exists for.
+
     A secret is better passed as bytes, for a narrow reason. Not the
     serialization, which is a loop over nine CPython digits and measures
     as noise, but the python arithmetic that produced the int, variable
@@ -112,16 +248,19 @@ def scalar(num: BytesLike | int, name: str) -> bytes:
     must not leak belongs where that can be promised.
 
     Args:
-        num: the scalar, exactly 32 octets or an int in [0, 2**256).
+        num: the scalar, exactly 32 octets or an int in [0, 2**256). The
+            octets may be a cffi array of them, which is passed on as it
+            stands rather than copied.
         name: what the scalar is, as the exception should call it.
 
     Returns:
-        The scalar as 32 bytes, big endian.
+        The scalar as 32 bytes, big endian -- or the caller's own cdata,
+        where that is what was handed in.
 
     Raises:
         TypeError: if the value is neither an int nor one of the types
-            `octets` takes, a bool counting as neither although python
-            makes it an int.
+            `octets` takes nor a cffi array of octets, a bool counting as
+            none of them although python makes it an int.
         ValueError: if it is not exactly 32 octets long, or if an int
             does not fit in 32 bytes. Whether the value is a valid
             scalar, i.e. in [1, n-1], is for libsecp256k1 to say.
@@ -146,10 +285,12 @@ def scalar(num: BytesLike | int, name: str) -> bytes:
         if not 0 <= num < 2**256:
             raise ValueError(f"the {name} must fit in 32 bytes")
         return num.to_bytes(32, "big")
-    # the domain here is octets' plus the int above, so a value that is
-    # neither is named here: what is left for octets is the length
+    # the domain here is octets' plus the int above plus a cdata, and it
+    # is asked in that order: everything a value can be is settled before
+    # `ffi.typeof` is called at all, so the common paths pay nothing for
+    # this one. What is left for octets is the length
     if not isinstance(num, (bytes, bytearray, memoryview)):
-        raise TypeError(f"the {name} must be bytes or an int, not {type(num).__name__}")
+        return _owned_octets(num, name)
     return octets(num, name, 32)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,135 @@ release-notes length in the first place, and are still in
   sentence that said the saving was "available and not yet taken" for
   `recovery` is now the one that says what taking it cost.
 
+### A scalar may be memory the caller can overwrite
+
+- **`_scalar.scalar` takes a cffi array of exactly 32 octets and hands it
+  on unconverted** (#247), so wherever a private key or a tweak is
+  accepted, `ffi.new("unsigned char[32]", ...)` is too. It is the only
+  argument of these bindings that is not copied at the boundary, and the
+  reason is what the copy would be: an immutable `bytes` of a secret,
+  which nothing can overwrite and which stays until the collector gets to
+  it. A caller signing again and again under one key can now hold it in
+  memory it owns and wipe that when done, with no `bytes` of the key made
+  per call. What that gives up is stated where `octets` states the
+  opposite -- the copy is also what stops a caller changing the octets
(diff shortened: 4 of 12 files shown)


fametrano:
READY b684a9c54c62908787a25cc5851cffb317e0d4c4

A scalar may now be a cffi array of exactly 32 octets, passed to
libsecp256k1 unconverted, so a caller can hold a private key in memory it
owns and wipe it. The body carries the reasoning; the two things worth
your attention first:

- **the issue's premise did not survive a prototype.** A buffer-held key
  already signed identically before this change, the private halves
  handing libsecp256k1 the pointer. What binds is the derivation, and the
  case that matters is `dsa._checked`'s failing branch answering a
  `TypeError` about the argument where it exists to say whether the
  argument or the computation was wrong. The change is scoped to that, and
  the README's speed verdict on a `dsa.Signer` is untouched
- **the `str` case is a bug the suite found.** `ffi.typeof` reads a str as
  a cdecl, so `"char[32]"` resolved, measured 32 octets, and would have
  been handed on as a str. Refused before the question now, with both
  spellings as cases

Evidence, local:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, 650 passed, 100.00%
- `uvx pre-commit run --all-files` — exit 0, 37 hooks, nothing modified (markdownlint fixed one list indent in `SECURITY.md` on the first pass, committed)
- by hand, on this sha: `scalar(buf, ...) is buf`; the same signature, the same derived key and the same BIP340 signature as the equivalent `bytes`; zero after `_secret.wipe`; `unsigned char *`, `secp256k1_keypair *`, `uint32_t[8]`, `unsigned char[31]`, `"char[32]"` and `"x" * 32` each refused with their own message; the fault path `RuntimeError`, the wrong key `ValueError`

`PYTHONDONTWRITEBYTECODE=1` throughout. Independent of PR 252 — different files, no stack.

README.md
```diff
@@ -570,6 +570,30 @@ the paragraph above describes, a second copy of the secret held for as
 long as the signer is. That trade is worth making for half a signature
 and not for a hundredth of one.
 
+**The other ground such a signer could be argued on is the memory, and
+it needs no signer.** Where a scalar is accepted, a cffi array of exactly
```
fametrano: **blocking** — "Where a scalar is accepted, a cffi array of exactly 32 octets is accepted too" is not true, and the same claim is at `SECURITY.md:149` and in the pull request body. Four call sites still copy through `ffi.new(..., scalar(...))`, and a cdata cannot initialize the array they ask for. Which of them breaks depends on the cdecl the *caller* chose, so no spelling works everywhere:

```text
                                 ffi.new("char[32]")   ffi.new("unsigned char[32]")
dsa.sign                         ok                    ok
keys.pubkey_from_prvkey          ok                    ok
keys.prvkey_negate               ok                    TypeError
keys.prvkey_tweak_add            ok                    TypeError
keys.prvkey_tweak_mul            ok                    TypeError
silentpayments.create_outputs    TypeError             ok
```

The message is cffi's, about an internal ctype, naming neither the argument nor this package:

```text
TypeError: initializer for ctype 'char[32]' must be a bytes or list or tuple,
           not cdata 'unsigned char[32]'
```

`unsigned char[32]` is the spelling `SECURITY.md`'s own example uses, and it is the one that breaks the three `keys.py` functions — `keys.py:130`, `keys.py:177`, `keys.py:221`. `silentpayments.py:139` is the mirror image, spelling its buffer `unsigned char[32]` and so refusing a `char[32]`. That asymmetry is incidental cffi behaviour rather than a design: an `unsigned char[]` array initializes element-wise from anything iterable of ints, which a cdata array of octets is, and a `char[]` array demands `bytes`.

**And one public entry point fails for half of all keys.** `ssa.nonce_bip340` reaches `keys.prvkey_negate` only where the key's point has odd y (`ssa.py:127-130`), so with a buffer-held key it is a coin flip:

```text
prvkey 5  parity 0  ->  ok
prvkey 6  parity 1  ->  TypeError: initializer for ctype 'char[32]' must be ...
prvkey 7  parity 0  ->  ok
```

Which is also why the suite is green: `PRVKEY = 7` and `SECRET` both have even y, and the three tests that hand in a buffer go to `keys.pubkey_from_prvkey`, `dsa.sign` and `ssa.sign` — none of the four copying sites, and never the odd-y half of `nonce_bip340`. A claim of the form "wherever a scalar is accepted" wants a test that walks the entry points that take one, not three of them.

**What the fix is is a decision, not a line**, which is why this is a finding rather than a diff. Those three `keys.py` sites allocate *because* libsecp256k1 mutates the buffer in place — `secp256k1_ec_seckey_negate`, `_tweak_add`, `_tweak_mul` all write through the pointer — and the caller's own key must not be overwritten by a function that returns a new one. So a copy there is required for correctness and a pass-through would be wrong; the honest resolutions are to copy out of the cdata (`ffi.new("char[32]", bytes(ffi.buffer(k)))`, with the comment saying that this is the one place the copy is owed) or to refuse a cdata at those entry points with this package's own message and narrow the claim to the paths that do take one. Either is defensible. A raw cffi error about `char[32]` is not, and it is exactly the misdiagnosis ISS 247 was filed about.

btclib_secp256k1/_scalar.py
```diff
@@ -90,7 +90,79 @@ def octets(value: BytesLike, name: str, size: int | None = None) -> bytes:
...
+    # is worse, being a cdecl that resolves, 32 octets wide, so the str
+    # would have been accepted and handed to libsecp256k1 as a str.
+    # `tests/test_core.py` holds both
+    if isinstance(num, str):
+        raise TypeError(not_a_scalar)
+    try:
+        ctype = ffi.typeof(num)
+    except TypeError:
+        raise TypeError(not_a_scalar) from None
+    if ctype.kind != "array" or ffi.sizeof(ctype.item) != 1:
```
fametrano: **blocking** — this gate checks the item's *width* and not its *type*, and cffi will not pass every 1-octet array to `const unsigned char *`. `uint8_t[32]` and `signed char[32]` are arrays of octets by this test — `kind == "array"`, `ffi.sizeof(ctype.item) == 1`, `ffi.sizeof(num) == 32` — so they clear all three refusals, and then die inside the call:

```text
>>> dsa.sign(msg, ffi.new("uint8_t[32]", list(prv)))
TypeError: initializer for ctype 'unsigned char *' must be a pointer to same type,
           not cdata 'uint8_t[32]'
>>> keys.pubkey_from_prvkey(ffi.new("signed char[32]", list(prv)))
TypeError: initializer for ctype 'unsigned char *' must be a pointer to same type,
           not cdata 'signed char[32]'
```

cffi treats `uint8_t` as a primitive of its own rather than an alias of `unsigned char`, and `signed char` likewise, so the set it will accept for that parameter is `char` and `unsigned char` and nothing else — which the width test cannot express. The docstring says "Which shapes are refused is the whole of the care here, because what follows is a bare pointer libsecp256k1 reads 32 octets from", and these are two shapes that reach that pointer and are refused by it in cffi's words instead of this function's.

`uint8_t[32]` is not an exotic spelling for a caller to reach for — it is the one a C programmer would write for 32 octets, and `stdint` types are what `ffi.new` accepts without a declaration of their own.

Two ways out, and they differ in what they promise:

- narrow the test to the item types cffi will pass — `ctype.item.cname in ("char", "unsigned char")` — so the refusal is this function's, with `cname` in the message as the other refusals already have it
- or widen the acceptance with `ffi.cast("unsigned char[32]", num)`, which is free and copies nothing, so any array of 32 octets a caller owns goes through whatever they declared it as

The second is the larger claim and the one the surrounding prose implies ("a cffi array of exactly 32 octets"); the first is the smaller and needs a sentence saying which spellings. Either way the two shapes above should not reach libsecp256k1 to be refused there.

fametrano:
Reviewed **b684a9c5** in a read-only worktree of my own at that sha — one commit on `5a8248b`, signed `G Ferdinando M. Ametrano`, `origin/main` an ancestor.

**Gates, both exit 0, which is why the findings below are not gate failures:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 650 passed, 100.00%

**Two blocking findings, both inline, both from probes rather than from reading**

1. [README.md:574](https://github.com/btclib-org/btclib-secp256k1/pull/253#discussion_r3803637680) — "where a scalar is accepted, so is a cffi array of exactly 32 octets" is false at four call sites, and which ones break depends on the cdecl the caller chose: `unsigned char[32]`, the spelling `SECURITY.md`'s own example uses, is refused by `keys.prvkey_negate`, `keys.prvkey_tweak_add` and `keys.prvkey_tweak_mul`; `char[32]` is refused by `silentpayments.create_outputs`. The message is cffi's, about an internal `char[32]`, naming neither the argument nor this package. **And `ssa.nonce_bip340` fails for half of all keys** — it reaches `prvkey_negate` only where the point has odd y, so prvkey 6 raises where 5 and 7 pass. The fix is a decision rather than a line: those three sites copy *because* libsecp256k1 writes through the pointer, so a copy there is owed rather than avoidable.
2. [_scalar.py:157](https://github.com/btclib-org/btclib-secp256k1/pull/253#discussion_r3803637810) — the shape gate tests the item's width, not its type. `uint8_t[32]` and `signed char[32]` clear all three refusals and then die inside the C call with `initializer for ctype 'unsigned char *' must be a pointer to same type` — cffi's words, from past the function whose stated purpose is to refuse a bad shape in its own.

**Non-blocking**

3. What let both of those past a green suite is the shape of the new tests: a buffer is handed to `keys.pubkey_from_prvkey`, `dsa.sign`, `ssa.sign`, `keys.prvkey_verify` and `_scalar.scalar` directly — five of the twenty-four call sites of `scalar`, none of the four that copy. A claim quantified over "wherever a scalar is accepted" is the kind that wants a test walking the entry points that take one; the same test would have caught both findings, and the parity case would have needed one odd-y key among them. `PRVKEY = 7` and `SECRET` are both even-y, which is the whole of why the suite is quiet.

**What I checked and found sound**, so that it is not re-litigated: the ordering in `scalar` really does leave the common paths untouched, `ffi.typeof` being asked only after `bytes`, `int` and the three `octets` types are settled; the `str` bug is real and both spellings are worth the cases they got — I confirmed `ffi.typeof("char[32]")` resolves to a 32-octet type, so that argument would indeed have reached libsecp256k1 as a `str`; the pass-through is genuinely a pass-through (`scalar(held, ...) is held`); the wipe works and the signature and derived key match the `bytes` answers; the pointer, struct-pointer, wider-item and short-length refusals all answer with this package's own message; and the narrowing of ISS 247's premise is right — a buffer-held key did reach `_signed` before this change, so the derivation and `dsa._checked`'s failing branch are the real subject, which is what the tests and prose now say.

I have filed no issues: both findings are inside this diff.

CHANGES REQUESTED b684a9c54c62908787a25cc5851cffb317e0d4c4

SECURITY.md
```diff
@@ -149,6 +145,31 @@ These are known and inherent, not vulnerabilities:
     type, a length or a magnitude and never on the content of a secret.
     Everywhere an `int` is accepted `bytes` is too, and for a secret that
     is the form to use
+- **the entry side takes one form that is not a copy**: where a scalar is
+    accepted, so is a cffi array of exactly 32 octets —
```
fametrano: **non-blocking, three of them** — this bullet states two obligations, the wipe and the synchronization, and they are the right two to state. Asked what else this argument opens, I probed rather than reasoned, and three things came back that the prose does not cover. None is a defect in the code; all three are the boundary of what the check can see, which is the sort of thing this file exists to write down.

**1. One call reads the buffer more than once, so the window is not "while libsecp256k1 reads it".** In the vendored tree, `secp256k1/src/secp256k1.c:555` loads the scalar with `secp256k1_scalar_set_b32_seckey(&sec, seckey)` and then line 563, inside the retry loop, hands *the same caller pointer* to the nonce function. Above that, `dsa._sign_` reads it again in `_grind`'s loop and a third time in `_checked`'s failing branch. So a caller who mutates between those reads does not get a torn value: they get a nonce derived from one key and a signature made with another, which makes false the identity `dsa.nonce_rfc6979` exists to reproduce and `tests/test_nonces.py` asserts. And it surfaces as `RuntimeError: signing produced a signature that does not verify` — the fault message, for a caller's race. Worth a clause, because "the synchronization was already yours" undersells what desynchronization costs here.

**2. A cdata view does not keep its owner alive, and no python type can do this.** A `memoryview` holds a reference to its bytearray; a cffi slice or cast does not hold one to the array it views. Measured, this sha:

```text
whole = ffi.new("unsigned char[64]", VALID + bytes(32))
whole[0:32]                                        octets=0102030405060708...  -> pubkey ok
ffi.new("unsigned char[64]", VALID + bytes(32))[0:32]   octets=0000000000000000...  -> ValueError
ffi.cast("unsigned char[32]", ffi.new("unsigned char[32]", VALID))   octets=00000000...  -> ValueError
```

The second and third read freed memory and signed nothing, being zero here by the allocator's grace rather than by any check. On a busier heap those 32 octets can be a valid scalar — including a secret that previously lived there — and then the answer is a signature under a key nobody chose, silently. `bytes` made this unreachable by copying at the boundary; accepting memory makes lifetime the caller's third obligation, and it is the one a python programmer has never had to think about. One sentence beside the wipe.

**3. The length check checks the declaration, not the memory** — which is worth saying where `_owned_octets` presents the three refusals as closing the `ffi.sizeof` trap. A cast reopens it exactly:

```text
lied = ffi.cast("unsigned char[32]", ffi.new("unsigned char[8]", b"\x01" * 8))
ffi.typeof(lied).cname   -> 'unsigned char[32]'      # kind array, item 1 octet
ffi.sizeof(lied)         -> 32                       # the declared width, not the allocation
bytes(ffi.buffer(lied))  -> 0101010101010101 0900000000000000 0900000000000000 ...
                                              ^ adjacent heap, read as private key
```

All three refusals pass and libsecp256k1 reads 32 octets from an 8-octet allocation. There is no fix in the code — cffi cannot be asked what was really allocated — so the honest form is the docstring saying the check trusts the caller's declaration, the same way this package trusts the `ffi` and `lib` it exposes. The pointer refusal is still worth having; it just stops one spelling of the mistake rather than the mistake.

**What I checked and found sound**, so it is not left hanging: `ffi.cast("unsigned char[32]", ffi.NULL)` is caught by libsecp256k1's own `ARG_CHECK(seckey != NULL)` — `prvkey_verify` answers False and `context.check()` reports `libsecp256k1 illegal argument: seckey != NULL`, no dereference. A slice whose owner *is* alive is a genuine view, 32 real octets, and derives the same public key as the `bytes` — so the useful case works and copies nothing. And none of this can make a bad signature verify: the octets reach a one-way function in every path, so an out-of-bounds read here is not a read primitive.

fametrano:
Addendum to the review of **b684a9c5**, on the security surface this argument opens rather than on the diff's correctness. Three more findings, all **non-blocking** and all documentation, in one thread: [SECURITY.md:149](https://github.com/btclib-org/btclib-secp256k1/pull/253#discussion_r3803699543).

In short, and each measured on this sha:

1. **one call reads the buffer more than once** — `secp256k1.c:555` loads the scalar and `:563` hands the same caller pointer to the nonce function, so a mutation between them yields a nonce and a signature that disagree, reported as the fault `RuntimeError`. "The synchronization was already yours" undersells that
2. **a cdata view does not keep its owner alive** — a slice or cast of a temporary reads freed memory and no python type can do this, `memoryview` holding its bytearray. Lifetime is the third obligation and the only one new to a python caller
3. **the length check checks the declaration** — `ffi.cast("unsigned char[32]", <8 octets>)` clears all three refusals with `ffi.sizeof` answering 32, and libsecp256k1 reads adjacent heap as the key. No code fix exists; the docstring should say the check trusts what the caller declared

And what I checked that holds, so it is not reopened: NULL is caught by libsecp256k1's own `ARG_CHECK(seckey != NULL)` with no dereference; a live slice is a real 32-octet view and derives the right key without a copy; and no path here can turn an out-of-bounds read into a read primitive, every consumer of those octets being a one-way function. The security *gain* is real and is the reason to take the change: a key held in a buffer no longer has an immutable `bytes` of itself made per call.

The verdict is unchanged and is the one already posted — the two blocking findings are the earlier ones, not these.

CHANGES REQUESTED b684a9c54c62908787a25cc5851cffb317e0d4c4

README.md
fametrano: Accepted, in `30b6b0c`. Reproduced your matrix first, including the coin flip: with `unsigned char[32]`, prvkey 5 (parity 0) and 7 (parity 0) answered and 6 (parity 1) raised `initializer for ctype 'char[32]' must be a bytes or list or tuple`.

**The decision: the copy stays, and it is taken from the cdata.** Refusing a cdata at those four entry points would have been a hole a caller cannot predict from the outside — `ssa.nonce_bip340` would take one for half the keys — so the four now copy through `_secret.scalar_buffer`, which is `ffi.new("char[32]")` followed by `ffi.memmove`. `memmove` takes a cdata source where `ffi.new`'s initializer cannot, and it makes no `bytes` of the secret in between, so the caller's reason for holding a buffer survives the copy.

Its docstring states which of *two* reasons each site has, because they are not the same: `secp256k1_ec_seckey_negate`, `_tweak_add` and `_tweak_mul` write the answer through the pointer, so the caller's key must not be the buffer; `silentpayments._create_outputs_` wipes its `seckeys` in a `finally`, so passing the caller's memory would zero their key on the way out of a call that only read it. Verified: after every one of those calls the caller's buffer still holds its octets.

`README.md` and `SECURITY.md` now say "where the call only *reads* it", name the four that copy with their reasons, and point at `into` for the new secret each of them answers.

**And the sweep you asked for.** `tests/test_bytes_like.py` drives its whole table with the private keys and tweaks handed in as `unsigned char[32]`, swapped by identity out of its own data, so an entry point added later is covered without being added twice. Two assertions per row: the same answer, and the caller's octets still there afterwards — the half a comparison of answers cannot see, and what a missing copy shows up as. Reverting `keys.py:130` alone fails it on the `keys.prvkey_negate` row, which I checked before writing the fix down. The parity branch has a case of its own beside it (5 and 6, both parities asserted to occur), since `PRVKEY` is 7 and no row reaches the negation through `nonce_bip340`.

btclib_secp256k1/_scalar.py
fametrano: Accepted, in `30b6b0c`, and I took your second option — but not with `ffi.cast`.

Both spellings reproduced (`uint8_t[32]` and `signed char[32]`, each dying with `initializer for ctype 'unsigned char *' must be a pointer to same type`). Then I measured the cast before building on it, and it fails a property the narrow gate would not have needed:

```python
def make():
    buf = ffi.new("uint8_t[32]", list(secret))
    return ffi.cast("unsigned char[32]", buf)
c = make(); gc.collect()
bytes(ffi.buffer(c)) == secret   # False
```

A cast does not keep its owner alive, so a caller whose buffer is a temporary would have had libsecp256k1 read freed memory — silently, and as a private key. `ffi.from_buffer(_OCTETS_TYPE, ffi.buffer(num))` gives the same pointer for the same nothing and keeps the chain alive: measured through the same `gc.collect()`, and measured to be a view rather than a copy (a write through the original is visible through the result, and pointer equality holds). So the acceptance is any item type an octet wide — `char`, `unsigned char`, `uint8_t`, `signed char` — with the three refusals unchanged in front of it.

The docstring says why it is `from_buffer` and not `cast`, with that measurement, because the cast is the obvious simplification for someone to reach for later.

SECURITY.md
fametrano: All three accepted, in `30b6b0c`, and they landed as one paragraph rather than three: `_owned_octets` now names **three obligations that pass to the caller with the copy**, prefaced by the fact that none of them exists for a `bytes`. That framing is yours — lifetime being the one a python caller has never had to think about is what makes the list worth reading.

1. **The multi-read.** Verified in the vendored tree: `secp256k1/src/secp256k1.c:555` is `secp256k1_scalar_set_b32_seckey(&sec, seckey)` and `:563` hands that same `seckey` to `nonce_function_rfc6979_impl`, with `dsa._sign_` reading it again per grinding attempt and once more in `_checked`'s failing branch. The docstring and `SECURITY.md` now say what a write in between actually produces — a nonce and a signature derived from two different keys, arriving as the fault `RuntimeError` — rather than "a different value read". You are right that "the synchronization was already yours" undersold it; that clause is gone.
2. **Lifetime.** Named as its own obligation, with the measurement, and with what this function does and does not do about it: `from_buffer` keeps its own view alive, and cannot rescue an argument that was already dangling.
3. **The declared length.** Reproduced — `ffi.cast("unsigned char[32]", ffi.new("unsigned char[8]", ...))` clears all three refusals, `ffi.sizeof` answering 32, and the derived key comes out of 24 octets of the next allocation. Documented as a limit rather than papered over: `ffi.sizeof` answers what the type says, and cffi has nothing to check that against, so neither has this.

Nothing here changed the code, which is why the round's diff is the two blocking ones plus prose. Thank you for sending them mid-round — they went in with the rest instead of costing another turn.

fametrano:
READY 30b6b0c195103ff9828869707b5c9b5d1bcfbdd6

Round 2, on `b684a9c` — a second commit, so `b684a9c..30b6b0c` is the round; both squash at landing. All five findings accepted, none declined, each answered in its thread.

**The two blocking ones changed the shape of the change:**

- the four copying sites now copy *from the cdata*, through `_secret.scalar_buffer` (`ffi.new` + `ffi.memmove`), which is the one statement of a copy that is owed for two different reasons — libsecp256k1 writing the answer through the pointer in the three `keys` functions, and this package wiping the buffer in `silentpayments`. Refusing a cdata there would have been a hole a caller cannot see from outside, `ssa.nonce_bip340` taking one for half of all keys
- the item type is no longer tested by width but made irrelevant: `ffi.from_buffer` over `ffi.buffer` re-views any 1-octet array as `unsigned char[32]` at the caller's own memory. **Not `ffi.cast`** — measured, a cast does not keep its owner alive, so a temporary would have had libsecp256k1 read freed memory as a key

**The sweep is the part that would have caught both.** `tests/test_bytes_like.py` drives its whole table with a buffer, swapping the scalars of its own data by identity, and asserts the answer *and* that the caller's octets survive the call. Reverting `keys.py:130` alone fails it on that row — checked before writing the fix down. The parity branch has its own case, `PRVKEY` being even-y.

**The addendum went in with the round**: three obligations named in `_owned_octets` and in `SECURITY.md` — the octets stay put for a call that reads them more than once (`secp256k1.c:555` loads the scalar, `:563` derives the nonce from the same pointer), the memory outlives the call, and the length is the declaration's word (a cast of 8 octets is accepted, and 24 octets of the next allocation are read as a key). Documented rather than fixed, cffi having nothing to check them against.

Evidence, local, on this sha:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, **716 passed**, 100.00% (845 statements, 188 branches)
- by hand: five entry points × four cdecls × the caller's buffer intact; `nonce_bip340` for keys 5, 6, 7; the reverted-site check above; `secp256k1.c:555`/`:563` read in the vendored tree; the `gc.collect()` measurements for `cast` and `from_buffer`

`PYTHONDONTWRITEBYTECODE=1` throughout.

fametrano:
Round 2 reviewed: the delta `b684a9c5..30b6b0c1`, one commit, signed `G Ferdinando M. Ametrano`. All three threads **resolved** — five findings addressed, none declined.

**Gates, in a fresh worktree of my own at 30b6b0c1:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 716 passed, 100.00%

**The two blocking findings, re-checked with the matrix that found them**

Every entry point, every 1-octet spelling, built with the initializer each one takes:

```text
                      dsa.sign  pubkey_from  negate  tweak_add  tweak_mul  create_outputs
char[32]              ok        ok           ok      ok         ok         ok
unsigned char[32]     ok        ok           ok      ok         ok         ok
uint8_t[32]           ok        ok           ok      ok         ok         ok
signed char[32]       ok        ok           ok      ok         ok         ok
int8_t[32]            ok        ok           ok      ok         ok         ok
```

and `ssa.nonce_bip340` answers for prvkey 5, 6 and 7 — both parities, where 6 was the coin flip.

**And the four properties the fix rests on, measured rather than read:**

- the re-view is a view: `scalar(held, ...)` comes back at the *same address*, retyped `unsigned char[32]`, and a write through the caller's original shows up in it
- the lifetime chain holds, which is the part `ffi.cast` would not have given you: a view of an array dropped immediately still held the key after `gc.collect()`, and still derived the right public key. Your reason for not using `cast` is the right one and it is the one I measured in the addendum
- the copy is taken where owed: after `prvkey_negate`, `prvkey_tweak_add`, `prvkey_tweak_mul` and `create_outputs`, the caller's octets are unchanged and every answer equals the `bytes` answer
- the sweep is not vacuous, which is the question its own shape invites. 34 of the 65 rows swap a buffer, no scalar hides in a `kwargs` the sweep does not retype, and all four copying sites plus `nonce_bip340` are covered by rows that do swap. Reverting the three `keys.py` sites turns exactly `[keys.prvkey_negate]`, `[keys.prvkey_tweak_add]`, `[keys.prvkey_tweak_mul]` and `test_the_negating_half_of_a_nonce_takes_a_buffer_too` red — 4 failed — and restoring them gives 197 passed

**And the safety net around the re-view itself**, since it is now the load-bearing line. Two hand mutations in my own worktree, `PYTHONDONTWRITEBYTECODE=1`, control run after each:

```text
return num                                          -> 1 failed  (test_core, the four item types)
return ffi.new(_OCTETS_TYPE, bytes(ffi.buffer(num))) -> 1 failed  (test_core, the aliasing assertion)
restored                                            -> 716 passed
```

So both halves of what `_owned_octets` promises — any item type, and a view rather than a copy — are pinned by an assertion rather than by the prose. That is the answer to the round-1 non-blocking finding, and it is a better answer than the sweep alone would have been.

**One observation, not a finding.** With *only* the `silentpayments` site reverted, nothing fails — the sweep passes and so does the whole suite. That is correct rather than a hole: the re-view now hands that site an `unsigned char[32]`, which is exactly what its old `ffi.new` initializer accepted, so the old spelling is no longer defective there. `scalar_buffer` is still right at that site, for the reason its docstring gives — the wipe, not the cdecl — and the reason is what the test is holding. Worth knowing that the *cffi* half of that site's justification is now historical.

The three security obligations are stated where I hoped they would be, in `SECURITY.md` beside the wipe and in `_owned_octets` beside the shapes it refuses, and the multiple-read one names `secp256k1.c:555` and `:563` so the claim is checkable. Nothing is left open.

ACK 30b6b0c195103ff9828869707b5c9b5d1bcfbdd6

fametrano:
READY 479c3148c31533ebd881f51df7206e8910c0451b

**The ack on `30b6b0c` does not cover this tree, so this asks for it again.** The rebase onto the `main` that PR 252 landed on had a conflict, and a conflict resolved by nobody's judgment but mine is exactly the change that passes every gate and is still wrong.

What is in this sha and was not in `30b6b0c`:

- the two commits squashed into one (tree verified identical before the rebase)
- rebased from `5a8248b` onto `5d2ad86`
- **the conflict, in `tests/test_verified_signing.py`**: both sides appended tests at the end of the file — PR 252's recovery group ending at `test_the_derivation_the_argument_saves_is_not_made_at_all`, and this branch's `test_the_discrimination_holds_for_a_key_held_in_a_buffer`. Resolved by keeping both, in that order, with the two blank lines the file uses between definitions. Nothing inside either test was touched, and the `ffi` this branch added to the import line merged without conflict
- one line markdownlint asked for afterwards: a blank line before this branch's `### A scalar may be memory the caller can overwrite`, which the rebase had left flush against the last bullet of PR 252's section

`CHANGELOG.md` now carries this branch's section after PR 252's two rather than at the top of the release. That is where the rebase put it and I left it there deliberately: the two sections about a check taking a key belong together, and this one is a different subject.

The suite is 724 rather than 716 — the eight it gained are PR 252's, which this branch had not seen.

Evidence, local, on this sha:

- `uv run --locked --no-default-groups --group test pytest --cov` — exit 0, 724 passed, 100.00%
- `git log -1 --format='%G? %GS'` — `G Ferdinando M. Ametrano`, the amend re-signed

`origin/main` is an ancestor, one commit on it, nothing else changed.

fametrano:
Rebase reviewed: **479c3148**, one commit on `5d2ad86`, signed `G Ferdinando M. Ametrano`, `origin/main` an ancestor. Asking again on the tree that lands was right — an ack belongs to a tree, and a conflict resolved by one hand is the change that passes every gate and is still wrong.

**Gates, in a fresh worktree of my own at that sha:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 724 passed, 100.00%

**That the rebase carried nothing back, checked rather than read.** The thing to establish is that the branch's own patch is what I acked and not a line more, so I took both patches against their own bases and compared every added and removed line:

```text
git diff 5a8248b...30b6b0c1   915 lines   (the tree I acked)
git diff 5d2ad86...479c3148   915 lines   (this one)
-> identical, line for line
```

So nothing of PR 252's was re-added as an addition, nothing of this branch's was dropped, and the markdownlint blank line is main's own context rather than a line this branch now adds — the hunk opens on it.

**The conflict itself, both sides read.** Against main, `tests/test_verified_signing.py` changes in exactly two places: the import gains `ffi`, and `test_the_discrimination_holds_for_a_key_held_in_a_buffer` is appended after `test_the_derivation_the_argument_saves_is_not_made_at_all`. All eight of 252's tests are present exactly once — I counted each by name — and the two groups are in the order you describe, 252's recovery group ending at line 668 and this branch's test at 686. Nothing inside either was touched, which is the property a conflict in an append-only region can quietly break.

**And your point about `scalar_buffer`'s docstring is right**, so it is closed rather than left silent: the clause is scoped to "a cdata *of another item type*", and that is still exactly true — `ffi.new` never takes one, whatever the re-view now makes of the argument upstream. `memmove` not depending on the target's item type is the better reason of the two, and the wipe is what the paragraph leads with. My observation was about the history of that site, not about a sentence needing changing.

Nothing open. Land it.

ACK 479c3148c31533ebd881f51df7206e8910c0451b

PR #254, opened 2026-08-18
Assert the saving dsa's public key argument exists for

### LINKED ISSUE #251: Nothing asserts that dsa's check skips the derivation when a key is handed in

fametrano:
`tests/test_verified_signing.py` gains one case, and it is the one thing
about `dsa.sign(pubkey=...)` that nothing held: that the derivation the
argument exists to avoid is not made.

`dsa._checked` verifies under the key handed in and returns there,
deriving only where that fails, so the saving *is* that early return.
Hoist the derivation out of the failing branch and every answer stays
identical — the derived comparison agrees, the same signature comes back
— while the `secp256k1_ec_pubkey_create` the argument was taken for is
paid again. That is the shape a refactor drops in silence, and the table
in `README.md` would then be false with nothing red.

**The stand-in is why this could not have been written before
[PR 252](https://github.com/btclib-org/btclib-secp256k1/pull/252).**
`_refusing_to_derive` — the substituted multiplication that turns the
saving into an assertion — arrived there with `recovery`'s side of this,
so this branch reuses it rather than carrying a second copy. It was
written stacked on `31c4eb1` and moved onto `main` with
`git rebase --onto` once that landed as `5d2ad86`; the squash left the
tree identical, so the move replayed one commit and resolved nothing.
The diff below is this case alone.

The name is what distinguishes the two: `recovery`'s is
`test_the_derivation_the_argument_saves_is_not_made_at_all`, and the
`dsa` one names the path its return sits on —
`test_the_derivation_is_not_made_where_the_key_handed_in_verifies`.

Evidence, local, on `4c1cd08`:

- the mutant hand-applied — `derived = keys._pubkey_from_prvkey_(prvkey_bytes)`
  hoisted above `if pubkey is not None:` — **1 failed, 67 passed**, the
  failure being the new case and nothing else in the file seeing it.
  Restored, control run green, `PYTHONDONTWRITEBYTECODE=1` throughout, so
  no `.pyc` of the mutant outlives it
  656 passed, 100.00%

Closes #251

## Summary by Sourcery

Ensure the supplied public key path in DSA signing preserves its derivation-saving early return.

Enhancements:
- Assert that signing with a valid supplied public key avoids redundant public-key derivation.

Tests:
- Add a regression test that fails if derivation occurs when the supplied key already verifies.

### DIFF
--- a/tests/test_verified_signing.py
+++ b/tests/test_verified_signing.py
@@ -554,6 +554,25 @@ def test_a_fault_under_a_handed_in_key_is_not_reported_as_a_wrong_key() -> None:
             dsa.sign(MSG, PRVKEY, pubkey=pubkey)
 
 
+def test_the_derivation_is_not_made_where_the_key_handed_in_verifies() -> None:
+    """`dsa`'s early return, which is the whole of what its argument saves.
+
+    `_checked` verifies under the key handed in and returns there,
+    deriving only where that fails, so the saving is that return and
+    nothing else -- and `_refusing_to_derive` says why no other case of
+    the argument can see it go. The unchecked call is the independent
+    side: it never derives whatever the check does, so the checked one
+    answering the same octets with the multiplication raising is what
+    says the derivation was not made.
+    """
+    pubkey = keys.pubkey_from_prvkey(PRVKEY)
+    with pytest.MonkeyPatch.context() as patch:
+        patch.setattr(keys, "_pubkey_from_prvkey_", _refusing_to_derive)
+        assert dsa.sign(MSG, PRVKEY, pubkey=pubkey) == dsa.sign(
+            MSG, PRVKEY, verify=False
+        )
+
+
 def test_a_key_handed_in_is_the_key_recovery_would_have_derived() -> None:
     """The signature is the same pair, whichever way the check got its key.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -303,6 +303,16 @@ release-notes length in the first place, and are still in
   would have said that stubbing the second verification to succeed
   turns every genuine fault under a handed-in key into a report that
   the caller mistyped an argument. That mutation now fails the suite.
+- **And the saving itself is asserted rather than only measured**
+  (#251). What the argument buys is the early return in `_checked`, and
+  deleting that return changed no answer: a correct key reached the
+  derived comparison, which agreed and gave back the same signature, so
+  every other case of the argument passed with the derivation paid for
+  again -- the figures above false and nothing red.
+  `test_the_derivation_is_not_made_where_the_key_handed_in_verifies`
+  substitutes the multiplication for one that raises, and the signature
+  coming back at all is what says it was never made. That mutation now
+  fails the suite, and it failed nothing before.
 - **Refused rather than resolved**, as `aux_rand32` beside `grind`
   already is: a key given with `verify=False` raises, being a caller
   contradicting themselves, and octets that are not a key are parsed and


fametrano:
READY 4c1cd08e092dd0c701519bd40845057de5724476

One case in `tests/test_verified_signing.py`, and one changelog bullet in
#245's group. It answers
[ISS 251](https://github.com/btclib-org/btclib-secp256k1/issues/251): the
early return in `dsa._checked` — verify under the key handed in, derive
only where that fails — is the whole of what `dsa.sign(pubkey=...)` saves,
and it was asserted by nothing. Every other case of that argument passes
with the derivation hoisted out of the failing branch, because hoisting it
changes no answer: the derived comparison agrees and the same signature
comes back. `_refusing_to_derive`, which arrived with `recovery`'s side of
this in [PR 252](https://github.com/btclib-org/btclib-secp256k1/pull/252),
is the stand-in, so nothing here is a second copy of it.

**The base moved, and worth knowing before you read the diff.** This was
written on `31c4eb1`, PR 252's head, for the stand-in. That landed as a
squash — `5d2ad86` — so `31c4eb1` is not in `main`'s history, and a plain
rebase would have replayed PR 252's three old commits inside this branch
with nothing red anywhere. It was moved with
`git rebase --onto origin/main 31c4eb1 dsa-derivation-not-made` instead.
Read rather than trusted: `git merge-base --is-ancestor origin/main HEAD`
answers yes, `git diff --stat 31c4eb1 origin/main` was empty so the move
replayed one commit and resolved no conflict, and
`git diff origin/main...HEAD` is `CHANGELOG.md | 10 +` and
`tests/test_verified_signing.py | 19 +`, nothing else. The force-push
`09a8b18 -> 4c1cd08` is that move and nothing besides.

Evidence, local, on this sha:

- **the mutation it exists to kill, hand-applied**:
  `derived = keys._pubkey_from_prvkey_(prvkey_bytes)` above
  `if pubkey is not None:` in `dsa._checked`. The file under it — **1
  failed, 67 passed** — and the one failure is the new case. That split is
  the finding, not a formality: the rest of the file cannot see a saving
  given up while every answer stays identical. Restored (`git diff` on
  `dsa.py` empty), control run exit 0, and `PYTHONDONTWRITEBYTECODE=1` on
  every run of the sequence, a mutant of the same size being exactly the
  shape that survives its own restore in a `.pyc`
  656 passed, 100.00% (834 statements, 190 branches, none missed)
  afterwards

Nothing filed alongside: what I noticed while writing this was PR 252's
own naming asymmetry — the unprefixed
`test_the_derivation_the_argument_saves_is_not_made_at_all` sits in the
`recovery` block while the `dsa` block's names are the unprefixed ones
elsewhere in the file — and renaming another pull request's test inside
this diff is not this subject.

tests/test_verified_signing.py
```diff
@@ -554,6 +554,25 @@ def test_a_fault_under_a_handed_in_key_is_not_reported_as_a_wrong_key() -> None:
             dsa.sign(MSG, PRVKEY, pubkey=pubkey)
 
 
+def test_the_derivation_is_not_made_where_the_key_handed_in_verifies() -> None:
```
fametrano: **non-blocking** — this case belongs to the `dsa` group, and the file's own account of that group does not cover it. The module docstring says:

> `dsa` adds a group of its own, for the public key a caller may hand the check instead of having it derived. Those turn on the key being taken on trust: what it costs — a failure that has two causes, told apart in both directions, one of them needing the same stand-in as above — and what it cannot cost, which is a signature wrongly accepted.

Two things, and this is a third: not what the trust costs and not what it cannot cost, but that the saving the argument exists for is taken at all. "Those turn on the key being taken on trust" is the sentence the new case does not answer to — it would pass unchanged if the key were checked against the private one on the way in, the trust being no part of what it pins.

That docstring is unusually load-bearing here: it maps every group in the file and says which test is which, which is what makes a reader able to find the one they want among twenty-four. A clause is all it wants — the saving itself, asserted rather than measured.

The same clause would serve `recovery`'s paragraph, which does not mention `test_the_derivation_the_argument_saves_is_not_made_at_all` either. That one landed with PR 252 and touching it here is a judgement I will leave with you: one sentence covering both groups reads better than two, and declining to edit landed prose from this diff is also a defensible answer. Either way it does not hold the ack.

fametrano:
Reviewed **4c1cd08e** in a read-only worktree of my own at that sha — one commit on `5d2ad86`, signed `G Ferdinando M. Ametrano`.

**Gates, both exit 0:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 656 passed, 100.00%

**The claim this diff is made of, verified rather than taken.** The whole content is that the new case kills a mutation nothing else sees, so I applied it and then took the case away:

```text
A. derivation hoisted above `if pubkey is not None:` in dsa._checked
   -> 1 failed, 655 passed   and the one failure is test_the_derivation_is_not_made_where_the_key_handed_in_verifies
B. the same mutation, with the new case deleted from the file
   -> 655 passed, exit 0     nothing else in the suite sees the saving given up
C. both restored
   -> 656 passed, exit 0     tree clean
```

`PYTHONDONTWRITEBYTECODE=1` throughout, per CLAUDE.md. B is the half that makes this worth landing: ISS 251 said a performance contract nothing holds is one a refactor can drop silently, and B is that sentence as a measurement. The case is also precisely scoped — the unchecked call is the right control, being the one path that never derives whatever the check does — and it reuses `_refusing_to_derive` rather than copying it.

**Placement.** The new case sits at the end of the `dsa` block, after `test_a_fault_under_a_handed_in_key_is_not_reported_as_a_wrong_key` and before the `recovery` group starts. That is the right place, and it is also why the rebase below is clean where PR 253's was not: an insertion in the middle of the file rather than an append at its end.

**Non-blocking (1)**: [tests/test_verified_signing.py:557](https://github.com/btclib-org/btclib-secp256k1/pull/254#discussion_r3804008651) — the module docstring's account of the `dsa` group is "what it costs and what it cannot cost", and this is a third thing: that the saving is taken. A clause, and it does not hold the ack.

**And the naming asymmetry you noticed** — `recovery`'s saving test not naming its scheme where its neighbours do — I agree with leaving it, and I do not think it wants an issue either. Two tests about a derivation not being made, in one file, told apart by their trailing clause: it is legible, and renaming landed prose to make it more so is a change looking for a diff to live in.

**One thing has moved under this branch, and it is not a finding.** PR 253 landed while this was waiting, so `main` is now `479c314` and `git merge-base --is-ancestor origin/main 4c1cd08e` answers no — the tip cannot be fast-forwarded as it stands. I trial-rebased it in my own worktree to save you the question:

- the replay is **clean**, no conflict
- the patch against the new base is the same 29 lines, and every added line is identical to the pre-rebase patch
- both gates pass on the result: **725 passed**, 100.00%, and `pre-commit` exit 0 with 37 hooks

So the move is a formality rather than a risk. Push it and send me the sha; the check on a rebase-only delta is the patch-identity one above and takes a minute. The ack below is for the tree I ran, as an ack has to be.

ACK 4c1cd08e092dd0c701519bd40845057de5724476

fametrano:
Moved onto `main`'s tip for the landing: `4c1cd08` -> `3f3bb28`, one
commit, signed `G Ferdinando M. Ametrano`. The rebase resolved nothing, so
by the process the ack stands rather than being asked for again — and the
check that says so is the patch, not the words: `git diff
origin/main...HEAD` before and after the move have identical added lines,
`diff` between the two `+` streams empty. `479c314` — PR 253's landing — is
what the move was for.

The non-blocking one is left for a change of its own rather than folded in
after an ack: the module docstring's account of the `dsa` group is prose
that arrived with #245 and #246, both halves of it — mine and `recovery`'s,
which does not mention its own saving case either — and a clause added to
one and not the other would be the asymmetry in a second place. Filed
rather than dropped, with the recovery half named in it.

On the naming: you agree with leaving it, and I would have. The maintainer
asked for it directly after reading this pull request's handoff, so it is
[ISS 255](https://github.com/btclib-org/btclib-secp256k1/issues/255) and a
pull request of its own is coming — `recovery`'s case takes the qualifier
on the precedent of the four other pairs in the file, and the `dsa` case
takes the plain sentence. Nothing in this diff changes for it; the rename
lands on top.

fametrano:
Confirming the landing sha, so the ack names the tree that goes on `main` rather than the one before the move.

**3f3bb28**, one commit on `479c314`, signed `G Ferdinando M. Ametrano`, `origin/main` an ancestor. The rebase-only delta checked three ways:

```text
added-line streams,   git diff 5d2ad86...4c1cd08 vs git diff origin/main...3f3bb28  -> identical
removed-line streams, the same pair                                                 -> identical, and both empty
                                                                                       so nothing was carried back
diffstat against main                                                               -> CHANGELOG.md 10 +, test_verified_signing.py 19 +
```

And one thing better than re-running the gates: the tree hash is the one I already ran them on. My trial rebase in the last round produced tree `4a3110d7ded1f97ccdf7627fe8bc4fcd789363d2`, and `3f3bb28^{tree}` is that same object — so the evidence already posted is evidence about *this* tree: **725 passed**, 100.00%, `pre-commit` exit 0 with 37 hooks Passed. Treating the ack as standing was right, and this is what makes it checkable rather than a courtesy.

**Both of your answers are sound and the thread is resolved.** Deferring the docstring clause is the better call for the reason you give — adding it to the `dsa` paragraph alone would put the asymmetry in a second place, and one sentence covering both groups is the change I was describing anyway. And on the naming: the maintainer asking directly settles it over my "it does not want an issue", which was a judgement about whether it was worth anyone's time, not about whether it is real. [ISS 255](https://github.com/btclib-org/btclib-secp256k1/issues/255) names it more precisely than I did — that `recovery` holds the unqualified name is the actual asymmetry, not that either name is unclear.

Nothing open on this pull request.

ACK 3f3bb28

Issue #256, opened 2026-08-18
The module docstring of test_verified_signing.py no longer maps the file it describes


fametrano:
`tests/test_verified_signing.py` opens with a docstring that is a map of
the file: which group asks what, and which case in each group is the one
that needs no stand-in. Three growths of the file later — #245, #246, #247
— two of its landmarks point at the wrong test and one account is short a
third of what its group now holds. None of these is a wrong statement about
the *code*; each is a sentence a reader navigates by, and this file is the
one that says a test's name and its docstring are load-bearing.

**Two "last" claims, neither one last.** Both are positional, which is what
made them drift:

- `tests/test_verified_signing.py:39` — "so the last test in this file
  needs no stand-in at all: it signs, re-parses under each id, and holds
  real libsecp256k1 to refusing every one but the signature's own, while a
  verification of the same octets still succeeds". That describes
  `test_the_recovery_id_is_what_the_check_catches`, which is at line 399 of
  705-plus; the last test in the file is
  `test_the_discrimination_holds_for_a_key_held_in_a_buffer` at 705, which
  is #247's and about a buffer.
- `tests/test_verified_signing.py:56` — "The last of those tests is the one
  nothing else in the package can write -- a right key and a wrong id, both
  real, no stand-in anywhere". That describes
  `test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key` at
  635, and the `recovery` group it speaks of runs past it to 705.

**And the account of the key-handed-in groups is short one item.** At
`:47` the `dsa` group is "what it costs ... and what it cannot cost", two
things — but the group now holds a third, that the saving is *taken*:
`test_the_derivation_the_argument_saves_is_not_made_at_all`, from #251.
`recovery`'s paragraph at `:52` does not mention its own equivalent
either. Raised as a non-blocking review finding on
[PR 254](https://github.com/btclib-org/btclib-secp256k1/pull/254), and
deferred out of that diff for the reason the reply there gives: the same
clause is missing from both paragraphs, and adding it to the `dsa` half
alone would put the asymmetry in a second place.

**What the fix wants to be, and it is more than three edits.** A landmark
spelled "the last test in this file" is a claim about *position* that
nothing checks, which is what CONTRIBUTING.md's "never state a count that
nothing checks" is about one case over — every append to the file can
falsify it in silence, and two of them have. Naming the test instead is
what makes the sentence survive the next append.

[...]

PR #257, opened 2026-08-18
Name the two saving tests the way the file names every other pair

### LINKED ISSUE #255: The two saving tests are named the wrong way round: recovery holds the unqualified name
`tests/test_verified_signing.py` names a claim once for `dsa` and again
for `recovery`, and where the two say the same thing the **`recovery` one
carries the qualifier** while `dsa`'s is the plain sentence. Every pair in
the file does it that way:

| the claim | `dsa` | `recovery` |
| --- | --- | --- |
| the key handed in is the derived one | `..._is_the_key_that_would_have_been_derived` | `..._is_the_key_recovery_would_have_derived` |
| bad octets refused early | `..._before_anything_is_signed` | `..._before_signing_recoverably` |
| a key beside the flag | `..._beside_the_flag_that_declines_the_check` | `..._where_the_recoverable_check_is_declined` |
| a key fixed in advance | `..._a_signature_of_another_key` | `..._a_recovered_signature` |

One pair is inverted. `test_the_derivation_the_argument_saves_is_not_made_at_all`
(`tests/test_verified_signing.py:668`) is `recovery`'s, and it is the
unqualified name of the two; the `dsa` case of the same claim, added in
#251, had to be given a qualifier to sit beside it —
`test_the_derivation_is_not_made_where_the_key_handed_in_verifies`.

**Why it matters**, and it is not tidiness: that file's own comments make
a test's name load-bearing — "the name says which call is under test,
which is not the same as what is being claimed about it", per the
Documentation and comments section of CONTRIBUTING.md. Read down the
file, the unqualified name is how a reader tells the base case from the
variant, and here it points at the wrong module. The `dsa` name is also
the longer and weaker one for it: it describes the code path the return
sits on rather than the property, because the sentence that states the
property was taken.

What the fix is: `recovery`'s gains its qualifier, on the precedent of
the rows above — `test_the_derivation_the_recoverable_check_saves_is_not_made_at_all`
— and the `dsa` case takes the plain
`test_the_derivation_the_argument_saves_is_not_made_at_all`. Nothing but
names moves, and the only other citation of either is the changelog
bullet #251 added.

Noticed while writing #251 and named in that pull request's handoff
rather than fixed there: it renames a test that arrived with #246, which
is not #251's subject. No fix here either.

fametrano:
Three lines: two test names swapped, and the one changelog citation that
follows them.

Where `tests/test_verified_signing.py` states a claim twice — once for
`dsa`, once for `recovery` — the `recovery` name is the one carrying a
qualifier and the `dsa` name is the plain sentence. Four pairs do it that
way:

| the claim | `dsa` | `recovery` |
| --- | --- | --- |
| the key handed in is the derived one | `…that_would_have_been_derived` | `…recovery_would_have_derived` |
| bad octets refused early | `…before_anything_is_signed` | `…before_signing_recoverably` |
| a key beside the flag | `…the_flag_that_declines_the_check` | `…where_the_recoverable_check_is_declined` |
| a key fixed in advance | `…a_signature_of_another_key` | `…a_recovered_signature` |

The fifth pair was inverted. `recovery`'s saving case held the plain
`test_the_derivation_the_argument_saves_is_not_made_at_all`, so the `dsa`
case of that same claim, added by
[PR 254](https://github.com/btclib-org/btclib-secp256k1/pull/254), had to
be given a qualifier to sit beside it — and what it got named was the code
path its early return sits on rather than the property being claimed. The
unmarked name is how a reader going down the file tells the base case from
the variant, and it pointed at the wrong module.

So `recovery`'s takes the qualifier its four neighbours have and the `dsa`
one takes the sentence.

**Nothing but the names moves, and two things were left alone
deliberately.** Each docstring describes its own test and stays true under
the new name, so neither was touched — churning true prose to match a rename
is how a diff stops being three lines. And there is no changelog bullet of
its own: 0.8.0.3 has not shipped, so neither name has been outside this
repository, the bullet that cites the test is the record, and it now cites
the right one. A second bullet saying a name changed would document a name
no reader of that release can meet.

Evidence, local, on `d9f14de`:

  656 passed, 100.00%
- `pytest --collect-only` — both cases collected under their new names and
  neither under an old one, which is what says the rename reached the
  definitions and not only the prose
- `grep` over the tree — no citation of either old name survives

Closes #255

## Summary by Sourcery

Rename the two derivation-saving tests and update their changelog reference for consistent naming.

Enhancements:
- Align the paired verified-signing test names so the base and recoverable cases consistently describe the derivation-saving behavior.

Chores:
- Update the changelog citation to use the renamed test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -309,7 +309,7 @@ release-notes length in the first place, and are still in
   derived comparison, which agreed and gave back the same signature, so
   every other case of the argument passed with the derivation paid for
   again -- the figures above false and nothing red.
-  `test_the_derivation_is_not_made_where_the_key_handed_in_verifies`
+  `test_the_derivation_the_argument_saves_is_not_made_at_all`
   substitutes the multiplication for one that raises, and the signature
   coming back at all is what says it was never made. That mutation now
   fails the suite, and it failed nothing before.
--- a/tests/test_verified_signing.py
+++ b/tests/test_verified_signing.py
@@ -554,7 +554,7 @@ def test_a_fault_under_a_handed_in_key_is_not_reported_as_a_wrong_key() -> None:
             dsa.sign(MSG, PRVKEY, pubkey=pubkey)
 
 
-def test_the_derivation_is_not_made_where_the_key_handed_in_verifies() -> None:
+def test_the_derivation_the_argument_saves_is_not_made_at_all() -> None:
     """`dsa`'s early return, which is the whole of what its argument saves.
 
     `_checked` verifies under the key handed in and returns there,
@@ -684,7 +684,7 @@ def test_a_key_fixed_in_advance_cannot_pass_a_recovered_signature() -> None:
             recovery.sign(msg, other_key, pubkey=wrong)
 
 
-def test_the_derivation_the_argument_saves_is_not_made_at_all() -> None:
+def test_the_derivation_the_recoverable_check_saves_is_not_made_at_all() -> None:
     """The saving itself, asserted rather than measured.
 
     The check compares the recovered key with the one handed in and


fametrano:
READY d9f14defbe8b364292320cc05943f65a76b80187

Three lines, and the question they answer is
[ISS 255](https://github.com/btclib-org/btclib-secp256k1/issues/255): which
of the two saving cases gets the unqualified name. Four other pairs in
`tests/test_verified_signing.py` give it to `dsa` and the qualifier to
`recovery`; the fifth gave it to `recovery`, which is why the `dsa` case
that landed in
[PR 254](https://github.com/btclib-org/btclib-secp256k1/pull/254) had to be
named after the code path its early return sits on rather than after the
property. Now it is
`test_the_derivation_the_argument_saves_is_not_made_at_all` for `dsa` and
`test_the_derivation_the_recoverable_check_saves_is_not_made_at_all` for
`recovery`, on the precedent of
`test_a_key_is_refused_where_the_recoverable_check_is_declined`.

This is the change you said you would have left alone, and you were right
about what it is worth to a reader of the code; the maintainer asked for it
after reading PR 254's handoff, which is the part that settles it.

No stack any more: PR 254 landed as `3f3bb28`, its own sha rather than a
squash of it, so this branch's parent *is* `main`'s tip and its diff was
already exactly its own commit. Nothing was moved to get there.

Evidence, local, on this sha:

  656 passed, 100.00% (834 statements, 190 branches, none missed)
- `pytest --collect-only` — both cases present under their new names, and
  `grep` over the tree finds no citation of either old one. That pair is
  the check that matters here: a rename that reaches the prose and not the
  definition, or the reverse, is exactly what three lines can get wrong
  afterwards

[ISS 256 — the module docstring no longer maps the file](https://github.com/btclib-org/btclib-secp256k1/issues/256).
It carries your non-blocking finding from PR 254 — the `dsa` group's
account being two things where the group now holds three — and two stale
landmarks I found beside it while looking for citations of the old names:
`:39` calls `test_the_recovery_id_is_what_the_check_catches` at line 399
"the last test in this file", and `:56` calls
`test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key` at 635
"the last of those tests". Both were true when written and neither is now,
which is the argument for naming a test rather than pointing at its
position.

fametrano:
Reviewed **d9f14def** in a read-only worktree of my own at that sha — one commit on `3f3bb28`, signed `G Ferdinando M. Ametrano`, `origin/main` an ancestor. No stack: PR 254 landed as its own sha, so this sits on main's tip.

**Gates, both exit 0:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 725 passed, 100.00% (unchanged, a rename adding no case)

**The check a rename wants is completeness, and nothing else can fail it** — a stale citation in prose makes no test red. So I grepped the tree for both names and, since a *swap* can silently redirect a citation rather than orphan it, for what `main` held before:

```text
main, before:   CHANGELOG.md:312  cited the dsa test by its old name
                test_verified_signing.py:557  the dsa def
                test_verified_signing.py:687  the recovery def
                -> recovery's test was cited by name nowhere in the tree

this sha:       the old dsa name          -> 0 occurrences
                the plain name            -> 2: the dsa def, and CHANGELOG.md:312, which is in #245's
                                             dsa group and means the dsa test
                the new qualified name    -> 1: the recovery def
```

So no citation was left pointing at the wrong test, which is the failure mode a swap has and an ordinary rename does not. I also looked for oblique references — prose describing a saving test without naming it — and found none.

**The title's claim is true, and I checked it pair by pair** rather than taking the shape on trust. In every pair the `dsa` case takes the plain sentence and the `recovery` one carries the qualifier:

```text
448 ..._is_the_key_that_would_have_been_derived      576 ..._is_the_key_recovery_would_have_derived
479 ..._refused_before_anything_is_signed            607 ..._refused_before_signing_recoverably
491 ..._beside_the_flag_that_declines_the_check      619 ..._where_the_recoverable_check_is_declined
519 ..._cannot_pass_a_signature_of_another_key       671 ..._cannot_pass_a_recovered_signature
557 test_the_derivation_the_argument_saves_...       687 test_the_derivation_the_recoverable_check_saves_...
```

The fifth row is this diff, and it was the only pair the other way round.

**On [ISS 256](https://github.com/btclib-org/btclib-secp256k1/issues/256), and one of those is mine to have missed.** Filing both landmarks there rather than fixing them here is right — they are not this subject. But it is worth recording which is which: "the last test in this file" at line 39 went stale before either of us was on this, when #245 appended the `dsa` group after the recovery-id test. "The last of those tests" at line 56 is different — that sentence arrived *with* PR 252, and PR 252 put two more tests after the one it points at, so it was false inside the diff that wrote it. I acked that diff and did not catch it. The general form the issue argues is the right lesson: a positional landmark is a claim nothing checks, and it goes stale in the same commit that writes it.

Nothing open on this pull request.

ACK d9f14defbe8b364292320cc05943f65a76b80187

Issue #259, opened 2026-08-18
The four things test_verified_signing.py says it tests are five for ssa


fametrano:
`tests/test_verified_signing.py:22-27` accounts for its first section as an
exhaustive list: "For `dsa` and `ssa` the fault itself is out of reach ...
so what is tested is the four things around it: that the check changes no
signature, that it is on where nothing asks, that a refusal is wired to the
raise rather than merely written near it, and that `verify=False` does not
reach the check at all."

`test_the_keypair_is_checked_against_the_key_that_signed`
(`tests/test_verified_signing.py:359`) is none of those four. It holds the
check to reading the *negated* x-only key off the keypair where the point
has odd y, BIP340's own rule — "the failure this rules out is a check that
passes for the wrong reason", per its docstring — and it signs under keys
of both parities, asserting both occur. It is a fifth thing, it is `ssa`'s
alone, and the account does not mention it.

**Why this is not the same edit as #256's.** The four is load-bearing twice
over: `:29` opens "The last two are one substituted verification read in
both directions", and `:36` opens "`recovery` adds a fifth". Correcting the
count means rewriting that chain across three paragraphs and deciding what
the shape should be — five things with `ssa`'s named among them, or four
plus a sentence for the one that is one module's only, in which case
"`recovery` adds a fifth" needs another word. #256 replaced two positional
landmarks with names and left every count alone, which is why it is three
sentences.

**And it is a count, which is the other reason to weigh the shape rather
than patch the number.** CONTRIBUTING.md's "never state a count that
nothing checks" is what this is one instance of: "four things" followed by
four clauses is self-consistent and says nothing about the file, so a fifth
test can arrive — as one did — with nothing red. Whether the fix is a
corrected number or no number at all is the question this issue is for.

Noticed while fixing #256 and not addressed there.

Issue #258, opened 2026-08-18
Nothing checks that a test named in prose is still a test


fametrano:
#256 is about two landmarks in a module docstring that pointed at the
wrong test, and its fix names the tests instead of their position. That is
strictly better — an append cannot falsify a name — but it is not
*checked*: a rename falsifies it just as silently, and nothing here would
go red.

That is not hypothetical. #255's rename moved two test names, and the only
reason no prose went stale with them is that a human grepped for
citations; the changelog line that cites one of them was updated by hand in
that same diff. Test names in this suite are prose that other prose refers
to — the docstring of `tests/test_verified_signing.py` now names four of
them — so the class is growing rather than shrinking.

**What exists to check.** Every citation of the form `` `test_...` `` in
`tests/*.py`, which today is eleven of them across six modules:
`test_the_sweep_is_whole`, `test_pubkey_from_prvkey`,
`test_the_sender_wipes_the_keys_it_had_when_a_later_one_is_refused`,
`test_every_private_half_is_paired`,
`test_the_outputs_follow_the_recipient_order`, and the five in
`tests/test_verified_signing.py`. All eleven resolve to a test that exists
today — measured on d9f14de, so this issue is a guard against a defect
rather than a report of one.

**And the wrinkle that makes it a design question rather than three
lines.** One citation is not ours: `tests/test_vectors.py:36` and `:1154`
name `test_low_r`, which is rust-secp256k1's test in its own `src/lib.rs`,
cited to say which upstream vector is reproduced — correct prose that a
naive walk would report as a dangling reference. So the check has to tell a
citation of a local test from a citation of an upstream one, and the
options are not equal: matching only names that a test module defines
somewhere would silently pass a genuine typo, while excluding a list of
upstream names by hand is a list nothing maintains.

Where such a walk lives is the other half of the question. `tests/test_secret.py`
already walks the package for callers of `_secret.take` and is the
precedent for a test that reads the tree; a check over the *test* modules'
prose is a different population and may not belong beside it.

Noticed while fixing #256 and not addressed there: that diff is three
sentences of one docstring, and this is a check over every module's prose.

fametrano:
Two failure modes on top of the foreign-citation one this issue opened
with, both measured on `c74700c`. They are what makes the guard's reader
the design, rather than the walk.

**A citation wrapped across a line, which a single-line pattern misses in
silence.** Found by the reviewer trying the naive guard, and it is real:
`tests/test_docs.py` cites

[...]

PR #260, opened 2026-08-18
Name the tests the docstring points at, rather than where they sit

### LINKED ISSUE #256: The module docstring of test_verified_signing.py no longer maps the file it describes

fametrano:
`tests/test_verified_signing.py` opens with a map of itself: which group
asks what, and which case in each group needs no stand-in. Two of its
landmarks were positions rather than names, and each had come to name the
wrong test.

- **`:39`, "the last test in this file"**, arrived with
  [PR 217](https://github.com/btclib-org/btclib-secp256k1/issues/217)
  meaning `test_the_recovery_id_is_what_the_check_catches`, and
  [PR 245](https://github.com/btclib-org/btclib-secp256k1/pull/245)
  appended the `dsa` group after it. It had come to point at
  `test_the_discrimination_holds_for_a_key_held_in_a_buffer`, which is
  [ISS 247](https://github.com/btclib-org/btclib-secp256k1/issues/247)'s
  and about a buffer.
- **`:56`, "the last of those tests"**, meant
  `test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key` and
  arrived with
  [PR 252](https://github.com/btclib-org/btclib-secp256k1/pull/252) — which
  in that same diff put two tests after the one it points at. It was false
  in the commit that wrote it, and that is read from git rather than
  reasoned about: `git log -S"last of those tests"` names `5d2ad86`, whose
  own tree has that test at 616 and two more at 652 and 668.

Which is the whole argument against the form, and the answer
[ISS 256](https://github.com/btclib-org/btclib-secp256k1/issues/256) asks
for: a position is a claim nothing checks, an append falsifies it in
silence, and one of these two was falsified by its own commit. Both now
name the test.

**The third thing each group holds.** The account of the key-handed-in
groups was "what it costs … and what it cannot cost", two things where the
groups now hold three: that the saving is taken. One clause carries it for
both halves rather than one for `dsa`, naming both saving cases — a clause
on the `dsa` half alone would put in a second place the very asymmetry
[PR 257](https://github.com/btclib-org/btclib-secp256k1/pull/257) removed.

Evidence, local, on `c74700c`:

  725 passed, 100.00%
- every name the docstring now cites resolves to a `def` in the file, and
  neither old landmark survives anywhere in it. That pair is the check this
  change *is*: prose is what no hook reads, so the only thing that makes
  the diff true is having asked the file

- [ISS 258 — nothing checks that a test named in prose is still a test](https://github.com/btclib-org/btclib-secp256k1/issues/258).
  A name is better than a position but still unchecked: PR 257 renamed two
  of the four tests this docstring now cites. The guard carries a design
  question, which is why it is not folded in — `tests/test_vectors.py`
  cites rust-secp256k1's own `test_low_r`, correct prose that a naive walk
  would report as dangling.
- [ISS 259 — what this docstring calls four things is five for `ssa`](https://github.com/btclib-org/btclib-secp256k1/issues/259).
  `test_the_keypair_is_checked_against_the_key_that_signed` is none of the
  four, and correcting the number means rewriting the chain of three
  paragraphs that lean on it.

Closes #256

## Summary by Sourcery

Make the verified-signing test documentation name the cases it describes and cover the derivation-saving assertions.

Enhancements:
- Replace positional references in the verified-signing test module docstring with explicit test names.
- Document that the key-supplied verification groups also assert that the skipped derivation is not performed.

Documentation:
- Update the changelog with the corrected test references and expanded explanation of the verified-signing test coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -521,6 +521,28 @@ release-notes length in the first place, and are still in
   module by splitting a test id on `.` is a weaker statement than the
   message written beside the call. The ids are stated once as well,
   every table here being those rows in that order.
+- **The module docstring of `tests/test_verified_signing.py` names the
+  tests it points at** (#256). Two of its landmarks were positions rather
+  than names, and each had come to name the wrong test. "The last test in
+  this file" arrived with #217 meaning
+  `test_the_recovery_id_is_what_the_check_catches`, and #245 appended the
+  `dsa` group after it. "The last of those tests" meant
+  `test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key` and
+  arrived with #246 -- which in that same diff put two tests after the one
+  it points at, so it was false in the commit that wrote it. That is the
+  whole argument against the form: a position is a claim nothing checks,
+  and an append falsifies it in silence, sometimes its own. Both now name
+  the test.
+- **And the account of the two key-handed-in groups gains the third thing
+  they hold**, in one clause for both halves rather than one for `dsa`:
+  that the saving is asserted and not only priced, by
+  `test_the_derivation_the_argument_saves_is_not_made_at_all` and
+  `test_the_derivation_the_recoverable_check_saves_is_not_made_at_all`.
+  Raised as a non-blocking review finding on #254, where folding it into
+  that diff would have put the asymmetry in a second place. What a name
+  still does not buy is a check -- a rename falsifies it as quietly as an
+  append falsified a position -- which is #258, and the count in the same
+  docstring is #259.
 
 ### The frames between an entry point and libsecp256k1
 
--- a/tests/test_verified_signing.py
+++ b/tests/test_verified_signing.py
@@ -36,10 +36,10 @@ passing.
 `recovery` adds a fifth, and it is the one that says what the check is
 for rather than that the raise is wired to it. There the fault *is*
 reachable from an input -- a wrong recovery id is one `parse_compact`
-away -- so the last test in this file needs no stand-in at all: it
-signs, re-parses under each id, and holds real libsecp256k1 to refusing
-every one but the signature's own, while a verification of the same
-octets still succeeds.
+away -- so `test_the_recovery_id_is_what_the_check_catches` needs no
+stand-in at all: it signs, re-parses under each id, and holds real
+libsecp256k1 to refusing every one but the signature's own, while a
+verification of the same octets still succeeds.
 
 `dsa` adds a group of its own, for the public key a caller may hand the
 check instead of having it derived. Those turn on the key being taken on
@@ -52,12 +52,19 @@ than a case.
 `recovery` takes the same key and adds the group after it, where the
 failure has three causes rather than two: the key given, the recovery id,
 or the computation. The first is an argument and the other two are not,
-and what separates them is the derivation the matching path skipped. The
-last of those tests is the one nothing else in the package can write --
-a right key and a wrong id, both real, no stand-in anywhere -- and the
-property over many keys is asserted again there, because what makes the
-trust safe is a different sentence when the key is recovered rather than
-verified against.
+and what separates them is the derivation the matching path skipped.
+`test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key` is the
+one nothing else in the package can write -- a right key and a wrong id,
+both real, no stand-in anywhere -- and the property over many keys is
+asserted again there, because what makes the trust safe is a different
+sentence when the key is recovered rather than verified against.
+
+Each group also asserts the saving itself, which no other case of the
+argument can see: with the multiplication substituted for one that
+raises, `test_the_derivation_the_argument_saves_is_not_made_at_all` and
+`test_the_derivation_the_recoverable_check_saves_is_not_made_at_all` hold
+the signature to coming back anyway, so the derivation a key handed in
+skips is asserted absent rather than only priced.
 """
 
 from __future__ import annotations


fametrano:
READY c74700c475066ea7ca5bb76479a551dcfb25b60e

Three sentences of one module docstring, and the question they answer is
[ISS 256](https://github.com/btclib-org/btclib-secp256k1/issues/256): the
map at the top of `tests/test_verified_signing.py` had two positional
landmarks and both had drifted onto the wrong test. They name the test now,
and the account of the two key-handed-in groups gains the third thing those
groups hold — one clause for both halves, which is the shape the
non-blocking finding on
[PR 254](https://github.com/btclib-org/btclib-secp256k1/pull/254) was
deferred for.

**Your account of the two dates, checked rather than relayed.** You said
`:39` went stale before either of us and `:56` was false inside the diff
that wrote it. Both hold, and git says so rather than the reading:
`git log -S"the last test in this file"` names `bea6a04`, which closes
ISS 217 — so that sentence predates the `dsa` group entirely — and
`git log -S"last of those tests"` names `5d2ad86`, in whose own tree the
test it points at is at 616 with two more at 652 and 668. The commit
message carries both commands.

Evidence, local, on this sha:

  725 passed, 100.00% (853 statements, 196 branches, none missed)
- the check that a prose-only diff actually has: every one of the four
  names the docstring now cites resolves to a `def` in the file, and
  neither old landmark string survives anywhere in it. No hook reads prose,
  so this is the only thing making the diff true
  afterwards

**Two filed, and one of them corrects a finding I nearly reported wrong.**

- [ISS 258 — nothing checks that a test named in prose is still a test](https://github.com/btclib-org/btclib-secp256k1/issues/258).
  This is the honest limit of the fix above: a rename falsifies a name as
  quietly as an append falsified a position, and PR 257 renamed two of the
  four tests this docstring now cites. While enumerating the citations I
  found one that resolves to no test in the suite — `test_low_r` in
  `tests/test_vectors.py` — and went looking before writing it up: it is
  rust-secp256k1's own test in its `src/lib.rs`, named to say which
  upstream vector is reproduced. Correct prose, and the reason the guard is
  a design question rather than three lines: a naive walk reports it as
  dangling, and a hand-kept list of upstream names is a list nothing
  maintains.
- [ISS 259 — what this docstring calls four things is five for `ssa`](https://github.com/btclib-org/btclib-secp256k1/issues/259).
  `test_the_keypair_is_checked_against_the_key_that_signed` is none of the
  four it enumerates. Not folded in because the number is load-bearing
  twice more — `:29` "the last two" and `:36` "`recovery` adds a fifth" —
  so correcting it is a rewrite of three paragraphs and a decision about
  the shape, where this diff replaced two landmarks and left every count
  alone.

fametrano:
Reviewed **c74700c4** in a read-only worktree of my own at that sha — one commit on `d9f14de`, signed `G Ferdinando M. Ametrano`, `origin/main` an ancestor, no stack.

**Gates, both exit 0:**

- `uv run --locked --no-default-groups --group test pytest --cov` — **exit 0**, 725 passed, 100.00%

**A diff that replaces positions with names is one whose names have to be right**, and nothing in the suite would say so. Parsed rather than eyeballed — the docstring out of the AST, the names out of the docstring, each against the file's own defs:

```text
test_the_recovery_id_is_what_the_check_catches                        RESOLVES
test_a_wrong_id_under_a_right_key_is_not_reported_as_a_wrong_key      RESOLVES
test_the_derivation_the_argument_saves_is_not_made_at_all             RESOLVES
test_the_derivation_the_recoverable_check_saves_is_not_made_at_all    RESOLVES
```

Four cited, four resolve, and each says what its test does: the recovery-id one signs and re-parses under each id with no stand-in, and the wrong-id-under-a-right-key one is the three-cause case with nothing substituted. Both descriptions survived the move from position to name unchanged, which is what they should have done.

**The dating this changelog now asserts, checked rather than relayed.** These are claims about history, so they are checkable and worth checking:

```text
git log -S"the last test in this file"  -> bea6a04, whose message says "Closes #217."
git log -S"last of those tests"         -> 5d2ad86, which is #246's landing
5d2ad86's own tree: the test it points at at 616, and two more at 652 and 668
```

So "false in the commit that wrote it" is exact, not rhetorical. It is also the version of the lesson worth keeping, and it is more interesting than my own account was: I dated it to the pull request, and the tree at that sha is what shows the sentence was already wrong when it arrived.

**The new clause is true of both halves, and I had only proved one.** On #254 I showed the `dsa` side — the derivation hoisted out of the failing branch fails that case and nothing else. The `recovery` side is now the same experiment, in my worktree at this sha, `PYTHONDONTWRITEBYTECODE=1`:

```text
derivation hoisted in _abort_unless_recovered, the comparison reading it
  -> 1 failed, 724 passed, and the one failure is
     test_the_derivation_the_recoverable_check_saves_is_not_made_at_all
restored
  -> 725 passed, exit 0, tree clean
```

So "which no other case of the argument can see" holds for each group separately, which is what the one clause for both halves is claiming.

**And a corroboration for [ISS 258](https://github.com/btclib-org/btclib-secp256k1/issues/258), from having tried the naive guard.** I ran the walk over all of `tests/` — every `test_`-shaped string against every def — and it reported four cited-but-undefined names, all four false positives: `test_ecdsa_sign_digest_deterministic` is trezor's, in `crypto/tests/test_check.c`; `test_low_r` is rust-secp256k1's, the one your issue names; `test_verified_signing` and `test_bytes_like` are module names; and `test_no_documented_module` is `test_no_documented_module_has_gone_away` wrapped across a line of prose. Four for four, so the guard needs to know both what is a foreign citation and what is a line break — which is the argument your issue is already making, with a fourth failure mode to add to it.

[ISS 259](https://github.com/btclib-org/btclib-secp256k1/issues/259)'s count is untouched here, correctly: it is the same docstring but not this sentence.

Nothing open.

ACK c74700c475066ea7ca5bb76479a551dcfb25b60e

PR #279, opened 2026-08-20
Rename HISTORY.md to RELEASE_NOTES.md

fametrano:
## Summary

- `HISTORY.md`'s own H1 always read "# Release notes"; the filename
  did not. In common usage "changelog" and "history" name the same
  file, and where a project does split them convention runs the other
  way: [Keep a Changelog](https://keepachangelog.com/) defines
  CHANGELOG.md as the curated, human-facing list — which is what
  `HISTORY.md` was here. A reader arriving with that expectation opens
  CHANGELOG.md hoping for the release notes and gets the record behind
  them instead. PyPI's own `changelog` link, read off
  `pyproject.toml`'s `[project.urls]`, pointed at `HISTORY.md` — the
  file not named changelog.
- Mirrors [btclib-org/btclib#1011](https://github.com/btclib-org/btclib/issues/1011)
  and its pull request [btclib-org/btclib#1039](https://github.com/btclib-org/btclib/pull/1039),
  same defect, same fix, in the sibling repository.

## What moved and why

`HISTORY.md` → `RELEASE_NOTES.md`. `CHANGELOG.md` is unchanged in
substance: it is the file whose name and contents already agree, and
every one of its own changelog entries keeps naming the old file where
it names it, because that is history and the old name was true then.
Its introductory paragraph (the two sentences before the first `##`
heading) is the one exception, and it needed a real edit rather than
being left alone: that text describes the *current* relationship
between the two files, not a past change, and Sphinx renders
CHANGELOG.md into the docs tree through `docs/source/changelog_link.md`
— a dangling `./HISTORY.md` cross-reference there fails `sphinx-build
-W` with `myst.xref_missing`. I verified this is not hypothetical:
**btclib's own precedent PR carries this same defect today** — its
"Build the documentation" check is red for exactly this reason
(`CHANGELOG.md:15` and `:20`, `myst cross-reference target not found:
'./HISTORY.md'`). I fixed it here rather than reproducing it, since
CLAUDE.md requires reproducing the docs gate and getting it green
before claiming the work is done. A reviewer of the btclib PR may want
to know this.

Everything else that named `HISTORY.md` moves too, discovered with
`git grep` rather than assumed from the sibling repo's file list (this
repo has no script engine and a different test suite, so the list
differs):

- **Load-bearing**: `release.yml`'s `awk` reads the file by name twice
  — once in "Check that the release notes are retitled for the tag"
  (refuses a tag whose `## v<version>` heading was not retitled) and
  once in "Create the release, with RELEASE_NOTES.md as its notes"
  (extracts that section as the GitHub release body). A missed rename
  in either would not fail loudly; it would silently misbehave at the
  next tag push — the first check would stop firing on this file, the
  second would produce empty release notes.
- **Load-bearing**: `.gitattributes` keeps `HISTORY.md merge=union`
  under the new name, so two branches each appending a release-note
  bullet at the top of the file still merge without a conflict.
- `docs/source/history_link.md` moves to
  `docs/source/release_notes_link.md`, its `{include}` target updated,
  and its `index.rst` toctree entry relabelled `RELEASE NOTES` to match
  the other wrapper pages' convention (`README <readme_link.md>`,
  `CONTRIBUTING <contributing_link.md>`, etc.).
- `pyproject.toml`'s `changelog` URL under `[project.urls]`, plus its
  explanatory comment.
- `.github/workflows/test.yml`'s `prose` regex, which decides whether a
  pull request touches only documentation and can skip the build
  matrix.
- `.github/PULL_REQUEST_TEMPLATE.md`, `.github/scripts/check_submodule_pin.py`
  (both the docstring and the runtime error message),
  `CLAUDE.md`, `CONTRIBUTING.md`, `RELEASING.md`, `REVIEWING.md`, and
  `tests/test_citations.py`'s docstring, which names both files as out
  of scope for its citation check.

This repository has no `tests/release_notes_test.py`-style fixture
naming both files by path, and no GitHub Pages / Jekyll site config —
its docs are published entirely through Read the Docs, which builds
straight from `docs/source` on every push, so there is no separate
static mirror of `HISTORY.md` to worry about.

## Redirect decision

No redirect for the old name. Unlike btclib.org, this repository does
not serve a website of its own that could carry a redirect — its only
published documentation is `btclib-secp256k1.readthedocs.io`, built by
Read the Docs directly from this tree — so there is nothing to add a
redirect *to*. The file's git history is where the old name stays
reachable, same reasoning as the sibling PR.

## Test plan

- [x] `uv run pytest --cov`: 735 passed, 100.00% coverage (submodule
      initialized, `uv sync` builds the extension)
- [x] `uv run pre-commit run --all-files`: every hook passes, including
      `codespell`, `markdownlint-cli2`, `mypy`, `actionlint`, `zizmor`
      --keep-going -b html docs/source docs/build/html`: builds clean,
      0 warnings (see the CHANGELOG.md preamble note above for why this
      needed an extra edit beyond the sibling PR's own diff)
- [x] `git grep -ln 'HISTORY\.md'` after the change: only `CHANGELOG.md`
      itself, in its own historical entries

## Summary by Sourcery

Rename HISTORY.md to RELEASE_NOTES.md and update all repository references so release notes, documentation, and publishing workflows remain consistent.

Enhancements:
- Rename the release-notes file from HISTORY.md to RELEASE_NOTES.md and update documentation, packaging metadata, workflows, release tooling, and contributor guidance to use the new name.
- Keep CHANGELOG.md as the historical record while correcting its current cross-reference to the renamed release-notes file.
- Update the Sphinx documentation wrapper and navigation to publish RELEASE_NOTES.md under a matching RELEASE NOTES label.

CI:
- Update documentation-only change detection and release workflow references for RELEASE_NOTES.md.

Documentation:
- Align release-related documentation and repository references with the RELEASE_NOTES.md filename.

Chores:
- Preserve the release-notes merge behavior and update related validation messages and scope references.


fametrano:
Fixed in 62808f8: both comments updated to `RELEASE_NOTES` — `docs/source/conf.py:97-101` and `docs/README.rst:50`. Gates green (`uv run pytest` 735 passed, `uv run pre-commit run --all-files`, sphinx `-W --keep-going`).

The collateral "Four pages" miscount and the missing `Closes #N` are left as the review itself framed them — the former predates this diff and is flagged for a person to decide on separately, the latter is a question for @fametrano rather than a finding.

fametrano:
Proactive fix in 2f6c9b3: the CHANGELOG.md entry cited btclib's issue 1011 as a bare `#1011`, which resolves in this repository rather than in btclib (no issue/PR 1011 exists here — verified 404). Caught before review flagged it, same class of defect as the `#1022` citation fixed by review on the sibling tag-integrity PR (#278).

PR #303, opened 2026-08-21
Fix the required-version comment: the floor tracks Dependabot, not the hook

### LINKED ISSUE #299: required-version's floor may break Dependabot's uv updates, and its comment is false
`[tool.uv] required-version` differs across the three publishing
repositories, and in two of them the comment above it is false:

| | floor | comment says |
| --- | --- | --- |
| btclib | `>=0.11.31` | Dependabot's bundled uv is the ceiling |
| btclib-secp256k1 | `>=0.12.0` | "the version the uv-lock hook pins" |
| bitcoin-core-rpc | `>=0.12.3` | the same |

**All three pin `astral-sh/uv-pre-commit` at `rev: 0.12.5`**, so the
stated reason is not the value in either of the two repositories that
state it.

## Why this may be more than a wrong comment

btclib's block was rewritten and says the opposite, and names a
constraint the other two do not:

> Dependabot runs `uv lock` with exactly that version regardless of this
> key, so any higher floor fails every update it attempts before it
> starts.

If that holds, `>=0.12.0` and `>=0.12.3` are not merely mis-commented —
they may be **silently breaking Dependabot's uv-driven updates** in those
two repositories, which would show up as lockfile updates that simply
never arrive rather than as anything red.

## Verify before changing

The claim is btclib's and it is not obviously self-evident. Check it
against a recent Dependabot pull request in each repository — one that
touches `uv.lock` — and read which uv it ran and whether it succeeded. If
Dependabot's bundled uv is at or above `0.12.3`, the floors are harmless
and only the comments are wrong; if it is below, the floors are a live
defect and the symptom is silence.

Either way both comments need replacing with whatever turns out to be
true.

Filed from the release-machinery survey; entry point
[btclib-org/btclib#1152](https://github.com/btclib-org/btclib/issues/1152).

fametrano:
## Summary
- The `[tool.uv]` comment above `required-version` claimed the floor was
  the version the uv-lock hook of `.pre-commit-config.yaml` pins. It
  never was: the hook pins `0.12.5` while the floor reads `>=0.12.0`.
- The real constraint is Dependabot's own uv-ecosystem updater: it runs
  `uv lock` with exactly the uv it bundles regardless of this key, and
  refuses instead of upgrading itself when that uv falls short of the
  key. Checked against `dependabot/dependabot-core`'s `uv/Dockerfile`
  (bundled uv currently `0.12.1`, last bumped in PR #15770) and against
  btclib's issue #485, where a floor ahead of Dependabot's bundled uv
  failed every lock update -- security updates included -- with
  `tool_version_not_supported` before a single one started.
- `0.12.1` still clears this repository's `0.12.0` floor, so the floor
  itself is not raised or lowered here -- only the comment, which now
  states the actual reason instead of a number that was never true.

Closes #299

## Test plan
- [x] `git submodule update --init && uv sync --locked`
- [x] `uv run pre-commit run --all-files` (exit 0)
- [x] `uv run pytest` (765 passed)
- [x] `uv run pytest --cov` (100.00% coverage, gate satisfied)
      --keep-going -b html docs/source docs/build/html` (build succeeded)

https://claude.ai/code/session_018uNzuzag82WGhvznByjFDH

## Summary by Sourcery

Clarify that the uv version floor must remain compatible with Dependabot's bundled uv rather than tracking the pre-commit hook version.

Bug Fixes:
- Correct the `[tool.uv]` `required-version` comment to accurately describe Dependabot's bundled uv as the constraint on the version floor.

Documentation:
- Document the corrected rationale for the uv version floor in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,22 @@ release-notes length in the first place, and are still in
   baseline sizes, one of which had already drifted:
   `.secrets.baseline`'s documented 53 findings against the actual 54.
 
+### CI
+
+- **`[tool.uv]`'s `required-version` comment stated the wrong reason**
+  (#299): it read as though the floor were the version the uv-lock hook
+  of `.pre-commit-config.yaml` pins, but that hook pins `0.12.5` while
+  the floor reads `>=0.12.0` -- the two numbers were never the same
+  one. The real constraint is Dependabot's own uv-ecosystem updater: it
+  runs `uv lock` with exactly the uv it bundles
+  (`dependabot/dependabot-core`'s `uv/Dockerfile`, `0.12.1` at the time
+  of writing) and refuses instead of upgrading itself when that uv
+  falls short of this key. btclib's issue #485 is what a floor ahead of
+  it costs: every lock update it attempted, security ones included,
+  failed with `tool_version_not_supported` before a single one started.
+  `0.12.1` still clears `0.12.0`, so the floor itself does not move here
+  -- only the comment does.
+
 ## v0.8.0.4
 
 ### `musig` wraps MuSig2, closing the one exception `lib` was for
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -243,16 +243,21 @@ dev = [
 ]
 
 [tool.uv]
-# the oldest uv allowed to touch uv.lock, and it is the version the
-# uv-lock hook of .pre-commit-config.yaml pins: that hook is what
-# regenerates the file, and its rev is something pre-commit.ci moves.
-# Declared here the floor holds for every uv command rather than for the
-# hook alone, so an older uv on a developer's machine says so instead of
-# rewriting the lock in a format the runners then fail to read — uv is
-# pre-1.0 and changes that format. It also reaches CI without a second
-# pin: setup-uv, given no version input, reads this key, and a bare
-# minimum specifier makes it install the latest uv, so the runners are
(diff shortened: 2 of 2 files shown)


fametrano:
The review left one thing open — that the `dependabot-core` claim was
cited rather than fetched, network being unavailable to it, and that
someone with access should spot-check it. Done, against `main`:

```text
uv/Dockerfile:               FROM ghcr.io/astral-sh/uv:0.12.1 AS uv
uv/helpers/requirements.txt: uv==0.12.1
uv/lib/dependabot/uv/file_updater/lock_file_error_handler.rb:
  UV_REQUIRED_VERSION_REGEX =
    /Required uv version `(?<required>[^`]+)` does not match the running version `(?<running>[^`]+)`/
```

Both pins agree, and the third file is the part that makes this more
than an inference: the updater has a named error for exactly this
condition and converts it to `tool_version_not_supported`, so the
refusal is upstream's designed behaviour rather than something to
deduce from a version comparison.

`0.12.1` clears this repository's `>=0.12.0`, so the floor here is
harmless and the comment was the only defect — which is what this pull
request changes and all it changes.

The same check found the floor in `bitcoin-core-rpc`, `>=0.12.3`, is
*not* cleared: a live defect rather than a wrong comment, tracked as
that repository's issue #159 and fixed by its pull request #167. This
one stays scoped to the comment.

PR #308, opened 2026-08-21
Coverage measures tests/, and the shebang pair is guarded

fametrano:
Part of btclib-org/.github#7, "Adopt the practices the audit sent back
into the standard". This repository already had branch coverage,
`exclude_also` and pydoclint; it owed the two below.

## `tests` added to `[tool.coverage.run] source`

`source = ["btclib_secp256k1", "tests"]`, matching the sibling
repositories. The suite enforcing the 100% ratchet was the one thing the
ratchet said nothing about. Measuring it found four gaps, none of them
cosmetic:

- **`point_add`'s identity and inverse cases** (`tests/test_vectors.py`
  108, 110). It is the independent arithmetic the recovery id 2 and 3
  fixture is built against, and its only caller, `point_mul`, never
  hands it `None` on the right nor two points that cancel. Closed by
  `test_point_add_at_infinity`, a real test: a reference implementation
  that is wrong here is wrong in a way no vector would report.
- **`_calls`' second reading** (`tests/test_secret.py`, partial branch
  67->63): a call whose function is neither a name nor an attribute
  contributes no name, which is exactly the blind spot the population of
  `test_every_function_that_takes_a_secret_out_offers_into` is defined
  by — a producer invoked as `table[key]()` is not in it. Closed by
  `test_calls_reads_nothing_from_a_call_that_names_no_function`, which
  runs the helper as well as reading it, so the two assertions are a
  finding (the secret does come out; the walk reports no call) rather
  than a restatement. The helper is module-level because
  `inspect.getsource` of a nested function answers an indented block
  that `ast.parse` refuses.
- **two `raise AssertionError` lines** (`tests/test_submodule_pin.py`
  206, `tests/test_verified_signing.py` 246), in stubs substituted to
  prove a path is never taken. No test can cover them, and for a reason
  stronger than reachability: executing one is the run in which the
  suite is red, so a green run cannot reach it and no amount of testing
  would. They join `raise RuntimeError` in `exclude_also`, with the
  reasoning beside it. The package raises no `AssertionError`, so
  nothing shipped is excluded. Two `# pragma: no cover` was the
  alternative, and it records the decision once per occurrence.

Coverage is back at 100.00% with `tests/` in scope.

## `check-shebang-scripts-are-executable`

It was not merely absent: it sat commented out beside
`check-executables-have-shebangs` under a single comment covering both.
That comment was right about the latter — `git ls-files -s` reports mode
100644 for every path here (the submodule gitlink aside), and that hook
selects on `types: [text, executable]`, so it matches no file and
`check-hooks-apply` would say so. It was wrong about the former, which
selects on `types: [text]` alone and therefore matches every file in the
tree whether or not any carries a shebang. So the same measurement
enables one and keeps the other off, and the comment is split in two.

What the hook found: nothing. `grep -l '^#!' scripts/*.py
.github/scripts/*.py` matches nothing and no tracked file carries the
execute bit — as it should be, given how these are invoked. hatchling
`exec()`s `scripts/hatch_build.py` and `scripts/cffi_build.py` from
inside the build, `scripts/README.md` gives the in-place build as
`uv run python scripts/cffi_build.py`, and the three under
`.github/scripts/` run as `python .github/scripts/<name>.py` from
`.pre-commit-config.yaml` and the workflows. The caller names the
interpreter every time, so a shebang would announce one nothing
consults — and, unpaired with the execute bit, would cost this guard.
`fix-byte-order-marker`, the entry of the same shape, was already on and
is not duplicated.

## Verification

Setup: `git submodule update --init --recursive` then `uv sync --locked`,
both exit 0.

| command | exit code |
| --- | --- |
| `uv run pre-commit run --all-files` | 0 |
| `uv run pytest` | 0 (767 passed) |
| `uv run --locked --no-default-groups --group docs sphinx-build -W --keep-going -b html docs/source docs/build/html` | 0 |

`uv run pytest --cov` also exits 0, at 100.00% with `tests/` measured.
The sphinx command is CONTRIBUTING.md's own, which differs from the
plain `uv run sphinx-build` form.

## Summary by Sourcery

Strengthen quality gates by measuring test coverage comprehensively and enforcing consistent shebang and executable permissions.

New Features:
- Enable the shebang/executable pairing pre-commit check to prevent future script permission mismatches.

Bug Fixes:
- Add coverage for elliptic-curve point-at-infinity and inverse cases and for secret extraction through dynamically selected calls.

Enhancements:
- Measure the tests alongside the package for complete coverage enforcement and exclude intentionally unreachable assertion failures from the coverage calculation.

Documentation:
- Document the expanded coverage gate and shebang-check policy in the changelog.

Tests:
- Extend tests to cover previously unmeasured arithmetic and AST-walking edge cases.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,12 +68,28 @@ repos:
         exclude: ^tests/bip3(40|24)_.*\.csv$
       - id: check-case-conflict
       # off, and measured rather than assumed: nothing tracked here carries
-      # the execute bit and no python file opens with a shebang, so both of
-      # these apply to no file at all. check-hooks-apply below is what would
-      # report them, a hook matching nothing being a rule that has stopped
-      # running rather than a rule that passes
+      # the execute bit -- `git ls-files -s` reports mode 100644 for every
+      # path, the submodule gitlink aside -- and this hook selects on
+      # types [text, executable], so it applies to no file at all.
+      # check-hooks-apply below is what would report that, a hook matching
+      # nothing being a rule that has stopped running rather than a rule
+      # that passes
       # - id: check-executables-have-shebangs
-      # - id: check-shebang-scripts-are-executable
+      # its converse is on, the same measurement reading the other way.
+      # This one selects on types [text] alone, so it matches every file
+      # here whether or not any carries a shebang: what it enforces is the
+      # pair, and the day a file needs one it also needs chmod +x. Nothing
+      # needs one today, which is a fact about how these scripts are
+      # invoked rather than an omission -- a shebang is read by the kernel
+      # only when a file is exec'd directly, and none is. hatchling
+      # `exec()`s scripts/hatch_build.py and scripts/cffi_build.py from
+      # inside the build, `uv run python scripts/cffi_build.py` is how
+      # scripts/README.md gives the in-place build, and the three under
+      # .github/scripts/ are run as `python .github/scripts/<name>.py` by
+      # this file and by the workflows. In each the interpreter is named
+      # by the caller, so a shebang would announce one nothing consults --
+      # and, unpaired, would cost this guard
+      - id: check-shebang-scripts-are-executable
       # finds nothing today, and stops an editor on another platform from
       # adding three bytes tomorrow
       - id: fix-byte-order-marker
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,67 @@ release-notes length in the first place, and are still in
   their own; `btclib`'s `test.yml` carries no Windows row and
   `bitcoin-core-rpc`'s carried one further instance of its own.
 
+### The gate
+
+- **Coverage measures `tests/` as well as the package**
+  (btclib-org/.github#7). `[tool.coverage.run] source` named
+  `btclib_secp256k1` alone, so the one thing the 100% ratchet said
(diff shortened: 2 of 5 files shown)


fametrano:
Rebased onto current `origin/main`: `main` had advanced past this
branch's base with #306 (CONTRIBUTING.md and CLAUDE.md catching up with
`test.yml`'s `--frozen`), leaving this PR `BEHIND` rather than
conflicted. The rebase applied with no conflict; what #306 brought in
touches `CHANGELOG.md`, `CLAUDE.md` and `CONTRIBUTING.md` only, and
neither the coverage `source` nor the pre-commit entry here is near it.
The `CHANGELOG.md` entries interleaved as intended -- `### The gate`
still follows `### Documentation` and `### CI` under v0.8.0.5.

Every gate re-run on the rebased tree, none of the earlier numbers
reused:

| command | exit code |
| --- | --- |
| `git submodule update --init --recursive` | 0 |
| `uv sync --locked` | 0 |
| `uv run pre-commit run --all-files` | 0 |
| `uv run pytest` | 0 (767 passed) |
| `uv run --locked --no-default-groups --group docs sphinx-build -W --keep-going -b html docs/source docs/build/html` | 0 |
| `uv run pytest --cov` | 0, 100.00% with `tests/` measured |

New tip is 4eb1505, signed (`git log -1 --format='%G? %GS'` reports
`G`).

Issue #178, opened 2026-08-16
A parity assertion that cannot fail, and one equation where two schemes need two


fametrano:
Two leftovers of #174 and #175, both found in review and both landed
because they hold nothing up: an assertion that cannot fail, and one
equation offered for two schemes.

## The parity assertion is an identity

`tests/test_properties.py`, in `test_serialization_round_trips`:

```python
uncompressed = keys.pubkey_from_prvkey(prvkey, compressed=False)
compressed = compress(uncompressed)
...
assert keys.pubkey_from_prvkey(prvkey) == compressed
...
y = int.from_bytes(uncompressed[33:], "big")
assert y % 2 == compressed[0] - 2
```

`compress` composes its first octet as `2 + (pubkey_bytes[64] & 1)`, and
`y % 2` of a big-endian read of `uncompressed[33:]` is that same bit. So
the assertion compares a parity with itself, and the fact it means to
state — that libsecp256k1's compressed prefix agrees with the y it wrote
in the other serialization — is what the assertion three lines above
already says, `compressed` being the composition and the left side
libsecp256k1's own answer.

The curve equation beside it is the line that reads the octets rather
than restating them, and stays.

## SECURITY.md gives one equation for two schemes

The nonce bullet says a nonce yields the private key given the signature
it made, and offers `d = (s·k − h)/r mod n`. That is ECDSA's, from
`s = k⁻¹(h + r·d)`. BIP340 signs `s = k + e·d` with `e = H(R ‖ P ‖ m)`,
so there it is `d = (s − k)/e mod n`. Two entry points, two schemes, two
equations — and this is the bullet whose whole point is that the danger
has one.

The same bullet opens in bold where the two around it open in plain text.

PR #312, opened 2026-08-21
check-dist gains a wheel-content check

fametrano:
## Summary

btclib-org/btclib#1160 surveys two half-measures for one question,
across the organization's three publishing repositories: `btclib`
runs a hand-written script (`verify_dist_contents.py`) comparing a
built wheel and sdist, member for member, against an allowlist; this
repository configures `check-wheel-contents` (`ignore = ["W003",
"W009"]`, for the shared library a dynamic wheel ships beside the
package) but has no script of its own; `bitcoin-core-rpc` had neither.

## Current state (established before deciding)

| repository | member-listing script | configured `check-wheel-contents` | runs on |
| --- | --- | --- | --- |
| `btclib` | `verify_dist_contents.py`, paired with `docs/source/package-content-policy.md` via a test comparing the two | not configured (defaults) | `test.yml`'s `dist` job **and** `release.yml`'s `build` job — the exact files it uploads |
| `btclib-secp256k1` (this repository, before this PR) | none | `ignore = ["W003", "W009"]` | `test.yml`'s `check-dist` job, which `release.yml`'s `test` job calls and whose artifacts the publish jobs upload — the exact files it uploads |
| `bitcoin-core-rpc` | none (as of this PR: `check-wheel-contents --package`, see PR #178) | not configured, before #178 | `test.yml`'s `dist` job only |

This repository already has the better artifact story of the three
(divergence #9 of btclib-org/btclib#1152): no separate rebuild in
`release.yml`, so `check-dist` inspects the same wheels the publish
jobs upload. What it lacked is any check of what a member actually
*is*.

## What check-wheel-contents cannot phrase here, measured

`--package btclib_secp256k1` is the flag that would do for this
repository what PR #178 does for `bitcoin-core-rpc`: diff the wheel's
package tree against the checkout's. It does not work here, and I
measured why before writing anything else. Every wheel this project
ships carries a compiled artifact at the wheel's own root, beside
`btclib_secp256k1/` rather than inside it — the extension module for a
static wheel, or the ABI-mode module and the shared `libsecp256k1` for
a dynamic one. Running `check-wheel-contents --package
btclib_secp256k1 --ignore W003,W009` against a real static wheel:

```
W102: Wheel library contains files not in package tree:
  _btclib_secp256k1.cpython-314-darwin.so
```

against the compiled extension this build is *supposed* to carry.
`--package` has no way to name a second, legitimate top-level entry
outside the tree it diffs, so the flag cannot be configured into
correctness here the way it was for `bitcoin-core-rpc`.

Unconfigured `check-wheel-contents` (today's state) also does not
catch a dropped `py.typed`: built a wheel with it stripped out and
`RECORD` edited to match, `check-wheel-contents` (with or without the
`ignore` this repository already carries) reports `OK`.

## Decision: a script, tailored to two wheel kinds, wheel-only

`.github/scripts/verify_wheel_contents.py` asks the two questions
`--package` cannot phrase without a false positive: that
`btclib_secp256k1/` in the wheel is exactly what this checkout's own
directory has (read fresh via `PACKAGE_DIR.rglob`, not a copy of the
file list — a module added to the package is a module the next wheel
has to carry, with nothing here to edit to match), and that the
kind-appropriate top-level artifact is present *and not empty* — a
question no `check-wheel-contents` code answers for any member, static
or dynamic wheel alike. The wheel's own filename tag (`py3-none-...`
vs `cpNN-...`) is read to tell the two kinds apart, the same
distinction `scripts/hatch_build.py` draws when it writes the tag.

`docs/source/package-content-policy.md` states the policy in prose,
wired into the toctree beside `btclib`'s own page.
`tests/test_wheel_contents.py` compares the page against the script's
six constants in both directions — the pairing mechanism `btclib`
carries, ported here rather than the constants themselves, which
describe a different wheel shape. 21 tests, one failure mode planted
per rule (a forbidden suffix, a forbidden name, `__pycache__`, an
undeclared `.dist-info` member, a missing dist-info file, a dropped
package file, an untracked package file, a second static extension, a
missing dynamic shared library, a versioned shared-library name, an
empty top-level artifact) plus the two pairing tests and CLI-level
tests, all passing against real wheels built both static and dynamic.

**The sdist is out of scope**, and says so in both the script's
docstring and the policy page: `[tool.hatch.build.targets.sdist]
exclude` here is an *exclude* list, the opposite shape from `btclib`'s
*include* `MANIFEST.in` — a file the tree gains ships by default here,
rather than silently not shipping — so the failure `btclib`'s script
exists for does not arise the same way. This repository's sdist has
roughly three hundred members, most of them the vendored `secp256k1/`
submodule; modelling that as an allowlist would be exactly what the
issue warns against — "a check whose allowlist nobody can read is a
check nobody maintains" — for a question `exclude` already answers by
construction. Also noted, out of scope for this PR: unlike `btclib`
and `bitcoin-core-rpc`, this repository has no `check-manifest` hook
at all, so nothing here answers "does the sdist match the tracked
tree" independently of the `exclude` list's own correctness — a
different, narrower question than this PR's, and a candidate for its
own issue if it turns out to matter given the `exclude` shape.

Wired into `test.yml`'s `check-dist` job, right after the existing
`check-wheel-contents` step, over the same `dist/*.whl` this job
already downloads from `build-cibuildwheel`, `build-dynamic` and
`build-windows`.

This PR leaves btclib-org/btclib#1160 open: the issue lives in
`btclib`, and its own pull request is where the decision documented
here gets closed out.

## Verification

- `uv run --locked --only-group lint pre-commit run --all-files`:
  exit 0 (36 hooks, no `check-manifest` hook in this repository — see
  above)
- `uv run --locked --no-default-groups --group test pytest --cov
  --cov-report=term-missing:skip-covered`: exit 0, 786 passed, 100%
  coverage
- `uv run --locked --no-default-groups --group docs sphinx-build -W
  --keep-going -b html docs/source docs/build/html`: exit 0,
  `package-content-policy` renders

`check-dist`'s own commands, from CONTRIBUTING.md, run by hand against
a real static wheel, a real dynamic wheel (`BTCLIB_LIBSECP256K1_DYNAMIC=true`,
`MACOSX_DEPLOYMENT_TARGET=11.0`) and a real sdist, all three built
from this branch:

- `uv build --wheel` (static), `BTCLIB_LIBSECP256K1_DYNAMIC=true uv
  build --wheel` (dynamic), `uv build --sdist`: exit 0 each
- `uv run --locked --only-group check twine check --strict dist/*`:
  exit 0, `PASSED` on all three files
- `uv run --locked --only-group check check-wheel-contents
  dist/*.whl`: exit 0, `OK` on both wheels
- `uv run --no-project --python 3.14
  .github/scripts/verify_wheel_contents.py dist/*.whl`: exit 0,
  "carries what it may and no more" on both wheels
- `uv run --locked --only-group check pyroma --min 10 dist/*.tar.gz`:
  exit 0, 10/10
- installed the static wheel alone, isolated, from an empty directory,
  and ran `import btclib_secp256k1 as m; print(m.__version__,
  m.lib)`: exit 0, `0.8.0.5 <Lib object for '_btclib_secp256k1'>`

Re-ran pre-commit and pytest after rebasing onto origin/main: both
exit 0, 786 passed.

## Summary by Sourcery

Validate published wheels against the repository package-content policy for both static and dynamic builds.

New Features:
- Add a wheel-content validation script that verifies package completeness, metadata members, permitted top-level artifacts, forbidden files, and non-empty compiled artifacts for both static and dynamic wheels.

Bug Fixes:
- Prevent incomplete, unexpected, or malformed wheel contents from reaching published distributions.

Enhancements:
- Document the wheel package-content policy and explicitly scope sdist validation out of this check.

CI:
- Run the wheel-content validation in the check-dist job against the same wheels used for release publishing.

Documentation:
- Add the package-content policy to the documentation to describe allowed and required wheel members.

Tests:
- Add comprehensive tests covering valid static and dynamic wheels, invalid members, missing files, artifact variants, CLI behavior, and policy/script consistency.

Chores:
- Record the new wheel-content validation in the changelog.

### DIFF
--- a/.github/scripts/verify_wheel_contents.py
+++ b/.github/scripts/verify_wheel_contents.py
@@ -1,0 +1,277 @@
+# Copyright (c) The btclib developers
+#
+# Distributed under the MIT software license, see the accompanying
+# LICENSE file or https://opensource.org/license/mit for the full text.
+
+r"""Check a built wheel against an allowlist of its members.
+
+`twine check`, `check-wheel-contents` and `pyroma` read the *metadata* of
+a wheel: whether it is syntactically valid, whether the long description
+renders, whether the classifiers a package should declare are there.
+None of them reads every member, and what is in the archive is what a
+user installs.
+
+`check-wheel-contents --package btclib_secp256k1`, configured in
+`pyproject.toml` for `btclib`, is the closest built-in answer, and it
+does not fit here: it diffs the wheel's whole library against the one
+named package tree, and this package's wheel is not one package tree.
+Every wheel this project ships also carries a compiled artifact at the
+wheel's own root, beside `btclib_secp256k1/` rather than inside it -- the
+extension module for a static wheel, that module's ABI-mode source and
+the shared `libsecp256k1` for a dynamic one -- which `--package` reports
+as "files not in package tree" whether or not they are the ones this
+project intends. `[tool.check-wheel-contents]` keeps `ignore = ["W003",
+"W009"]` for the same reason, on the dynamic wheel specifically: the
+checks those codes name are about a *non-package* file at the top level,
+which for this project's dynamic wheel is not a mistake but the shared
+library.
+
+Three questions, and what answers them:
+
+**Which top-level artifact must be there.** A wheel's tag says whether it
+is static (an `infer_tag`ed `cpNN-...`, one compiled extension) or
+dynamic (`py3-none-<platform>`, an ABI-mode module and the shared library
+beside it) -- `scripts/hatch_build.py`'s own distinction, and this reads
+the same tag it writes. Which artifacts a wheel of that kind must carry,
+and refusing an extra one -- a stray file that is not part of
+`btclib_secp256k1/` and not one of them -- is what `--package` cannot
+phrase without also refusing the artifact that belongs.
+
+**Whether the artifact is really there**, and not a zero-byte member a
+half-finished build step left behind: `scripts/cffi_build.py` raises when
+it cannot find one to copy, not when it copies an empty one, so an empty
+file is a shape that check has no report for. `toplevel_reason` below
+checks the kind-appropriate artifact for size as well as presence, which
+none of `check-wheel-contents`'s codes do for any member.
+
+**Whether `btclib_secp256k1/` itself is complete.** Everything under
+that directory in the wheel has to be everything the checkout has under
+it -- source, `py.typed`, nothing more and nothing less -- which is
+`--package`'s own question, asked here by hand rather than through a
+flag that would also judge the sibling artifact.
+
+The sdist is not this script's subject. `btclib`'s sibling script checks
+it because `MANIFEST.in` there is an *include* list -- a file the tree
+gains and MANIFEST.in does not name is a file the sdist silently drops,
+which is the failure this whole class of check exists for.
+`[tool.hatch.build.targets.sdist] exclude` here is the opposite shape: a
+file the tree gains ships by default, and has to be named to be left
+out, so the failure mode is an sdist too wide rather than one silently
+narrow. What that leaves unchecked -- an exclude pattern that lists a
+file this build still needs -- is a build failure, not a silent gap,
+which is a different question with a different check to answer it.
+
+Run it against one built wheel:
+
+    uv run --no-project --python 3.14 \\
+        .github/scripts/verify_wheel_contents.py dist/*.whl
+
+`test.yml`'s `check-dist` job runs it on every wheel it downloads --
+`build-cibuildwheel`, `build-dynamic` and `build-windows` between them --
+which are the artifacts `release.yml`'s own `test` job produces and the
+publish jobs upload: one build, checked and shipped, so this judges the
+files this project actually publishes rather than a copy of them.
+"""
+
+from __future__ import annotations
+
+import sys
+import zipfile
+from pathlib import Path
+
+# the package this repository ships, read fresh from the checkout rather
+# than hard-coded: a module added to it is a module the next wheel has to
+# carry, and a copy of the file list here would drift the day one is
+# added or removed without anybody editing this script to match
+PACKAGE_DIR = Path(__file__).parents[2] / "btclib_secp256k1"
+
+# no member of this wheel carries any of these, whichever top-level
+# artifact the kind below admits: a `.pth` executes at interpreter
+# startup, before the first import; a `.pyc` is a compiled module nobody
+# reviewed; an archive inside an archive is a second package installing
+# with the first. Checked against `btclib_secp256k1/` and the
+# top-level artifact -- not against `.dist-info`, which is already
+# validated member by member, against its own allowlist, above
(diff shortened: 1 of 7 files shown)


fametrano:
## Verdict

**Measured, not assumed**: this repository's wheel is not one package
tree, so `check-wheel-contents --package btclib_secp256k1` — the flag
`bitcoin-core-rpc` configured directly in its own PR
(btclib-org/bitcoin-core-rpc#178) — was tried here first and rejected on
evidence, not by inspection. Run against a real static wheel with
`--ignore W003,W009` (this repository's existing config) added:

```
W102: Wheel library contains files not in package tree:
  _btclib_secp256k1.cpython-314-darwin.so
```

against the compiled extension that wheel is *supposed* to carry.
`--package` has no way to name a second, legitimate top-level entry
outside the tree it diffs, so the flag cannot be configured into
correctness here the way it was for `bitcoin-core-rpc`. That false
positive is why this PR is a script rather than a configuration line,
and it is stated in the script's own docstring, the policy page and
the PR body — not only here.

Unconfigured `check-wheel-contents` also does not catch a dropped
`py.typed`: built a wheel with it stripped and `RECORD` edited to
match, and the existing (or a hand-tightened) config reports `OK`
regardless. The new script's own test plants that exact case
(`test_a_missing_package_file_is_reported`) and 20 others.

## Local verification (sha 1f37015af1304cdbe315dd70e0c2c029dedb2153)

- `uv run --locked --only-group lint pre-commit run --all-files`:
  exit 0 (no `check-manifest` hook in this repository — noted, not
  fixed here, see the PR body)
- `uv run --locked --no-default-groups --group test pytest --cov
  --cov-report=term-missing:skip-covered`: exit 0, 786 passed, 100%
  coverage
- `uv run --locked --no-default-groups --group docs sphinx-build -W
  --keep-going -b html docs/source docs/build/html`: exit 0,
  `package-content-policy` renders
- `check-dist`'s own commands, run by hand against a real static
  wheel, a real dynamic wheel and a real sdist, all built from this
  branch: `uv build --wheel` (both kinds, exit 0 each), `uv build
  --sdist` (exit 0), `twine check --strict` (exit 0, `PASSED` on all
  three files), `check-wheel-contents` (exit 0, `OK` on both wheels),
  `verify_wheel_contents.py` (exit 0, "carries what it may and no
  more" on both wheels), `pyroma --min 10` (exit 0, 10/10), installed
  the static wheel alone, isolated, from an empty directory (exit 0)
- All of the above re-run clean after rebasing onto origin/main

## CI

Completed, green: `Build the documentation`, `Lint and type-check`,
`pre-commit.ci - pr`.

**Outstanding, as outstanding**: `Claude review` and `Decide whether
the matrix has anything to check` are still `pending` as of this
comment — the wheel-build matrix (`build-cibuildwheel`,
`build-dynamic`, `build-windows`) and `check-dist` itself have not yet
reported. `Claude answers a mention` is `skipping` (no mention on this
PR), expected and not a gate. Local verification above covers the same
ground `check-dist` runs in CI, on both wheel kinds and the sdist, but
is not a substitute for that job actually passing on this sha.

ACK 1f37015af1304cdbe315dd70e0c2c029dedb2153

